PR #297, opened 2014-10-13
Allows writing lint-friendly tests

### LINKED ISSUE #41: Allow writing JSLint/JSHint friendly tests
When using expect() for writing asserting, some of the assertion are using properties as the mean for calling assert.

For example

expect(foo).to.exist

JSLint/JSHint will complain about such code, since it expects an assignment or function call.

Hiding a function call, behind getting a property can be sweet but in tests, it leads to such warnings.

I know that one can use assert calls, or not use JSHint/JSLint or configure JSHint/JSLint to ignore such warnings, but I just want to know what is your feeling about this problem.

joshperry:
This makes the final assertion checks noop functions so that they can be formatted like function calls. This keeps linters (JSLint/JSHint) off our back for having an inop expression.

Fixes #41

### DIFF
--- a/test/expect.js
+++ b/test/expect.js
@@ -92,6 +92,7 @@ describe('expect', function () {
 
     expect(foo).to.exist();
     expect(foo).to.exist().and.contain('bar');
+    expect(foo).to.exist.and.contain('bar');
   });
 
   it('arguments', function(){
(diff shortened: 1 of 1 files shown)


elliotf:
I just found this and wanted to say thank you to @joshperry ..  developers on my team have had issues when learning the assertions library and done things like `expect(foo).to.exists;` (or somesuch) not realizing that `exists` is not an actual assertion.  Tests didn't fail because no assertion was being made.  With this change, they'll get an error saying so, as long as they call it as a function!

Thank you!

PR #306, opened 2014-11-13
Revert "Allows writing lint-friendly tests"

keithamus:
Reverts chaijs/chai#297

@logicalparadox @joshperry

### DIFF
--- a/lib/chai/assertion.js
+++ b/lib/chai/assertion.js
@@ -6,7 +6,6 @@
  */
 
 var config = require('./config');
-var NOOP = function() { };
 
 module.exports = function (_chai, util) {
   /*!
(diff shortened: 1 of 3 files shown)


joshperry:
Just wonder if you guys are going to revert this or just leave it until people start to use the new syntax in their code... The chaijs website was also updated to document this new syntax. If we wait too long, reverting may end up being a backcompat issue.

keithamus:
@joshperry indeed. `master` has been reverted we just need to cut a new release - which is dependant on #337

johanneswuerbach:
I think this change should be published asap (and marked as breaking) as it is advertised for ~ 2 months on your homepage and we (and I think others) already converted our projects to the new syntax.

keithamus:
Yes, you're very right @johanneswuerbach. I'm going to work on getting a release out as soon as I can.

sankethkatta:
@paul-barry-kenzan I was wondering the same thing. There seems to be some more context here: https://github.com/chaijs/chai/issues/371#issuecomment-75033714

Issue #330, opened 2014-12-27
Add an assertion that checks how much a value has changed.


oveddan:
This would be used to check if, when something is invoked, something else is changed by x.

For example:

```javascript
var someArray = [1, 2, 3];

expect(function(){
  someArray.pop();
}).to.change(someArray, 'length').by(-1);
```

I'd be glad to take a stab at this.

lucasfcosta:
This one is done.
Implemented at #333 and #621.

More discussion about this was done at #339 too.

PR #333, opened 2014-12-28
Added `change`, `increase` and `decrease` assertions with `by` chain (#330)

cmpolis:
Added `change`, `increase` and `decrease` assertions with `by` chain:

```javascript
buildUser = function() { ... };
removeUsers = function() { ... };
buildUser.should.change(users, 'length');
buildUser.should.change(users, 'length').by(1);
removeUsers.should.decrease(users, 'length');
...
```

I've used `change` a bunch in other testing setups and it allows for more readable and terse tests *(instead of `before=val ... val.should.eq(before+1)`)*.


cmpolis:
Sorry... accidentally committed before coffee :smile: Rebased and put in those changes you noted, thanks!

keithamus:
LGTM :smile:

PR #349, opened 2015-01-22
add the which chain method as per issue #347

### LINKED ISSUE #347: Suggestion: Add a "which" chaining word to be interchangeable with "that"
I find the "that" chaining word can be somewhat unintuitive at times and find myself trying to use "which" instead of "that".

An example being: `...with.property('length').which.equals(10)` seems more intuitive to me than `...with.property('length').that.equals(10)`

Obviously both make sense but I thought I'd mention this given the aim of chaining words, and because there might be others who feel the same.

toastynerd:
Not sure what the etiquette is here but I saw issue #347 and thought I would take a stab at it.

Fixes #347

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -25,6 +25,7 @@ module.exports = function (chai, _) {
    * - been
    * - is
    * - that
+   * - which
    * - and
    * - has
    * - have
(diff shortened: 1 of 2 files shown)


keithamus:
The etiquette is: congratulations @toastynerd you get a million internet points :smile:!

toastynerd:
@keithamus Hooray, internet points!

PR #353, opened 2015-01-28
Support jspm

Alxandr:
This commit also needs to be tagged for jspm to pick it up.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,5 +38,8 @@
     , "karma-phantomjs-launcher": "0.1.1"
     , "mocha": "1.21.x"
     , "istanbul": "0.2.x"
+  },
+  "jspm": {
+    "main": "chai.js"
   }
 }


keithamus:
Sorry @Alxandr I seem to have missed this PR. 

Could you explain why we need this extra option in our package.json? I'm not familiar with jspm, so do not understand why it needs extra fields to cope with Chai. Can it not just use existing package manager configurations, e.g. the component.json?

Alxandr:
Your question is basically just like asking if `bower` can't just use `package.json`. Jspm expects fields in the `package.json`. That's just how it works.

Though, that is a truth with modification, jspm allows for endpoints, so it can actually speak bower, and npm. However, if I install `npm:chai` for instance, it pulls in all of chais npm dependencies (some of which aren't too god with browser usage). And bower isn't standard (at least yet), so if you want jspm users to be able to "just install chai", you need to tell it how it should install. Because currently it defaults to the `index.js` file, which just doesn't work.

keithamus:
The decision to drop in bower's `component.json` predates my presence in this project (I would have likely initially objected unless there was an overwhelming consensus).

The problem with these package managers is each asks for their own syntax & paradigms. It's a slippery slope, where we get into a position of supporting `N` number of package managers each with `N` number of quirks and syntaxes.

I understand that this is a 3 line change - and so seems like a nominal change. But it buys us (the chai community) into an ecosystem (jspm) which we (the chai community) has to then support. 

So... I know it's quite frustrating to hear this @Alxandr, but I'm not sold on merging this. I wont close it - as I am not the community, and the community ultimately makes the decision - but I wont merge until I see a strong set of votes in favour of chai supporting jspm.

Alxandr:
Well, by the end of the day, it's all a matter of wish for audience. By not supporting jspm, you effectively force the users of it to use another testing library. It just means someone will either have to fork chai, or write a new one from scratch, and you'll lose out on some potential users.

keithamus:
@Alxandr for a lot of packages on jspm - an intermediary library has been created, which shoulders the support & compatibility between jspm and the package. Take for example jquery, in the jspm registry, which links to https://github.com/components/jquery (not a jquery project but a "shim repository"). 

The same could be achieved for chai. It alleviates the burden from us to support that ecosystem, while still making the package available on it.

Alxandr:
This is true, but since your repository already contains built sorces (which is one reason why a lot of repositories have intermediary repositores, they put built sources there), and you already have bower support in the main repository, supporting jspm is just a matter of pointing jspm to the correct file (instead of `main` which it defaults to).

Though, it could obviously also be achieved by having a CI autobuild on new releases to the chai repo, add the jspm config bits, and publish to a intermediary repo. It's just more of a "jspm config tends to go where the release bits are, and in your case they are already in the main repo".

keithamus:
Hey @Alxandr I'm going to close this one for now. We behave well for Bower and NPM and to me that feels like enough (certainly enough to have to support). We can maybe revisit this later after v.next or v.next+1 when I anticipate some large changes surrounding this stuff.

PR #356, opened 2015-01-30
Added fail() method to Should and Expect interfaces

Soviut:
Should and Expect now have `fail()` methods like Assert does.
```js
should.fail(0, 1, 'failure message');
expect().fail('failure message');
```
Please pay special attention to the Expect implementation as I wasn't totally certain how all of it worked. It may not be fully chainable, though it's not really intended to be.

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -215,16 +215,6 @@ module.exports = function (chai, _) {
   Assertion.addChainableMethod('contains', include, includeChainingBehavior);
   Assertion.addChainableMethod('includes', include, includeChainingBehavior);
 
-  // TODO: Add jsdoc
-
-  function assertFail (msg) {
-    if (msg) flag(this, 'message', msg);
-    var obj = flag(this, 'object');
-    throw new chai.AssertionError(msg, {}, obj);
-  }
-
-  Assertion.addMethod('fail', assertFail);
-
   /**
    * ### .ok
    *
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -215,6 +215,16 @@ module.exports = function (chai, _) {
   Assertion.addChainableMethod('contains', include, includeChainingBehavior);
   Assertion.addChainableMethod('includes', include, includeChainingBehavior);
 
+  // TODO: Add jsdoc
+
+  function assertFail (msg) {
+    if (msg) flag(this, 'message', msg);
+    var obj = flag(this, 'object');
+    throw new chai.AssertionError(msg, {}, obj);
+  }
+
+  Assertion.addMethod('fail', assertFail);
+
   /**
    * ### .ok
    *
--- a/lib/chai/interface/expect.js
+++ b/lib/chai/interface/expect.js
@@ -9,6 +9,19 @@ module.exports = function (chai, util) {
     return new chai.Assertion(val, message);
   };
 
+  /**
+   * ### .fail(actual, expected, [message], [operator])
+   *
+   * Throw a failure.
+   *
+   * @name fail
+   * @param {Mixed} actual
+   * @param {Mixed} expected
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -21,7 +21,7 @@
   },
   "main": "./index",
   "scripts": {
-    "test": "make test"
+    "test": "mocha --require ./test/bootstrap"
```
keithamus: This looks like an extraneous change. Could you rebase it out please?

lib/chai/core/assertions.js
```diff
@@ -215,6 +215,16 @@ module.exports = function (chai, _) {
...
+  // TODO: Add jsdoc
+
+  function assertFail (msg) {
+    if (msg) flag(this, 'message', msg);
+    var obj = flag(this, 'object');
+    throw new chai.AssertionError(msg, {}, obj);
+  }
+
+  Assertion.addMethod('fail', assertFail);
+
```
keithamus: Curious why you added it as a method for all assertions, rather than just tacking it onto `expect`. I was thinking you would just add `expect.fail = function` much like you did with `should.fail`

keithamus:
Some good work so far @Soviut. Couple of notes above which I'd like to see addressed - if you could be so kind. Also, as you've noted in the code, it'd be great to get some documentation put in here too :smile:

lib/chai/core/assertions.js
```diff
@@ -215,6 +215,16 @@ module.exports = function (chai, _) {
   Assertion.addChainableMethod('contains', include, includeChainingBehavior);
   Assertion.addChainableMethod('includes', include, includeChainingBehavior);
 
+  // TODO: Add jsdoc
+
+  function assertFail (msg) {
+    if (msg) flag(this, 'message', msg);
+    var obj = flag(this, 'object');
+    throw new chai.AssertionError(msg, {}, obj);
+  }
```
keithamus: Another point to make here; for API consistency I think `should.fail` and `expect.fail` should work exactly the same as `assert.fail`. Method signature and all.

Soviut: I agree with the consistency. I chose this route because I was trying to be consistent with the interface. Namely, that values are usually passed along the chain with `expect`. I'll try reworking it.

lib/chai/core/assertions.js
Soviut: I was having issues adding `fail` directly to `expect` because it's a function. Are there any examples of how functions are directly added to expect anywhere else in the code base?

Soviut: @keithamus Can you advise on how to add a `fail` method directly to `expect`? I can't find concrete examples and it seems somewhat impossible because `expect` is a function, unless I'm misunderstanding the architecture.

keithamus: Shouldn't matter that `expect` is a function. Fail can just be added as a property of it. See the example below:

```js
> function x() {}
undefined
> x.fail = function () {}
[Function]
> x
{ [Function: x] fail: [Function] }
> x.fail
[Function]
>
```

Soviut:
@keithamus There is now an `expect.fail(0, 1, 'message');` method. Does this look good now?

keithamus:
Looks fantastic @Soviut.

Could you add a little bit of documentation above each of the methods in the code? This way it'll get carried over onto the site documentation when the site is regenerated.

Soviut:
Will do.

Soviut:
@keithamus documentation added. I basically just copied what was already in the `assert` interface.

keithamus:
Perfect. Consider it merged :+1:

sieira:
Sorry for bothering you, but I can't manage to find the docs for this...

keithamus:
@sieira the website is outdated right now. We're working on it. https://github.com/chaijs/chai-docs/issues/74

PR #359, opened 2015-02-04
Arrays passed to `.keys()` assertions are sorted (unexpectedly) by side-effect

gregglind:
## Example
```
var chai = require("chai");

var orig = ["b", "a"];
var mykeys = ["b","a"];
var O = {
  "b": 1,
  "a": 2
};

chai.expect(mykeys).deep.equal(orig);
chai.expect(O).keys(mykeys);
chai.expect(mykeys).deep.equal(orig);
// AssertionError: expected [ 'a', 'b' ] to deeply equal [ 'b', 'a' ]
```

This is caused by the `.sort()` at https://github.com/chaijs/chai/blob/master/lib/chai/core/assertions.js#L1070

## Possible Repairs

1. remove the `sort`()
2. clone / copy the `expected` list earlier in the function.
3. Something else

## Other implications

- unknown, thus my ask :)  I don't know Chai's code base and there may be other `sort` issues.  I didn't see any.

## Action Request 

If this sounds good, I can write up a PR tomorrow with whichever solution makes the most sense.

### DIFF
--- a/test/expect.js
+++ b/test/expect.js
@@ -668,6 +668,15 @@ describe('expect', function () {
 
   });
 
+  it('keys(array) will not mutate array (#359)', function () {
+      var expected = [ 'b', 'a' ];
+      var original_order = [ 'b', 'a' ];
+      var obj = { "b": 1, "a": 1 };
+      expect(expected).deep.equal(original_order);
+      expect(obj).keys(original_order);
+      expect(expected).deep.equal(original_order);
+  });
+
   it('chaining', function(){
     var tea = { name: 'chai', extras: ['milk', 'sugar', 'smile'] };
     expect(tea).to.have.property('extras').with.lengthOf(3);
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -1067,7 +1067,7 @@ module.exports = function (chai, _) {
         ok
       , 'expected #{this} to ' + str
       , 'expected #{this} to not ' + str
-      , expected.sort()
+      , expected.slice(0).sort()
       , actual.sort()
       , true
     );
--- a/test/should.js
+++ b/test/should.js
@@ -525,6 +525,15 @@ describe('should', function() {
     }, "expected { foo: 1, bar: 2 } to not have keys 'foo', or 'baz'");
   });
 
+  it('keys(array) will not mutate array (#359)', function () {
+      var expected = [ 'b', 'a' ];
+      var original_order = [ 'b', 'a' ];
+      var obj = { "b": 1, "a": 1 };
+      expected.should.deep.equal(original_order);
+      obj.should.keys(original_order);
+      expected.should.deep.equal(original_order);
+  });
+
   it('throw', function () {
     // See GH-45: some poorly-constructed custom errors don't have useful names
     // on either their constructor or their constructor prototype, but instead


gregglind:
(updated PR for style.  Hard to find prior art on anons/closures in the tests!)

keithamus:
Fantastic issue @gregglind! I wish every issue was as detailed as this :smile:

test/expect.js
```diff
@@ -602,6 +602,16 @@ describe('expect', function () {
...
+    // issues/359
+    (function fix359() {
+      var expected = [ 'b', 'a' ];
+      var backup = expected.slice(0);
+      var obj = { "b": 1, "a" :1 };
+      expect(expected).deep.equal(backup);
+      expect(obj).keys(expected);
+      expect(expected).deep.equal(backup);
+    })()
+
```
keithamus: Stylistically I think these can just be inlined - I dont think there is a need for a closure. If you want to you could create a new `it()` test instead.

gregglind:
0.  Thank for the compliments! `Chai` is super useful for me on my latest project at Mozilla.  Thanks for getting me to full testing and good practices!

1. Oh jank!  Wrote up several versions of this in comment, then lost the comment!  Ended up thinking that a named function (called immediately after) was better than alternatives (new `it`), because:
    
   - keeps it in the `keys()` test
   - nested `it` doesn't work.
   - no other 'regression' style tests to use as reference.

2.  Happy to go through one more round on this  I can switch it out to a new test if you like.  Mostly it determines how you want to see result when it fails / regresses.

test/expect.js
gregglind: 0.  Thanks for the compliments!  `Chai` is super useful on my most recent project at Mozilla.

1.  Closure it because 
    - it seemed a little gross no matter where I put it.  
    - I don't want to leak any of the vars out, and want to 
    - make it clear it's part of  `keys(array)` testing.

2.  Nested 'it' isn't called, afaict (so sad!), so tried this:

```javascript
  it.only("keys(array), array should keep original order (#359)", function () {
    var expected = [ 'b', "a" ];
    var backup = expected.slice(0);
    var obj = { "b": 1, "a" :1 };
    expect(expected).deep.equal(backup);
    expect(obj).keys(expected);
    expect(expected).deep.equal(backup);
  });
```

It's clearer, but less elegant.  

As an inline alternative.  

```javascript
    var issue359 = {
      original:  ['b', 'a'],
      expected: [ 'b', "a" ],
      obj: { "b": 1, "a" : 1 };
    }

    expect(issue359.expected).deep.equal(issue359.original);
    expect(issue359.obj).keys(issue359.expected);
    expect(issue359.expected, "keys should maintain original order").deep.equal(issue359.backup);

```
Or 

```javascript
    var keys_maintain_original_order_issue359 = function () {
      var expected = [ 'b', 'a' ];
      var original_order = expected.slice(0);
      var obj = { "b": 1, "a" :1 };
      expect(expected).deep.equal(original_order);
      expect(obj).keys(original_order);
      expect(expected, "sorting happened unexpectedly in keys()")
        .deep.equal(original_order);
    }
    keys_maintain_original_order_issue359();
```

Anyway, this is a weird case.  I think I lean toward the last style.  And this is probably more time on this than is really needed :)  (But fun to consider alternatives!)

(pull updated to reflect last)

test/expect.js
```diff
@@ -602,6 +602,18 @@ describe('expect', function () {
...
+      var expected = [ 'b', 'a' ];
+      var original_order = [ 'b', 'a' ];
+      var obj = { "b": 1, "a": 1 };
+      expect(expected).deep.equal(original_order);
+      expect(obj).keys(original_order);
+      expect(expected, "sorting happened unexpectedly in keys()")
+        .deep.equal(original_order);
+    }
+    keys_maintain_original_order_issue359();
+
```
keithamus: Sorry @gregglind, I don't think I made myself clear in the last message. There is no need for these assertions to be made inside a new function scope at all. Line 606, 614 and 615 can be completely removed, so as to keep the code inline with the other assertions - at the same indentation level. Alternatively if you feel like separating the test, it should go in a new `it()` e.g.

```js
it('keys(array) does not mutate array'. function () {
...
})
```

If you do a new `it()` test, it will need to be on the same indentation level as the other `it()`s.

gregglind:
All right, went with new `it` test.  Coverage is still 100%, all tests pass.

keithamus:
Thanks so much @gregglind for sticking with it. Code is perfect now, fantastic job :smile:

PR #361, opened 2015-02-07
Fix #265

### LINKED ISSUE #265: Allowing .keys() to accept Object
```js
describe('myFunction', function() {
  beforeEach(function() {
    this.bar = { baz: stubFn, tah: stubFn };
    this.foo = myFunction(this.bar);
    // >> { baz: [{ callback: stubFn, context: this.bar }], tah: [{ callback: stubFn, context: this.bar }] }
  });

  it('should work with an array', function() {
    expect(this.foo).to.have.keys(Object.keys(this.bar));
  });

  // vs.

  it('should work with an object', function() {
    expect(this.foo).to.have.keys(this.bar);
  });
});
```

gregglind:
1.  We didn't throw on 'too many args' before.  Should we now?  If so, what?
2.  (Do docs run as tests, if so how?)
3.  (Tests inbound.  WIP mostly to claim the bug :) )

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -992,23 +992,36 @@ module.exports = function (chai, _) {
    *     expect({ foo: 1, bar: 2 }).to.have.any.keys('foo');
    *     expect({ foo: 1, bar: 2 }).to.contain.any.keys('bar', 'baz');
    *     expect({ foo: 1, bar: 2 }).to.contain.any.keys(['foo']);
+   *     expect({ foo: 1, bar: 2 }).to.contain.any.keys({'foo': 6});
    *     expect({ foo: 1, bar: 2 }).to.have.all.keys(['bar', 'foo']);
+   *     expect({ foo: 1, bar: 2 }).to.have.all.keys({'bar': 6, 'foo', 7});
    *     expect({ foo: 1, bar: 2, baz: 3 }).to.contain.all.keys(['bar', 'foo']);
+   *     expect({ foo: 1, bar: 2, baz: 3 }).to.contain.all.keys([{'bar': 6}}]);
+   *
    *
    * @name keys
    * @alias key
-   * @param {String...|Array} keys
+   * @param {String...|Array|Object} keys
    * @api public
    */
 
   function assertKeys (keys) {
     var obj = flag(this, 'object')
       , str
-      , ok = true;
+      , ok = true
+      , mixedArgsMsg = 'keys must be given single argument of Array|Object|String, or multiple String arguments';
 
-    keys = keys instanceof Array
-      ? keys
-      : Array.prototype.slice.call(arguments);
+    switch (_.type(keys)) {
+      case "array":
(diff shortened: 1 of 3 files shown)


gregglind:
See #265

lib/chai/core/assertions.js
```diff
@@ -992,23 +992,38 @@ module.exports = function (chai, _) {
...
+      } else {
+        keys = Object.keys(keys);
 
-    keys = keys instanceof Array
-      ? keys
-      : Array.prototype.slice.call(arguments);
+      }
+    } else {
+      keys = Array.prototype.slice.call(arguments);
+    }
```
keithamus: It might read clearer to use the `_.type` function:

```js
if (_.type(keys) === 'object') {
    keys = Object.keys(keys);
} else if (arguments.length > 1) {
    keys = Array.prototype.slice.call(arguments);
}
```

lib/chai/core/assertions.js
```diff
@@ -992,23 +992,38 @@ module.exports = function (chai, _) {
...
   function assertKeys (keys) {
     var obj = flag(this, 'object')
       , str
-      , ok = true;
+      , ok = true
+      , mixedArgsMsg = "keys: if first argument is instanceof Array|Object, must be only argument."
+
+    if (keys instanceof Array || keys instanceof Object) {
+      if (arguments.length > 1) {
+        throw (new Error(mixedArgsMsg));
```
keithamus: Not entirely sure about throwing errors here. Think this may be overcomplicating the method a little is all.

gregglind: The 'throw' was mentioned in the bug, so I added it.  I would think it would be very strange to do `keys(Array, string...)` as dev, but it could happen by accident.  Low-cost to check :)

gregglind:
Went to switch based, using `_.type`.  I think it's very explicit now, for good or ill.

lib/chai/core/assertions.js
```diff
@@ -992,23 +992,36 @@ module.exports = function (chai, _) {
...
   function assertKeys (keys) {
     var obj = flag(this, 'object')
       , str
-      , ok = true;
-
-    keys = keys instanceof Array
-      ? keys
-      : Array.prototype.slice.call(arguments);
+      , ok = true
+      , mixedArgsMsg = "keys: if first argument is instanceof Array|Object, must be only argument."
```
keithamus: I think this error needs rephrasing, maybe to something like `'keys must be given single argument of Array, Object, or multiple String arguments'`

gregglind: Rephased to:  

```
    var mixedArgsMsg = 'keys must be given single argument of Array|Object|String, or multiple String arguments'
```
I think that is accurate.  I do wonder if we should emphasize the 'multiple arguments' part, because that is what triggered the error.

keithamus:
I'd say this is good to go. Let's get it merged in. :+1:

Issue #100, opened 2012-10-02
Assertion.includeStack = true doesn't work on Firefox


jfirebaugh:
Stack trace collection [requires](https://github.com/chaijs/chai/blob/87880253629ab76ad9a1912139a54341514828d1/lib/chai/error.js#L37-L40) `Error.collectStackTrace`, which is a [V8 extension](http://code.google.com/p/v8/wiki/JavaScriptStackTraceApi).

For other browsers, can we maybe throw an Error, immediately catch it, and copy over the stack property?

/cc @ragalie

PR #372, opened 2015-02-19
fix to #100

simonzack:
A wild PR appears!


simonzack:
Think this belongs to assertion-error, closing.

PR #381, opened 2015-02-28
Add a minor precision that empty asserts on strings too.

astorije:


keithamus:
Nice work @astorije :smile:

PR #382, opened 2015-02-28
Add missing docstring for showDiff argument of assert

astorije:
Considering this `assert` method is extensively used when you build your own helper, I think it should be added to the [*Plugin Utilities* API page](http://chaijs.com/api/plugins/).

What do you think?

### DIFF
--- a/lib/chai/assertion.js
+++ b/lib/chai/assertion.js
@@ -88,10 +88,11 @@ module.exports = function (_chai, util) {
    *
    * @name assert
    * @param {Philosophical} expression to be tested
-   * @param {String or Function} message or function that returns message to display if fails
+   * @param {String or Function} message or function that returns message to display if expression fails
    * @param {String or Function} negatedMessage or function that returns negatedMessage to display if negated expression fails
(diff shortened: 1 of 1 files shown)


keithamus:
I think that is a great idea @astorije. Do you want to raise a separate issue for it?

As for this PR. Thanks for your work on it, looks good :smile:.

keithamus:
Hah, never mind about raising the issue - just noticed the other PR :wink:

astorije:
Hey @keithamus! Actually yes, the arguments are described in [the guide](http://chaijs.com/guide/plugins/#basic-assertion).

Do you still think it's worth adding an entry in [the API page](http://chaijs.com/api/plugins/) as well?
If so, should it be at the same level than `.type()`, `.addMethod()`, ... even if they are all methods of `utils` while `assert` is a method of `Assertion`?

Just trying to see where I can put that if I open a PR.

keithamus:
I don't have strong opinions on this - if you see a gap in the docs, fill it to how you'd like to see the docs become. I trust your judgement :smile:

PR #383, opened 2015-03-01
config.truncateThreshold docs: simpler wording

gurdiga:


keithamus:
Thanks @gurdiga :smile: 

LGTM, but don't worry about changing `chai.js` - it gets build every time a release is made - and so shouldn't be edited per commit. I'll merge because it doesn't really matter.

gurdiga:
Noted. :)

PR #385, opened 2015-03-02
fix indexing into array with deep propery

eldritch-fossicker:
* added unit test to expect with code from documentation
* added default to res in _getPathValue

### DIFF
--- a/lib/chai/utils/getPathInfo.js
+++ b/lib/chai/utils/getPathInfo.js
@@ -32,7 +32,7 @@ module.exports = function getPathInfo(path, obj) {
       last = parsed[parsed.length - 1];
 
   var info = {
-    parent: _getPathValue(parsed, obj, parsed.length - 1),
+    parent: parsed.length > 1 ? _getPathValue(parsed, obj, parsed.length - 1) : obj,
     name: last.p || last.i,
     value: _getPathValue(parsed, obj),
   };
@@ -89,7 +89,7 @@ function parsePath (path) {
 
 function _getPathValue (parsed, obj, index) {
   var tmp = obj
-    , res = obj;
+    , res;
 
   index = (index === undefined ? parsed.length : index);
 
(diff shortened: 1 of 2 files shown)


keithamus:
@eldritch-fossicker thanks very much for the PR, but I must admit I can't really tell what this is solving or doing. Could you please elaborate on your explanation with what this aims to fix; perhaps with examples of currently broken code that will be fixed by this PR? Thanks

eldritch-fossicker:
certainly the I had a test that kept failing which was very similar to 
```javascript
expect(deepObj).to.have.property('teas')
  .that.is.an('array')
  .with.deep.property('[2]')
    .that.deep.equals({ tea: 'konacha' });
```
from the chain-able portion of the property documentation trying to walk into the array by index always failed. So, I added the snipped from the documentation to the unit tests. Then I solved the problem by changing `_getPathValue ` to return a value other than undefined to `info.parent` for a path of exactly 1

keithamus:
I guess what is not making sense to me is the fix. It seems to be relying on a side-effect added to `_getPathValue` rather than explicitly tackling the problem. It feel as though the current fix is somewhere non-obvious and may well trip people up when refactoring. To me, it seems better to change [L35](https://github.com/eldritch-fossicker/chai/blob/master/lib/chai/utils/getPathInfo.js#L35) from:

```js
parent: _getPathValue(parsed, obj, parsed.length - 1),
```

to:

```js
parent: parsed.length > 1 ? _getPathValue(parsed, obj, parsed.length - 1) : obj,
```

As this is much more explicit as an indicator for what is going on. What are your thoughts @eldritch-fossicker?

eldritch-fossicker:
@keithamus that will certainly fix the problem. My first conclusion with I looked at the function was that walking path to an index of  0 should return the object itself;  therefore, the function should fixed. I think  a conditional return or comment would be in order. But I'm just inserted in fixing the error whichever solution.

keithamus:
I see where you're coming from now @eldritch-fossicker. Ultimately it's neither here nor there - but the original patch seems slightly less explicit. On your point about if it were called with an index of 0 - I'd suggest that it shouldn't be called at all if that is the case.

If you could amend the patch to my reflect my comments above I'd be very happy at merging this though :smile:

keithamus:
:tada:

PR #397, opened 2015-03-10
Add missing parameter showDiff and make docstring "documentationable"

astorije:
I discussed some time ago about this on #382 with @keithamus.

I couldn't find a documentation on how to make a docstring appear on the doc site, so I made the assumption that docstrings starting with `/*!` were ignored and those starting with `/**` were added to the web pages. Please make fun of me if I am completely mistaken.

This goes along with chaijs/chai-docs#80.

### DIFF
--- a/chai.js
+++ b/chai.js
@@ -845,8 +845,8 @@ module.exports = function (_chai, util) {
     util.overwriteChainableMethod(this.prototype, name, fn, chainingBehavior);
   };
 
-  /*!
-   * ### .assert(expression, message, negateMessage, expected, actual)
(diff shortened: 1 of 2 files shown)


keithamus:
@astorije you are indeed correct. FYI the docs are written in [dox](https://github.com/tj/dox) which does indeed ignore `/*!`. 

Unsure why these were initially written as `/*!` - so I might just let @logicalparadox merge this one as he might have a specific reason for using `/*!`

astorije:
Fair enough, thanks for the update!

Issue #398, opened 2015-03-10
.deep.property('https://schema.org/mentions[0]') is not resolved


oncletom:
I have the following object structure:

```json
{
  "https://schema.org/mentions": [
    "http://dbpedia.org/resource/London_Daily_News",
    "http://dbpedia.org/resource/Jonathan_Aitken"
  ]
}
```

I then run the following assertion:

```js
expect(dataStructure).to.have.deep.property('https://schema.org/mentions[0]', 'http://dbpedia.org/resource/London_Daily_News');
```

An exception is raised:

```
AssertionError: expected { Object (https://schema.org/mentions) } to have key 'https://schema.org/mentions'
```

If I change the `https://schema.org/mentions` key to `mentions` only, the assertion works.

keithamus:
Hi @oncletom thanks for the issue.

`deep.property` allows for object syntax which means using `.` and `[]` syntaxes to get into properties.

Using a URL is not compatible as URLs have `.`s in them. So your lookup is broken down into:

```js
['https://schema', 'org/mentions', 0]
```

We could attempt to implement escaping of the control characters (`.`/`[`/`]`), probably using a `\` but this feels messy, you should probably just avoid using `.deep` for these circumstances - so your assertion could become:

```js
expect(dataStructure).to.have.property('https://schema.org/mentions').have.property(0, 'http://dbpedia.org/resource/London_Daily_News');
```

I'm going to close this issue - as I feel like the above will solve your problem. If you still really want to use `.deep.property` - please file a new issue about escaping control characters in deep property, and we can have a full discussion about it there.

If you have any problems further to this issue, feel free to comment here or raise a new issue.

oncletom:
Thanks for your feedback. This is how I ended up anyway. I just was not sure if I did something wrong or not.

So :+1:

keithamus:
Hey @oncletom I've got some good news! #402 enabled escaping of .deep.properties, so you can now do:

```js
expect(dataStructure).to.have.deep.property('https://schema\\.org/mentions[0]', 'http://dbpedia.org/resource/London_Daily_News')
```

oncletom:
You rock @keithamus :-) thanks for the follow-up!

PR #402, opened 2015-03-17
Escaping dot should be taken in deep property

umireon:
According to [getPathInfo.js] (https://github.com/chaijs/chai/blob/b6e1b573e07d108b4ecd6c9da733f20f010986a8/lib/chai/utils/getPathInfo.js#L65), `deep.property(name)` is assumed to support escaping dot `.` by prepending backslash `\`. However, escaping dot in `deep.property` seems to work in the unexpected way. Here is the example:

```
> expect({"foo.bar": "baz"}).to.have.deep.property("foo.bar", "baz");
AssertionError: expected { 'foo.bar': 'baz' } to have a deep property 'foo.bar'
// fail expectedly

> expect({"foo.bar": "baz"}).to.have.deep.property("foo\\.bar", "baz");
AssertionError: expected { 'foo.bar': 'baz' } to have a deep property 'foo\\.bar'
// fail unexpectedly

> expect({"foo\\.bar": "baz"}).to.have.deep.property("foo\\.bar", "baz");
// success unexpectedly
```

The current implementation of `deep.property`never match `{"foo\\.bar": "baz"}` because backslash `\` is not removed on processing escape. This PR fixes this problem. In addition, it includes the test case to ensure the problem fixed.

Noted that this PR obviously breaks some backward compatibility.

### DIFF
--- a/lib/chai/utils/getPathInfo.js
+++ b/lib/chai/utils/getPathInfo.js
@@ -67,7 +67,7 @@ function parsePath (path) {
     var re = /\[(\d+)\]$/
       , mArr = re.exec(value);
     if (mArr) return { i: parseFloat(mArr[1]) };
-    else return { p: value };
+    else return { p: value.replace(/\\\./g, '.') };
   });
 }
 
--- a/lib/chai/utils/getPathInfo.js
+++ b/lib/chai/utils/getPathInfo.js
@@ -54,6 +54,7 @@ module.exports = function getPathInfo(path, obj) {
  *
  * * Can be as near infinitely deep and nested
  * * Arrays are also valid using the formal `myobject.document[3].property`.
+ * * Literal dots and brackets (not delimiter) must be backslash-escaped.
  *
  * @param {String} path
  * @returns {Object} parsed
@@ -61,14 +62,15 @@ module.exports = function getPathInfo(path, obj) {
  */
 
 function parsePath (path) {
-  var str = path.replace(/\[/g, '.[')
-    , parts = str.match(/(\\\.|[^.]+?)+/g);
-  return parts.map(function (value) {
-    var re = /\[(\d+)\]$/
-      , mArr = re.exec(value);
-    if (mArr) return { i: parseFloat(mArr[1]) };
-    else return { p: value.replace(/\\\./g, '.') };
-  });
+  /* assumed /^(?:(?:\\[.\[\]]|[^.\[\]])+|\[\d+\])(?:\.(?:\\[.\[\]]|[^.\[\]])+|\[\d+\])$/ */
+  var re = /(?:\\[.\[\]]|[^.\[\]])+|\[(\d+)\]/g
+    , parsed = []
+    , value;
+  while ((value = re.exec(path)) !== null) {
+    if (value[1]) parsed.push({ i: parseFloat(value[1]) });
+    else parsed.push({ p: value[0].replace(/\\([.\[\]])/g, '$1') });
+  }
+  return parsed;
 }
 
 
--- a/lib/chai/utils/getPathInfo.js
+++ b/lib/chai/utils/getPathInfo.js
@@ -62,15 +62,14 @@ module.exports = function getPathInfo(path, obj) {
  */
 
 function parsePath (path) {
-  /* assumed /^(?:(?:\\[.\[\]]|[^.\[\]])+|\[\d+\])(?:\.(?:\\[.\[\]]|[^.\[\]])+|\[\d+\])$/ */
-  var re = /(?:\\[.\[\]]|[^.\[\]])+|\[(\d+)\]/g
-    , parsed = []
-    , value;
-  while ((value = re.exec(path)) !== null) {
-    if (value[1]) parsed.push({ i: parseFloat(value[1]) });
-    else parsed.push({ p: value[0].replace(/\\([.\[\]])/g, '$1') });
-  }
-  return parsed;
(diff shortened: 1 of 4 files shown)


keithamus:
Thanks for the PR @umireon!

While it does strictly break backwards compatibility - I'd be shocked if anyone was bitten by this - so I'm on the fence whether or not to merge it here or wait for 3.x.x. Having said that - as you mentioned, trying to escape results in wild behaviour so likely if people have seen this, they've seen bugged out behaviour - so I'd consider this a bug fix.

Irregardless, a few notes:

 - [ ] Documentation is missing for the new escaping "feature" - if you could add documentation to `.deep.property` about this that'd be great.
 - [ ] If we're escaping `.`s we should probably also escape `[]`s. Would you be happy to implement `[]` escaping too?

umireon:
Hi, @keithamus !
I've written some documentations and implemented escaping `[]`.

In fact, the path strings which the current implementation of  `.deep.property` accepts is very wide.
I assumed the input of `.deep.property` matches /^(?:(?:\\[.\[\]]|[^.\[\]])+|\[\d+\])(?:\.(?:\\[.\[\]]|[^.\[\]])+|\[\d+\])$/ to keep the implementation clean.
Therefore, any path string which does not may match the expression results in an unexpected way.

I'm looking forward to hearing from you,

deep-path

lib/chai/utils/getPathInfo.js
```diff
@@ -61,14 +62,15 @@ module.exports = function getPathInfo(path, obj) {
...
-    var re = /\[(\d+)\]$/
-      , mArr = re.exec(value);
-    if (mArr) return { i: parseFloat(mArr[1]) };
-    else return { p: value };
-  });
+  /* assumed /^(?:(?:\\[.\[\]]|[^.\[\]])+|\[\d+\])(?:\.(?:\\[.\[\]]|[^.\[\]])+|\[\d+\])$/ */
+  var re = /(?:\\[.\[\]]|[^.\[\]])+|\[(\d+)\]/g
+    , parsed = []
+    , value;
+  while ((value = re.exec(path)) !== null) {
```
umireon: I need to use `while` instead of `Array#map` to avoid matching /\[(\d+)\]/ twice.

keithamus:
Well, `.[]` isnt officially supported - it was more a by-product of implementation, and I'm not sure its something we need to explicitly support (especially as its not familiar syntax for javascript). Would that not simplify parsing a lot?

umireon:
I'm sorry if I didn't get your point, but you mean it is nice to implement
escaping using shorter regular expression with modifying the string?
2015/03/19 7:21 "Keith Cirkel" <notifications@github.com>:

keithamus:
Pretty much yes. The regexp seems a little complex - and unless I'm mistaken a fair chunk of the complexity comes from detecting `.` vs `\\.` vs `.[]` vs `\\.[]` vs `.\\[]`, but in reality we just need to find `.`, `[]` and escape on `\\.` and `\\[]`.

Please correct me if I'm wrong though.

umireon:
Frankly speaking, I'm not confident to keep the behavior unchanged with parsing the regular language (= the simple syntax of `.deep.property`'s name) by the mechanism more complex than regular expression.
Regular expression is so simple syntax that it can easily be proved to match the desired string.
Therefore, I unintentionally wrote the code whose everything is in one regular expression but it is not concern of this PR.
Finally, it's my fault to mangling the regular expression in this PR, very sorry :cry:.

lib/chai/utils/getPathInfo.js
```diff
@@ -61,13 +62,13 @@ module.exports = function getPathInfo(path, obj) {
  */
 
 function parsePath (path) {
-  var str = path.replace(/\[/g, '.[')
+  var str = path.replace(/([^\\])\[/g, '$1.[')
```
umireon: Without (`([^\\])`, `$1`), `a\\[`results in accessing `a.[`.

lib/chai/utils/getPathInfo.js
```diff
@@ -61,13 +62,13 @@ module.exports = function getPathInfo(path, obj) {
...
     , parts = str.match(/(\\\.|[^.]+?)+/g);
   return parts.map(function (value) {
-    var re = /\[(\d+)\]$/
+    var re = /^\[(\d+)\]$/
```
umireon: This `^` ensures `test\\[1]` parsed as the property `test[1]`, not the index `1`

umireon:
Actually, `]` don't have to be escaped.

umireon:
> The regexp seems a little complex - and unless I'm mistaken a fair chunk of the complexity comes from detecting . vs \\. vs .[] vs \\.[] vs .\\[], but in reality we just need to find ., [] and escape on \\. and \\[].
The expression was complex simply because three regular expression and memory rewriting was combined.

keithamus:
@umireon that now looks much clearer to me. Good job :smile: 

I'd like to pester you for one more bit:

 - [ ] Add documentation for `.deep.property` escaping in the `deep` flag docs ([assertion.jsL68-80](https://github.com/umireon/chai/blob/escaping-dot-should-be-taken/lib/chai/core/assertions.js#L68-L80)).

Once that's done I'm happy to wrap up this PR :smile:

umireon:
@keithamus I've made it done!

keithamus:
Great work @umireon! :tada:

keithamus:
Linking #398

PR #403, opened 2015-03-18
Assert ifError now throws the error passed in, to keep it consistent wit...

### LINKED ISSUE #369: ifError should throw its argument
Why `ifError(arg)` doesn't throw its argument as node.js' included assert module do?

cjqed:
...h node's assert.ifError

Resolves #369 for 3.x.x


keithamus:
Good work @cjqed :smile:

PR #405, opened 2015-03-20
Tweak documentation on `.deep` flag.

keithamus:
Hopefully @umireon won't mind but I've slightly tweaked the documentation on the `.deep` flag regarding escapes (#402). 

This adds an example of escaping, just like `.deep.property` does - to save people jumping around the docs.


umireon:
@keithamus Thanks for improvement! :cake:  I'm not so good at documentation.

Issue #416, opened 2015-04-03
Message of a CustomError-Class not correctly checked if thrown without arguments


Charminbear:
If you define a custom error class and then throw it without arguments, `.throws()` will not check the message property.

See the following, passing test:
```JavaScript
it('Custom Error Message is not checked if called without arguments', function () {
    // CREATE A CUSTOM ERROR CLASS
    var myCustomError = function (message) {
        this.name = 'myError';
        this.message = message;
    };
    require('util').inherits(myCustomError, Error);
    expect(require('util').isError(new myCustomError())).to.be.true; // PASSES

    var throwing = function () {
	// THROW ERROR WITHOUT ARGUMENTS
        throw new myCustomError();
    };
    expect(throwing).to.throw(myCustomError, 'MyMessage'); // UNEXPECTEDLY PASSES
});
```

It only seems to behave like this as long as the message property is `null` or `undefined`. If I change the constructor to `this.message = message || ''` or throw the Error with an empty String (`throw new myError('')`, it fails as expected.

Is this a bug or am I doing something wrong?

keithamus:
@Charminbear do you think it is reasonable to close this given your 3.x.x PR is merged? I aim to release 3.x at some point this month, as I don't want to keep it hanging around too long.

Charminbear:
@keithamus yes of course. No sense in putting work in fix thats obsolete in 3 weeks. :D

Im excited for 3.x.x :+1:

PR #418, opened 2015-04-04
Replace type.js with the "chaijs/type-detect" library

Charminbear:
Hi Guys,

i finally found the time to create this pull request, moving chai.js to use type-detect. 

I had to change one minor line within `assertions.assertThrows`, as *type-detect* actually returns the type `'error' ` for a correctly inherhited Custom Error, not `'object'`. Funny enough, this also fixes the other bug-issue i reported [here](https://github.com/chaijs/chai/issues/416).

You still want to be able to test thrown 'normal' Objects with a message-property for their messages? In this case, i would add another `or`-Condition like:
```JavaScript
('object' === _.type(err) || 'error' === _.type(err)) && "message" in err
```


keithamus:
Looks good @Charminbear!

I'm a tiny bit concerned about the `'object'`/`'error'` thing. Especially as I fear some users might have some code like `expect(error).to.be.an('object')`. This could be enough to push this change into `v3.0.0`, we don't want people upgrading chai and breaking tests (unless they upgrade to a new major version of course).

What are your thoughts?

keithamus:
Also @Charminbear could you please update the docs for `an()` with this PR. 

https://github.com/chaijs/chai/blob/master/lib/chai/core/assertions.js#L127-L148

It would be good to add some examples of how to detect new primitives, like Promises, typed arrays, DataViews etc, as well as an example of how to override Symbol.toStringTag to check your own types. Something like:

```js
  /**
   * ### .a(type)
   *
   * The `a` and `an` assertions are aliases that can be
   * used either as language chains or to assert a value's
   * type.
   *
   *     // typeof
   *     expect('test').to.be.a('string');
   *     expect({ foo: 'bar' }).to.be.an('object');
   *     expect(null).to.be.a('null');
   *     expect(undefined).to.be.an('undefined');
   *     expect(new Promise).to.be.a('promise');
   *     expect(new Float32Array()).to.be.a('float32array');
   *
   *     // es6 overrides
   *     expect({[Symbol.toStringTag]:()=>'foo'}).to.be.a('foo');
   *
   *     // language chain
   *     expect(foo).to.be.an.instanceof(Foo);
   *
   * @name a
   * @alias an
   * @param {String} type
   * @param {String} message _optional_
   * @api public
   */
```

Charminbear:
Hm, good Point @keithamus, i didnt even think that far. If we go strict by the semver rules this would be a breaking change, so yeah, i guess it should be in 3.0.0. Would also be the safer route to take as you said. Even though i think checking your error to be an object is not the most likely use-case, but I guess that's not the point. ;)
You want me to move this PR to the 3.x.x branch then?

Oh and will do the docs-update.

keithamus:
Looks great :smile:. But yes I think it may be best as a breaking change - therefore if you could put it in 3.0.0 branch that'd be swell.

PR #423, opened 2015-04-07
Fix a small inconsistency in the .null assertion docs

ehntoo:
The other assertions for undefined and the like all use '.to.not' in their examples - The slight increase in consistency is nice. :)


keithamus:
:+1:

Issue #427, opened 2015-04-14
Length() should be removed from documentation


matthijsgroen:
`expect('something').to.have.length(9)`

and

`expect('something').to.have.lengthOf(9)`

are the same... on paper

but consider this test:

```
describe 'overriding properties', ->
  beforeEach ->
    global.expect = require("chai").expect
    global.when = require("when")

  it 'works without overrides', ->
    expect('something').to.have.length(9)

  it 'breaks for each property', ->
    chai = require "chai"
    chai.Assertion.addChainableMethod 'have', ->
    expect('something').to.have.length(9) # Fails with: TypeError: Property 'length' of object [object Function] is not a function

  it 'breaks the length method', ->
    chai = require "chai"
    chai.Assertion.addChainableMethod 'to', ->
    expect('something').to.have.length(9) # Fails with: TypeError: Property 'length' of object [object Function] is not a function

  it 'does not break the lengthOf method', ->
    chai = require "chai"
    chai.Assertion.addChainableMethod 'to', ->
    expect('something').to.have.lengthOf(9)
```

The breaking change is in: https://github.com/chaijs/chai/blob/master/lib/chai/utils/addChainableMethod.js#L79

A function object is constructed, and the object has a `length` property of its own, and thus overriding the length method the original asserter had.

`console.log (->).length # ==> 0`

`lengthOf` is not a magical javascript property and is left alone. We should encourage people to use `lengthOf` instead of `length` in the syntax of their tests.

valscion:
:+1: For removing `length` — this is some really funky behaviour.

keithamus:
I concur. We'll need to keep it around for backwards compatibility. Maybe we'll do this:

 - v.next: Remove docs
 - v.next+1: Log deprecation warning when accessing .length()
 - v.major: Remove .length()

Open to PRs removing docs for now.

valscion:
Great! Wouldn't it be more friendly for chai users to tell about the deprecation in the docs instead of removing the mention altogether?

keithamus:
Sure sounds like a better plan.

valscion:
I'll quickly draft a PR for that now.

valscion:
Hey @keithamus, I noticed that even using `length` as a chain precursor fails:

```coffee
describe 'overriding properties', ->
  beforeEach ->
    global.expect = require("chai").expect
    global.when = require("when")

  it 'works without overrides', ->
    expect('something').to.have.length(9)

  it 'breaks for each property', ->
[...]

PR #429, opened 2015-04-16
Add .length(value) deprecation warning to documentation

valscion:
Fixes https://github.com/chaijs/chai/issues/427

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -938,16 +938,10 @@ module.exports = function (chai, _) {
   Assertion.addMethod('haveOwnPropertyDescriptor', assertOwnPropertyDescriptor);
 
   /**
-   * ### .length(value)
+   * ### .length
    *
-   * Asserts that the target's `length` property has
-   * the expected value.
-   *
-   *     expect([ 1, 2, 3]).to.have.length(3);
-   *     expect('foobar').to.have.length(6);
-   *
-   * Can also be used as a chain precursor to a value
-   * comparison for the length property.
+   * Sets the `doLength` flag later used as a chain precursor to a value
+   * comparison for the `length` property.
    *
    *     expect('foo').to.have.length.above(2);
    *     expect([ 1, 2, 3 ]).to.have.length.above(2);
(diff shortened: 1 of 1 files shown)


chai.js
```diff
@@ -1847,16 +1847,59 @@ module.exports = function (chai, _) {
   Assertion.addMethod('haveOwnProperty', assertOwnProperty);
 
   /**
-   * ### .length(value)
+   * ### .ownPropertyDescriptor(name[, descriptor[, message]])
```
valscion: The built source seems to create a funky diff — I haven't touched the `ownPropertyDescriptor` matcher at all.

keithamus:
Sorry @valscion - I should have mentioned you don't need to update chai.js - it will be updated as part of the release cycle. Could you pretty please remove it from the commit?

lib/chai/core/assertions.js
```diff
@@ -956,8 +950,25 @@ module.exports = function (chai, _) {
    *     expect('foo').to.have.length.within(2,4);
    *     expect([ 1, 2, 3 ]).to.have.length.within(2,4);
    *
+   * *Deprecatin notice:* Using `length` as an assertion will be deprecated
+   * in version 2.2.1 and removed in 2.3. Code using the old style of
```
keithamus: Next release will be `2.3.0`, and we'll have to remove it in `3.0.0` to keep semver. So could you please change this message to:

```js
   * *Deprecation notice:* Using `length` as an assertion will be deprecated
   * in version 2.3.0 and removed in 3.0.0. Code using the old style of
```

(Also you had a typo in `Depricatin`(sic))

valscion:
> Sorry @valscion - I should have mentioned you don't need to update chai.js - it will be updated as part of the release cycle.

Ah, it wasn't so clear based on `CONTRIBUTING.md` so I figured I'd rather do it first and then ask later whether it was ok or not :smile:

As for the deprecation notice text, I changed them to what you suggested.

valscion:
Hey @keithamus, is everything OK now?

keithamus:
Sorry @valscion this one slipped by me. Version is a number out now, but I'll fix that as it's my bad. All looks good, thanks for your hard work :smile:.

PR #439, opened 2015-05-04
Add Assertion alias: matches => match

### LINKED ISSUE #438: .match() aliased to .matches()
So it can be written like this:

```js
chai.expect({ foo: 'bar' }).to.have.property('foo').that.matches(/^bar$/)
```

thejameskyle:
Resolves #438


keithamus:
Don't worry about the failing build, it's something else. Tests pass on my machine, so consider this merged :smile:

Issue #371, opened 2015-02-18
breaking change in 2.0 (null, undefined, arguments, ok, true, false, empty, exist)


BigstickCarpet:
Before the 2.0 release of Chai, these assertions could be written as properties or as functions.  For example:
````javascript
expect(foo).to.be.true;
expect(foo).to.be.true();

expect(foo).to.be.empty;
expect(foo).to.be.empty();

expect(foo).to.exist;
expect(foo).to.exist();

expect(foo).to.be.null;
expect(foo).to.be.null();
````
The documentation on [chaijs.com](http://chaijs.com/) even explained that this dual-syntax was there to play nice with some IDEs and linting tools that flag the non-function syntax as an error.   But in 2.0, the function syntax no longer exists.   Was it removed by accident, or was there a reason for it?

adrianblynch:
I see the same. Our test suite was updated to use function call syntax and under 2.0.0 they all fail.

Include `undefined()` and `false()` in that list.

adrianblynch:
Look at the bottom of the [2.0.0 change log](https://github.com/chaijs/chai/blob/9c10c9f03e71a38964e0bdc347866ca5455c8712/History.md).

`Revert "Allows writing lint-friendly tests"`

Was this intentional?

keithamus:
Thanks for the issue @BigstickCarpet. 

This was fully intentional. The new method assertion style (`.to.be.true()`) broke the old property assertion style (`.to.be.true`), meaning it was a case of using one or the other. We had to pick one single interface to stick to, so we went back to the old property assertion style (`.to.be.true`).

Because developers have been using the new style, and dropping the new style would break their code, we pushed the change as a Version 2.0.0. According to [SemVer](http://semver.org/) (which I'd like us to start following for future releases), major versions mean breaking changes - which is to say that the contract of "you wont have to refactor your test code" is invalidated. 

If you'd like to keep the method assertion syntax (`.to.be.true()`), feel free to keep using Chai 1.10.0. Alternatively, if you feel the need to upgrade to 2.0.0 then you have a few options:

  - Refactor your tests to the new (actually old) property assertions; `.to.be.true`, `to.be.empty`, etc.
  - Refactor your code so as not to avoid these assertions (`.to.be.true` becomes `.to.equal(true)`) and avoid the problem altogether
[...]

PR #440, opened 2015-05-06
Adds assertions .to.be.defined

solsson:
There is of course `.to.exist` and `.to.be.ok` but since #371 we can no longer recommend that developers end every expectation with () which means we do get no-op expectations now and then. Test-first is a solution of course, but there's also test maintenance.

I didn't understand the rationale for the revert in #306 but I suppose there is good reasons. I think the use of .defined nearby .undefined is the most common mistake when asserting without ().

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -326,6 +326,29 @@ module.exports = function (chai, _) {
   });
 
   /**
+   * ### .defined
+   *
+   * Asserts that the target is not `undefined`.
+   *
+   *     expect(undefined).to.be.defined;
+   *     expect(null).to.not.be.defined;
+   *
+   * Needed because it is intuitive to invert .undefined,
+   * which would always pass if .defined is nonexistent.
+   *
+   * @name undefined
+   * @api public
+   */
+
+  Assertion.addProperty('defined', function () {
+    this.assert(
+        undefined !== flag(this, 'object')
+      , 'expected #{this} to be defined'
+      , 'expected #{this} not to be defined'
+    );
+  });
+
+  /**
    * ### .exist
    *
    * Asserts that the target is neither `null` nor `undefined`.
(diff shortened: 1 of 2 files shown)


keithamus:
Thanks for the PR @solsson.

Firstly, there's a pretty comprehensive history of the whole property/method assertion syntax [here](https://github.com/chaijs/chai/issues/371#issuecomment-75034361).

I must admit I'm a little dubious to include `.to.be.defined`, as we already have `.to.exist` (as you suggest). If this is simply to try to alleviate a need for `()` at every statement, I'd suggest just using the excellent chaining feature that chai has. I personally have linting rules to enforce `()`, but still happily use should - I'm just a little more careful about how I use it:

```js
expect(foo).to.exist.and.equal(1);
expect(bar).to.exist.and.be.above(3);
expect(baz).to.exist.and.be.a('function');
```

Usually I find if I'm asserting something exists, I can easily make slightly more specific assertions about it too.

I'll mark this as `more-discussion-needed` to give you and others a chance to make a case for adding this, but my vote is no for this addition.

solsson:
We do have the same issue that @feugy points out in #371. But we've continued to use chai expect style and with your advice on chaining we do get the advantages of `()`. Misspelling is only a problem with `.be.undefined` where chaining is useless.

keithamus:
These are the same assertions:

```js
expect(foo).to.be.undefined
expect(foo).to.not.exist
expect(foo).to.equal(undefined);
```

Feel free to use the one with `()`, which jshint won't complain about.

I'm still not sold on `.to.be.defined` - I'm going to close this because of the reasons I mentioned above, but rest assured that I'm also putting thought into how we can alleviate the jshint problems, as this issue is still coming up a lot. For now, you should use [dity-chai](https://www.npmjs.com/package/dirty-chai) if you really want method style assertions.

keithamus:
Also @solsson, please don't let this put you off filing issues and making future PRs. It might be best to discuss your ideas as issues before turning into PRs though, as I'd hate you to feel as though your time is wasted.

Issue #442, opened 2015-05-08
chai can't cope with AssertionErrors raised from node's assert


danielbprice:
I'm opening a new issue for the problem described in
https://github.com/chaijs/chai/issues/206#issuecomment-33762758

Here is a sample coffeescript program which illustrates this problem:

```coffeescript
node_assert = require("assert")
chai = require("chai")

expect = chai.expect

foo = () ->
    node_assert("blue" is "red")

expect(() -> foo()).to.throw(node_assert.AssertionError)
```

When run this produces:

```
TypeError: Cannot read property 'actual' of undefined
  at new AssertionError (assert.js:42:24)
  at Assertion.assertThrows (/t/node_modules/chai/lib/chai/core/assertions.js:1052:17)
  at Assertion.ctx.(anonymous function) [as throw] (/t/node_modules/chai/lib/chai/utils/addMethod.js:40:25)
  at Object.<anonymous> (/t/foo.coffee:10:1)
  at Object.<anonymous> (/t/foo.coffee:2:1)
  at Module._compile (module.js:456:26)
```

However, I would expect this to pass properly.  Please see the referenced comment from
@freeformsystems for some investigation into why this is happening.  He also proposed a fix
there, as well.

keithamus:
The fix is a little too specific for me. 

I think a much simpler fix for this would be to simply drop the XOR assignment to `name`. So where we set name ([assertion.js#L1219-1222](https://github.com/chaijs/chai/blob/master/lib/chai/core/assertions.js#L1219-L1222)) we currently have

```js
      name = constructor.prototype.name || constructor.name;
      if (name === 'Error' && constructor !== Error) {
        name = (new constructor()).name;
      }
```

which could become

```
      name = constructor.prototype.name;
      if (!name || (name === 'Error' && constructor !== Error)) {
        name = constructor.name || (new constructor()).name;
      }
```

This solves the specific problem you want to address, while also keeping away from requiring in Node's assert, meaning browser friendliness :smile:.

Happy to see a PR for this (with some tests to back it up of course).

danielbprice:
@keithamus, I can confirm that the fix you designed works perfectly for this case.  I applied it and then tried to invent some test cases for it, but we're a little too deep into javascript innards that I don't feel familiar with.  I tried building out my own class called AssertionError but I couldn't get it to trip in the same manner as node's AssertionError.

keithamus:
[...]

PR #446, opened 2015-05-15
Issue #442 chai can't cope with AssertionErrors raised from node's assert

danielbprice:
I'm happy to implement some test cases for this PR, but despite several tries I had trouble figuring out how to mock up the exact scenario we hit with AssertionError in nodejs.  If there's a way to implement nodejs specific tests, I can do that too.


keithamus:
:+1: sorry it took so long to merge @danielbprice

PR #451, opened 2015-05-27
Switch to use Browserify for builds

### LINKED ISSUE #362: Self Named Module is not good
First off, let me say that you guys are doing a great job with this project.  I find it really invaluable.

I just went to upgrade to the latest version (1.10) and things broke.  I eventually traced the problem to [here](https://github.com/chaijs/chai/blob/1.10.0/chai.js#L4796) in that the define names the module.

I'm not sure why the change happened but it breaks projects that include a paths configuration in their requireJS (or equivalent) config.

The guys at MomentJS ran into the exact same [issue](https://github.com/moment/moment/issues/1095) last year.  James Burke himself chimed in to say that this was not a good thing for lib modules to do and explains why.  In the end, they ended taking it out and going back to defining an anonymous module.

Please have a look at the thread of comments for their issue and considering doing the same for the same reasons already discussed.  Thanks.

csnover:
Fixes #362.

Refs theintern/intern#400. This is a blocker issue for us so any prompt release would be very much appreciated.


keithamus:
Hey @csnover thanks very much for doing this. I'll look into merging this and releasing with other changes in the `3.x.x` branch shortly - so `3.0.0` will be released in a day or two.

csnover:
Awesome, thanks!

sholladay:
@keithamus friendly *poke* here :)

keithamus:
Sorry, been busy the last few days. Was meant to release over the weekend but other stuff came up. Will cut a release now.

PR #458, opened 2015-06-01
Remove stray character in `assert.notInclude` docs

BinaryMuse:
This small error was causing a display issue on the site; namely, the examples for `assert.notInclude` did not render properly:

chai docs


keithamus:
Good catch :+1:

PR #468, opened 2015-06-12
Remove moot `version` property from bower.json

kkirsche:
Per bower/bower.json-spec@a325da3

Also their maintainer says they probably won't ever use it: http://stackoverflow.com/questions/24844901/bowers-bower-json-file-version-property

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ chai.js: node_modules lib/*
 #
 
 define release
-	./node_modules/.bin/bump -y --$(1) *.json lib/chai.js
+	./node_modules/.bin/bump -y --$(1) package.json lib/chai.js
 	make chai.js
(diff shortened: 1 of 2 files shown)


keithamus:
Thanks for the PR @kkirsche!

If we're going to merge this, you'll also need to exclude it from the Makefile bump utility on [Makefile#L24](https://github.com/chaijs/chai/blob/master/Makefile#L24), so change the line to something like

```bash
	./node_modules/.bin/bump -y --$(1) package.json lib/chai.js
```

from 

```bash
	./node_modules/.bin/bump -y --$(1) *.json lib/chai.js
```

kkirsche:
@keithamus pushed that change for you

keithamus:
:+1:

Issue #472, opened 2015-06-28
should undefined be classed as empty in expect empty  


Daveloper87:
I have found a potential issue with the .empty assertion in expect. Running in a node.js context on Mac OSX
If you run :
```
expect().to.be.empty
expect(undefined).to.be.empty;
```
This will happily work as a passing test assertion. Is this an intended behavior?
My only problem would be if you are creating a collection dynamically and it comes through as undefined, you may assume your code is working when it is just returning undefined.

Obviously you could add something like:
```
expect(someData).to.be.an('object');
```
But I feel it should fail if it receives undefined to avoid assertions passing when no collection, object or string was passed.

Also as a side note, the string check here does not execute:
https://github.com/chaijs/chai/blob/master/lib/chai/core/assertions.js#L379

It still works because the code runs
```
!expected
```
 which would return true for an empty string. It looks like it supposed to be obj instead of object
```
 if (Array.isArray(obj) || 'string' === typeof obj)
```
 The 'string' check could be removed based on !expected, but it is probably nicer to be explicit anyway.

keithamus:
Hi @Daveloper87 thanks for the issue!

You're right on the latter - it should be `obj`. I welcome a pr fixing this.

As for `.empty `, changing it would be a breaking change for something that is philosophically/arguably wrong, rather than objectively. Therefore I'm not as convinced. If you're concerned that your code may be returning `undefined` then add some more assertions to test this, as you mentioned. Personally I always assert return types for all my tests unless it's obvious (e.g through equality checks).

keithamus:
Just to clarify - I'm marking this as pr wanted only for the `obj` fix :smile:.

Daveloper87:
Hi @keithamus , Thanks for responding. I have opened a pull request for that fix here: 
https://github.com/chaijs/chai/pull/473

In regards to the .empty on undefined checks. I only actually noticed it because I was writing my tests first, and was slightly surprised that my empty function passed, but I very much agree with your point about checking the return type and it not being worth a breaking change, I just did not expect undefined to pass the empty check. 

One little thing I wanted to note as well just in case someone else comes across this issue.
```
expect(null).to.be.empty;
```
only fails because the line: https://github.com/chaijs/chai/blob/master/lib/chai/core/assertions.js#L381
```
else if (typeof obj === 'object') {
```
because typeof null is === 'object',   it just fails with a TypeError trying to Object.keys on null, otherwise that would pass too.

[...]

PR #473, opened 2015-06-28
corrected var name to 'obj' instead of 'object' - linked to issue472

Daveloper87:
- Corrected condition so that empty string is getting checked as intended
- Raised in issue #472

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -376,7 +376,7 @@ module.exports = function (chai, _) {
     var obj = flag(this, 'object')
       , expected = obj;
 
-    if (Array.isArray(obj) || 'string' === typeof object) {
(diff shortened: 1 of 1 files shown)


keithamus:
Now that I look at the code - I wonder if the whole thing could just be shortened to `Object.keys(obj).length === 1`. Strings and Arrays will both have keys extracted by this - and it takes the whole assertion to a one liner.

Do you fancy making this change @Daveloper87? I'd expect the final code to look something like:

```js
     this.assert(
         Object.keys(flag(this, 'object')).length > 0
       , 'expected #{this} to be empty'
       , 'expected #{this} not to be empty'
     );
   });
```

Daveloper87:
Closing this to create new pull request against 4.x.x

PR #474, opened 2015-06-29
Empty fail undefined

Daveloper87:
- Added handling of undefined and null to fail if passed in

@keithamus Is chai 4.x.x going to be focused towards ES6? If so I can use Object.keys() to coerce primitives.

Also there is the option to do this:
```
else if (obj.hasOwnProperty('length')) {
  expected = obj.length === 0;
 }
```
Which would cover any new objects with length, but it would add support for checking if a function takes parameters or not (not sure that's what you would want from an empty check..).

Just an idea. I will wait for your feedback.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ chai.js: node_modules lib/*
 #
 
 define release
-	./node_modules/.bin/bump -y --$(1) *.json lib/chai.js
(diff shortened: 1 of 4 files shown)


Daveloper87:
Sorry. Branched off master... Will reopen from 4.x.x.

Issue #457, opened 2015-05-31
Chai Roadmap


keithamus:
### TL;DR

Over the next few versions we're going to try to do the following:
- Refactor messages to support advanced template strings (per [chaijs/chai#393](https://github.com/chaijs/chai/issues/393))
- Refactor plugin declarations to support new interfaces (per  [chaijs/chai#117](https://github.com/chaijs/chai/issues/117))
- Split out chai codebase into various modules
- Turn `chai` into a metapackage which combines these modules
### Refactor Messages

@svallory has done some excellent work into refactoring the message syntax for chai assertions. This will make writing plugins simpler, as the need for message variants is removed.

See [chaijs/chai#393](https://github.com/chaijs/chai/issues/393) for more.
### Refactor plugin declarations

Right now the plugin system is simple to use, and has lead to a successful plugin community. But it could be better.

Assertions should declare the flags they use in an up-front manner, allowing deeper introspection for interfaces. This will allow us to put effort into making interfaces which are pain points for the community - such as an assert interface that works with <=IE8, and expect/should interfaces which use method assertions over property assertions.

Two of the biggest causes of fragmentation of chai users is due to low browser support, and the property assertions - which spawns off codebases like the following:
- [hapijs/code](https://github.com/hapijs/code)
- [rowanmanning/proclaim](https://github.com/rowanmanning/proclaim)
- [moll/js-must](https://github.com/moll/js-must)
- [wrumsby/ristretto](https://github.com/wrumsby/ristretto)

While we can't be all things to all people, some of these libraries exist purely
because Chai has failed them, and so we see a lot of duplication of effort.

Some of this has been discussed in  [chaijs/chai#117](https://github.com/chaijs/chai/issues/117).
### Split out chai codebase into various modules

We've been putting efforts into splitting chai out into smaller modules, and it's been working well - but I think we can go further. The way I see things, we can split chai into the following:
- chai-core (plugin loader interface, `lib/assert`, `lib/chai/utils.[add/overwrite]*.js`)
- chai-common-assertions (existing assertions, `lib/chai/core/assertions.js`)
- message-formatter (`lib/chai/utils/getMessage.js`, and @svallory's work)
- checkError (complex logic within [`lib/chai/core/assertions.js`](https://github.com/chaijs/chai/blob/master/lib/chai/core/assertions.js#L1215-L1330) .throw assertion)
[...]

PR #475, opened 2015-06-29
Added handling for undefined and null on expect empty (fails on negate flag too)

Daveloper87:
- Added handling for undefined and null on expect empty 

@keithamus Is chai 4.x.x going to be focused towards ES6? If so I can use Object.keys() to coerce primitives.

Also there is the option to do this:
```
else if (obj.hasOwnProperty('length')) {
  expected = obj.length === 0;
 }
```

Which would cover any new objects with length, but it would add support for checking if a function takes parameters or not (not sure that's what you would want from an empty check..).

Just an idea. I will wait for your feedback.

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -376,14 +376,16 @@ module.exports = function (chai, _) {
     var obj = flag(this, 'object')
       , expected = obj;
 
-    if (Array.isArray(obj) || 'string' === typeof object) {
-      expected = obj.length;
-    } else if (typeof obj === 'object') {
-      expected = Object.keys(obj).length;
+    if (obj == null) {
+      expected = flag(this, 'negate');
+    } else if (Array.isArray(obj) || 'string' === typeof obj) {
+      expected = obj.length === 0;
+    } else if ('object' === typeof obj) {
+      expected = Object.keys(obj).length === 0;
     }
 
     this.assert(
-        !expected
+      expected
       , 'expected #{this} to be empty'
       , 'expected #{this} not to be empty'
     );
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -373,19 +373,10 @@ module.exports = function (chai, _) {
    */
 
   Assertion.addProperty('empty', function () {
-    var obj = flag(this, 'object')
-      , expected = obj;
-
-    if (obj == null) {
-      expected = flag(this, 'negate');
-    } else if (Array.isArray(obj) || 'string' === typeof obj) {
-      expected = obj.length === 0;
-    } else if ('object' === typeof obj) {
-      expected = Object.keys(obj).length === 0;
-    }
-
+    var obj = flag(this, 'object');
+    new Assertion(obj).to.exist;
     this.assert(
-      expected
+        Object.keys(Object(obj)).length === 0
       , 'expected #{this} to be empty'
       , 'expected #{this} not to be empty'
     );
--- a/test/expect.js
+++ b/test/expect.js
@@ -397,6 +397,35 @@ describe('expect', function () {
     err(function(){
       expect({foo: 'bar'}).to.be.empty;
     }, "expected { foo: \'bar\' } to be empty");
+
+    err(function(){
+      expect(0).to.be.empty;
+    }, "expected 0 to be empty");
+
+    err(function(){
+      expect(null).to.be.empty;
+    }, "expected null to be empty");
+
+    err(function(){
+      expect(undefined).to.be.empty;
+    }, "expected undefined to be empty");
+
+    err(function(){
+      expect().to.be.empty;
+    }, "expected undefined to be empty");
+
+    err(function(){
+      expect(null).to.not.be.empty;
+    }, "expected null not to be empty");
+
+    err(function(){
+      expect(undefined).to.not.be.empty;
+    }, "expected undefined not to be empty");
+
+    err(function(){
+      expect().to.not.be.empty;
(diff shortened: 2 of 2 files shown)


lib/chai/core/assertions.js
```diff
@@ -376,14 +376,16 @@ module.exports = function (chai, _) {
     var obj = flag(this, 'object')
       , expected = obj;
 
-    if (Array.isArray(obj) || 'string' === typeof object) {
-      expected = obj.length;
-    } else if (typeof obj === 'object') {
-      expected = Object.keys(obj).length;
+    if (obj == null) {
+      expected = flag(this, 'negate');
```
keithamus: This is not ideal - setting the negate flag will apply to the rest of the chain. Better to do a simple type check at the beginning and error out early.

lib/chai/core/assertions.js
```diff
@@ -376,14 +376,16 @@ module.exports = function (chai, _) {
...
+    } else if (Array.isArray(obj) || 'string' === typeof obj) {
+      expected = obj.length === 0;
+    } else if ('object' === typeof obj) {
+      expected = Object.keys(obj).length === 0;
     }
```
keithamus: As mentioned in my previous PR - just using `Object.keys()` seems to be the most optimum path, nicely.

test/expect.js
```diff
@@ -397,6 +397,35 @@ describe('expect', function () {
     err(function(){
       expect({foo: 'bar'}).to.be.empty;
     }, "expected { foo: \'bar\' } to be empty");
+
+    err(function(){
+      expect(0).to.be.empty;
+    }, "expected 0 to be empty");
```
keithamus: FYI the current behaviour is `expect(0).to.be.empty` passes. Can you please explain why you want it to now fail?

test/expect.js
```diff
@@ -397,6 +397,35 @@ describe('expect', function () {
...
+      expect(null).to.be.empty;
+    }, "expected null to be empty");
+
+    err(function(){
+      expect(undefined).to.be.empty;
+    }, "expected undefined to be empty");
+
+    err(function(){
+      expect().to.be.empty;
+    }, "expected undefined to be empty");
```
keithamus: The error messages for these feel very ambiguous - `"expect undefined to be empty"` is a semantic noodle scratcher, I'd prefer to see more specific error messages.

keithamus:
Overall some good work @Daveloper87 - but see my comments above.

I think in summary, I see the assertion for `empty` being something like this, taking a leaf from [the `lengthOf` assertion](https://github.com/Daveloper87/chai/blob/empty_handle_undefined/lib/chai/core/assertions.js#L989-L1002).

```js
    new Assertion(obj).to.exist;
    this.assert(
      Object.keys(obj) === 0
      , 'expected #{this} to be empty'
      , 'expected #{this} not to be empty'
    );
  });
```

This should assert first that the object is not null or undefined - after which it will look at the keys to ensure they are 0.

test/expect.js
Daveloper87: I wasn't sure about this when I added it, and kind of put it here for discussion. It was more because the empty seemed like a means to check something has no properties or length. Happy to remove that though if its deemed ok for numbers to pass.

Daveloper87:
Thanks for your comments, The lengthOf approach and the approach you have outlined in your comment looks a lot cleaner. Was not sure if it was ok to use other assertions, but looking at everything again it seems like its used throughout. So that looks good to me.

My only question is around Object.keys(). Is chai 4.x.x only supporting ES6? because ```Object.keys('some string')``` will fail because it can not be coerced to an object so will throw a TypeError.

keithamus:
Using other assertions will probably one day not be ideal - but don't worry about that for now, just follow existing code conventions. 

Do you have evidence about ES6 Object.keys? Without offending I'd be surprised if ES6 makes breaking changes, so I'm somewhat doubtful of this. Looking at the spec[[1]](http://www.ecma-international.org/ecma-262/6.0/index.html#sec-object.keys) it seems to use [`[[ToObject]]`](http://www.ecma-international.org/ecma-262/6.0/index.html#sec-toobject) which explicitly mentions turning string literals into `new String(value)` (which have keys).

Daveloper87:
Sorry, I did not explain that well. 

Object.keys('hello') would fail in browsers that do not currently have support for ES6. This will include node (without harmony). If you try that in node console or a browser like Safari, you should see a TypeError

keithamus:
Ah, [ES5 Object.keys](http://es5.github.io/#x15.2.3.14) is missing `[[ToObject]]` call. Learn something new every day :smile:! Luckily we can easily emulate this by wrapping the value in `Object`:

```js
var obj = 'hello';
Object.keys(Object(obj)); // ["0", "1", "2", "3", "4"]

var obj = [1, 2, 3];
Object.keys(Object(obj)); // ["0", "1", "2"]

var obj = 42;
Object.keys(Object(obj)); // []
```

Thoughts?

Daveloper87:
Ah great! Likewise! I have pushed a new commit based on your comments.

numbers and functions will be classed as empty, but I guess if you expected an empty array and got a function, you have bigger problems...

keithamus:
Great work @Daveloper87 :smile:.

It'll be a while before the 4.x.x branch gets merged - #457 is our roadmap for now, which we're trying hard to focus on (namely modularising the chai codebase). If you've caught the PR bug then feel free to take a look at some of those and submit some PRs, they'd be very welcome!

Daveloper87:
@keithamus Great! I will have a look over the coming weeks and try to dig in :)

PR #484, opened 2015-07-17
Add respondsTo and satisfies as aliases

couchand:
This pull requests adds the assertion alias `respondsTo` for `respondTo` and `satisfies` for `satisfy`.  This is to complete the API for subordinate clause grammar, such as:

```javascript
expect(myObject).to.have.property('child').that.respondsTo('toString');
expect(myObject).to.have.property('child').that.satisfies(myPredicate);
```

When adding this I noticed one other place the API was incomplete: there is no `exists` property.  This seems reasonable, though, since you can only get to a subordinate clause by previously asserting the existence of something.

Are there docs that I should update for this as well?

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -1368,12 +1368,13 @@ module.exports = function (chai, _) {
    *     expect(Klass).itself.to.respondTo('baz');
    *
    * @name respondTo
+   * @alias respondsTo
    * @param {String} method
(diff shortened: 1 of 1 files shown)


keithamus:
Thanks @couchand! You've already updating the docs, by virtue of adding the `@alias` directives in the docblocks!

couchand:
:beers:

PR #489, opened 2015-07-17
Various work on aliases

astorije:
Here is what this PR does:

- Add `is*` aliases for `extensible`, `notExtensible`, `sealed`, `notSealed`, `frozen`, `notFrozen` and update their respective tests to test the aliases
- Add `is*` aliases for `ok` and `notOk` and update their respective tests to test the aliases
- Make the use of `throws` consistent: before these changes, the name of the assertion in the documentation was `throws`, the examples were using `throw` and the code was using `Throw`. Now it's `throws` everywhere and the rest is aliases
- Add tests for aliases of `throws` (`throw` and `Throw`)
- Since we already had `isTrue`, `isNull`, `isObject`, `isBoolean`, ... (20 in total), I made the `is*` aliases the authoritative source, and kept the original ones as aliases for backward-compatibility. For example, `isOk` should be the "official" version (like `isTrue`), but `ok` still exists. If I am not committing any typo, nothing in the API will be broken, just the way it's presented is more consistent.

This PR doesn't update the main `chai.js` file.

### DIFF
--- a/lib/chai/interface/assert.js
+++ b/lib/chai/interface/assert.js
@@ -72,6 +72,7 @@ module.exports = function (chai, util) {
    *     assert.ok(false, 'this will fail');
    *
    * @name ok
+   * @alias isOk
    * @param {Mixed} object to test
    * @param {String} message
    * @api public
@@ -90,6 +91,7 @@ module.exports = function (chai, util) {
    *     assert.notOk(false, 'this will pass');
    *
    * @name notOk
+   * @alias isNotOk
    * @param {Mixed} object to test
    * @param {String} message
    * @api public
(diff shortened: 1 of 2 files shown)


astorije:
Note that I haven't updated the main `chai.js` file.

astorije:
I can also add tests for theses aliases, to make sure they run OK. Let me know if you want that.

keithamus:
If you would kindly add tests, then I'll happily merge :smile:

astorije:
@keithamus here you go with the tests... I also took some liberties as you can see, making this PR broader but still focused on aliases for the `assert` API. I believe it makes things better by adding consistency and tests, I hope you're gonna like it :)

Let me know if you need anything else on this!

astorije:
Note that since there are a lot of changes, I made sure individual commits were self-contained so that you can go through each of them instead of having to digest the resulting blob in one go!

keithamus:
LGTM :+1:

astorije:
Thanks @keithamus!!

PR #490, opened 2015-07-17
Fix wrong call to the underlying method when checking if an object is frozen

astorije:


keithamus:
Ooops, good call. Missed that in the PR. Thanks @astorije

astorije:
For the record, here is why this buggy version passes the tests:

```js
> var v = { foo: "bar" }
undefined
> Object.isSealed(v)
false
> Object.isFrozen(v)
false
> Object.freeze(v)
{ foo: 'bar' }
> Object.isSealed(v)
true
> Object.isFrozen(v)
true
```

So if you are *frozen*, you are *sealed*.

(Note that the opposite is not true:

```js
> var v2 = { foo: "bar" }
undefined
> Object.isSealed(v2)
false
> Object.isFrozen(v2)
false
> Object.seal(v2)
{ foo: 'bar' }
> Object.isSealed(v2)
true
> Object.isFrozen(v2)
false
```
)

astorije:
@keithamus, can you quickly release version `3.1.1` to make sure anyone who gladly switched to version `3.1.x` to have these methods does not fall in this bug?

PR #495, opened 2015-07-20
JSDoc blocks aligned with corresponding functions

cezarykluczynski:
JSDoc blocks for isAbove, isBelow and isTrue were not directly before the corresponding functions. This PR puts the blocks in their right places.

git diff --patience makes this commit easier to read.


keithamus:
Thanks @cezarykluczynski :smile:

PR #499, opened 2015-07-24
Shortened check for empty

Daveloper87:
- Updated empty based on conversation on issue #472


keithamus:
Looks good, thanks @Daveloper87 :smile:

PR #500, opened 2015-07-24
isAtLeast, isAtMost assertions added to assert interface

wraithan:
I noticed the `assert` interface was missing these methods. One can use the `operator` method but that feels clunky. The `CONTRIBUTING.md` says to not expand the interface without asking due to possible rejection. I'm 100% fine with this being rejected if the expansion of the interface is deemed not useful. I attempted to match style best I could but there is no linter so I'm not 100% positive I've hit every point.

Also, it'd be nice if the `CONTRIBUTING.md` was expanded to discuss whether one should include an updated `chai.js` or not. I assumed not since that appeared to be part of the release process but I wasn't sure.

I went with is* for these to match the rest of the `assert`, I'm happy to rename them to something else that y'all like. I started with `most` and `least` like the `should`/`expect` interfaces but decided matching the interface I was adding them to was more important than matching what the other interfaces call them.

I also opted to not add the aliases `gte` and `lte` as there didn't appear to be any other shorthand aliases in `assert`. Happy to add those as well if those are desired.

Thanks for ChaiJS

### DIFF
--- a/lib/chai/interface/assert.js
+++ b/lib/chai/interface/assert.js
@@ -244,6 +244,25 @@ module.exports = function (chai, util) {
   };
 
    /**
+   * ### .isAtLeast(valueToCheck, valueToBeAtLeast, [message])
+   *
+   * Asserts `valueToCheck` is greater than or equal to (>=) `valueToBeAtLeast`
+   *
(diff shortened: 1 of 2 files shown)


keithamus:
Hey @wraithan thanks for the PR, good work :+1:

Issue #501, opened 2015-07-25
What to expect when you're expecting


astorije:
I was making tests trying to be right on #496 (:smile:) and found that if you assert:
 
```js
expect(3).to.include(2);
```

You end up with:

```js
     TypeError: undefined is not a function
      at Assertion.include (lib/chai/core/assertions.js:217:30)
      at Assertion.assert (lib/chai/utils/addChainableMethod.js:83:49)
      at Context.<anonymous> (test/expect.js:1131:18)
```

The `assert.include(3, 2);` twin assertion reacts similarly, but the key difference is that the `assert.include(haystack, needle)` method definition explicitly says that the haystack should be `{Array|String}` (so one just needs to read the doc). The type of the argument given to `expect()` is never specified in the doc (or at least I couldn't find it), meaning this should `Mixed`.

Making sure the type is controlled for `expect(...).to.include(...)` is similar to issue #491 and PR #496.

I am guessing there are other cases like that but I haven't tested them all. I am not sure if there should be a global way to do that in the code or if we should handle that on a case-by-case basis as what we are leaning to on #496.

keithamus:
Good call as usual @astorije. We should check for other primitives in assertions like these.

I think for these kind of assertions we should probably throw an error explaining the problem. In this instance maybe `AssertionError: expected 3 to include 2, but cannot include multiple values` or something. Thoughts?

astorije:
What do you mean by `cannot include multiple values`?

I would go for something simple and explicit like `AssertionError: argument of expect() must be an Array or a String, number given` (or `3 given`).
It's not necessary or useful to remind the value (`2`) as this is not causing any error and therefore can be misleading.

keithamus:
Agree, yours sounds much better than mine. `3 given` I think would be best. PR welcome :smile:

astorije:
Okay, I'll see if I can find an hour to send you that within the next few of days. I'd personally prefer `[type] given` as a user, but it's up to you.

berkerpeksag:
This is also fixed in https://github.com/chaijs/chai/commit/d7d9420a24a50d88468644386a82f7b0b08262bf (just pasting the merge commit) and can be closed now.

Issue #324, opened 2014-12-11
Deep expect anonymous function


EtienneLem:
Is there any way to deeply expect an anonymous function? i.e. Jasmine has `jasmine.any(Function)`

```js
o = { foo: 'bar', baz: function() {  } }

// Chai (with Mocha)
expect(o).to.deep.equal({
  foo: 'bar',
  baz: Function
})

// Jasmine
expect(o).toEqual({
  foo: 'bar',
  baz: jasmine.any(Function)
})
```

I know about `expect(o).to.have.property('baz').that.is.a('function')`, but it can be a little bit more painful to deal with when the object has more keys.

emmenko:
Any news on this?

keithamus:
Closing this for a more focussed discussion in #644

Issue #97, opened 2012-09-22
inclusion/containment semantics


jfirebaugh:
Chai's `include`/`contain` test equality via `Array#indexOf`, i.e. strictly.

I'm currently working on chai-jasmine (a compatibility library), and Jasmine's equivalent is `toContain`, which [winds up using natural/deep/eql equality](https://github.com/pivotal/jasmine/blob/master/src/core/Env.js#L265-273). It would be nice if this could be implemented entirely in terms of chai assertions and flags.

Options:

* Provide natural/deep/eql semantics for the `include`/`contain` assertion if the deep flag is set, e.g.:

        expect(['A', {some:'object'}, 'C']).to.deep.contain({some:'object'})

  This phrasing suggests to me that things like `expect({a: {b: {c: 'd'}}}).to.deep.contain({c: 'd'})` should also be supported. But then should arrays be recursively searched as well? (Jasmine's `toContain` does not support this.)
* Introduce another flag, e.g. `to.naturally.contain({some: 'object'})`. But what's a good word?
* Introduce another assertion. Again, what would it be called?
* Give up and don't try to write the chai-jasmine `toContain` assertion in terms of chai assertions.

Thoughts?

keithamus:
Closing this for a more focussed discussion in #644

PR #503, opened 2015-07-27
Checking that argument given to expect is of the right type when using with include

astorije:
Relates to #501.

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -125,7 +125,7 @@ module.exports = function (chai, _) {
   });
 
   /**
-   * ### .a(types)
+   * ### .a(type)
    *
    * The `a` and `an` assertions are aliases that can be
    * used either as language chains or to assert a value's
@@ -146,35 +146,23 @@ module.exports = function (chai, _) {
    *     // language chain
    *     expect(foo).to.be.an.instanceof(Foo);
    *
-   * When an array of strings is passed as argument of the assertion, the
-   * value's type must be one of them.
-   *
-   *     expect(myScalar).to.be.a(['string', 'number', 'boolean']);
-   *     expect([]).to.not.be.a(['string', 'number', 'boolean']);
-   *
    * @name a
    * @alias an
-   * @param {String|Array} types
+   * @param {String} type
    * @param {String} message _optional_
    * @api public
    */
 
-  function an (types, msg) {
+  function an (type, msg) {
     if (msg) flag(this, 'message', msg);
-    if (_.type(types) === 'string') types = [types];
-    types = types.map(function (t) { return t.toLowerCase(); });
-    var obj = flag(this, 'object');
-
-    var str = types.map(function (t, index) {
-      var art = ~[ 'a', 'e', 'i', 'o', 'u' ].indexOf(t.charAt(0)) ? 'an' : 'a';
-      var or = types.length > 1 && index === types.length - 1 ? 'or ' : '';
-      return or + art + ' ' + t;
-    }).join(', ');
+    type = type.toLowerCase();
+    var obj = flag(this, 'object')
+      , article = ~[ 'a', 'e', 'i', 'o', 'u' ].indexOf(type.charAt(0)) ? 'an ' : 'a ';
 
     this.assert(
-        types.some(function (expected) { return _.type(obj) === expected; })
-      , 'expected #{this} to be ' + str
-      , 'expected #{this} not to be ' + str
+        type === _.type(obj)
+      , 'expected #{this} to be ' + article + type
+      , 'expected #{this} not to be ' + article + type
     );
   }
 
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -163,13 +163,18 @@ module.exports = function (chai, _) {
     if (msg) flag(this, 'message', msg);
     if (_.type(types) === 'string') types = [types];
     types = types.map(function (t) { return t.toLowerCase(); });
-    var obj = flag(this, 'object')
-      , article = ~[ 'a', 'e', 'i', 'o', 'u' ].indexOf(types[0].charAt(0)) ? 'an ' : 'a ';
+    var obj = flag(this, 'object');
+
+    var str = types.map(function (t, index) {
+      var art = ~[ 'a', 'e', 'i', 'o', 'u' ].indexOf(t.charAt(0)) ? 'an' : 'a';
+      var or = types.length > 1 && index === types.length - 1 ? 'or ' : '';
+      return or + art + ' ' + t;
+    }).join(', ');
 
     this.assert(
         types.some(function (expected) { return _.type(obj) === expected; })
-      , 'expected #{this} to be ' + article + types.join(', ')
-      , 'expected #{this} not to be ' + article + types.join(', ')
+      , 'expected #{this} to be ' + str
+      , 'expected #{this} not to be ' + str
     );
   }
 
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -125,7 +125,7 @@ module.exports = function (chai, _) {
   });
 
   /**
-   * ### .a(type)
+   * ### .a(types)
    *
    * The `a` and `an` assertions are aliases that can be
    * used either as language chains or to assert a value's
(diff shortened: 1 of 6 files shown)


astorije:
Hi @keithamus,

I have a question related to this PR. I know the following is possible:

```js
expect(10).to.be.above(5).and.below(15);
```

(I know that `within` exists but that's for the sake of the example)

but is there a way we could make use of `or` possible? Something like:

```js
expect(10).to.be.above(5).or.below(0);
```

There would be other nice uses of `or`, such as the one I am actually seeking:

```js
expect(myVar).to.be.a('string').or.an('array');
```

For that latter example, `a`/`an` could be modified to accept multiple inputs:

```js
expect(myVar).to.be.a('string', 'array');
```

but that would be nice to have a nice `or` operator. Any thoughts?

astorije:
@keithamus, although my previous question about `or` still holds for Chai's API, I went ahead and added support of multiple types in cba291e (the object's type must be one of those supplied). That should make a lot of sense when the PR is complete, plus it adds a nice feature to the BDD API.
Note that I'll extend `assert.typeOf` to have the same support for consistency, but in future PR as my focus is the original issue and I needed this for my PR.

astorije:
Voilà @keithamus, I am done. A couple of things though:

- Assuming you are a testing a function that returns `null` although you weren't expecting so:

  ```js
  expect(myMysteriousFunction()).to.include(someValue);
  ```
  You would get the following message:
  ``` 
  AssertionError: expected null to be an array, an object, or a string
  ```
  Is that clear enough?

- With my additions on the `a`/`an` assertion, we can easily add this check to all assertions. How can we nicely and uniformly add that info to their docstrings?

Let me know if you have any comments :-)

astorije:
@keithamus... friendly ping? :)

keithamus:
Hi sorry @astorije this was another that slipped through the next.

I'm not sold on passing an array of types to `an`. I'd prefer to see an `or` but the problem is with deeply chained assertions an `or` operator becomes confusing as to where it is switching, e.g.

```js
expect({ foo: { bar: 1 }}).to.have.deep.property('foo.bar').and.be.a('number').and.equal(1).or.equal(2);
```

Which of the following does the above assertion do?
  ```js
 obj.foo.bar && typeof obj.foo.bar === 'number && (foo.bar === 1 || foo.bar === 2)
 ```
 ```js
 obj.foo.bar && (typeof foo.bar === 'number' && foo.bar === 1) || foo.bar === 2
 ```
 ```js
(obj.foo.bar && typeof foo.bar === 'number' && foo.bar === 1) || obj === 2)
 ```

We could potentially have something like `oneOf` which could have a set of assertions passed to it (`expect('foo').to.be.oneOf(expect.to.be.a('number'), expect.to.be.a('boolean'))`) - but obviously our assertions are imperative and so we'd need a fundamental redesign of how they work - or introduce matchers (which, to be honest, is something I've been thinking about anyway).

Case in point is, while #501 is definitely something I'd like to see Chai fix, I feel like this PR is going in the wrong direction for it. Make sense? Thoughts?

astorije:
> Hi sorry @astorije this was another that slipped through the next.

Hey @keithamus, no worries :)

> I'd prefer to see an `or` but the problem is with deeply chained assertions an `or` operator becomes confusing as to where it is switching, e.g.

That is indeed the issue with an `or` operator, so I do not think this would be suitable as is.

> We could potentially have something like `oneOf` which could have a set of assertions passed to it `(expect('foo').to.be.oneOf(expect.to.be.a('number'), expect.to.be.a('boolean')))` - but obviously our assertions are imperative and so we'd need a fundamental redesign of how they work

I thought about a `oneOf`/`either` operator that could also be used like:

```js
expect('foo').to.be.oneOf(function (either)
  either.a('number');
  either.a('boolean');
);
```

but considering how `expect` works around exceptions, I couldn't see a way of doing that easily without, indeed, redesigning Chai as its core, not a suitable solution either.

> or introduce matchers (which, to be honest, is something I've been thinking about anyway).

What do you mean by that? What is a matcher?

> Case in point is, while #501 is definitely something I'd like to see Chai fix, I feel like this PR is going in the wrong direction for it. Make sense? Thoughts?

I don't think this PR is going in the wrong direction (surprising? :D), but let me explain myself.

> I'm not sold on passing an array of types to `an`.

I was not very happy with that at first as well, but a few things convinced me:

- It can actually be useful to test that a returned value is one of several types, and without this change, I do believe it is currently impossible to cleanly achieve using Chai (well, one can wrap multiple `expect(...).to.be.a(...)` in a `try...catch` and hack around that, but boy this is ugly and I sure hope to never find this in my tests!)
- The `any.keys(key1, key2, ...)` assertion already works similarly, as shown by the examples:

  ```js
  expect({ foo: 1, bar: 2 }).to.have.any.keys('foo', 'baz');
  ```

I could add a third bullet point to say that I really need that feature for this check to work, but that wouldn't be a fair argument :-)

If you think this PR is going in the wrong direction, the best thing I can offer is to wrap this check in a helper as mentioned in [my comment above](https://github.com/chaijs/chai/pull/503#issuecomment-127079337). That way, whenever an alternative solution arises, we can just edit it there without breaking our API or plugins that would use that helper.
Furthermore, I can also move the whole one-of-types check from the `a`/`an` assertion to that helper if you really want to keep the assertion array-less, but my opinion is that it's a nice feature to expose to the API.

Overall, what I want is to be done with this PR. I think having a non-ideal solution to improve over time is better than having no solutions until we find the one true perfect solution. Again, with my proposals we can make this work without affecting the API (despite my comment on the first list item) if that's your main concern.

What do you think?

keithamus:
> What do you mean by that? What is a matcher?

Sorry, I should have clarified this. I mean matcher as in the same kind of device [Sinon](http://sinonjs.org/docs/#matchers) has, where they can be used as "generics" in combination with assertions. Done right I think it could give Chai a lot of additional power, but it needs a lot of thinking through and so it hasn't been on the short or medium term roadmap (per se).

The way I see a matcher is that it could be interspersed with other assertions, to enable things like: `expect({ foo: 1, bar: { ... }}).to.deep.equal({ foo: 1, bar: chai.match.object });`  or `expect(myChaiSpy).to.have.been.calledWith(1, 2, chai.match.function.that.has.lengthOf(2));`. In this would provide a generic interface to do these "rough" assertions, and could be used here; for example: `expect(obj).to.match.oneOf(chai.match.number, chai.match.boolean)`.

This has been discussed a bit over in #324 and #97, but as I said nothing really firm yet.

> I was not very happy with that at first as well, but a few things convinced me...

I agree with those points, but I think a good generic solution as described above could fix those - so Im hesitant to add functionality until we know its the right functionality to add.

> Overall, what I want is to be done with this PR. I think having a non-ideal solution to improve over time is better than having no solutions until we find the one true perfect solution. Again, with my proposals we can make this work without affecting the API (despite my comment on the first list item) if that's your main concern.

I agree with this, but with the caveat that a public API is a difficult thing to change, and so "the non-ideal solution" thats put in front of people and people start to use becomes "the solution". If we're discussing the same thing -  how `.an` is changing to take an array might not the be "ideal" solution - then I'd rather not have it, and have `.includes` have private behaviour which we can make public when we land on the right solution. Make sense?

astorije:
OK, a few things here, and I'll try to be concise this time, it's usually my weakness :)

- Regarding the matchers, I have no experience with Sinon matchers and I'm not sure I see the whole concept and benefits at 1am, but what I see is that the discussions started in 2012 and I'm not sure there will be activity around it and/or something I can use before a few months (years?). I cannot rely on that for this existing PR or for #501.
- I think we misunderstood each other: whether it may or may not be the right tool for the job here, I do think having `a`/`an` accept an array (oh, just to be clearer, it's not breaking the existing `a`/`an` assertion, it's adding to it) is a really nice feature to have. I see where I would use it myself, and I'm sure others would too, same as `keys()`. Maybe it deserves its own PR, but I am in favor of extending the existing BDD API with this new feature.
- Now, about the "ideal" and "non-ideal" solutions: I do think having a utility function/helper, like `util.expectedTypes()` (name TDB), same as `util.inspect()` or `util.flag()` **is** the ideal solution. Chai's assertions could use it, as well as other plugins (mine would most likely!). It would provide a nice interface and a clear message/exception when the object tested against an assertion does not respect the accepted types. What I called "non ideal" (potentially non ideal actually) is how we would do the internal machinery. But even when this gets improved, that helper is still there.

So I do think having such helper, whether or not `a`/`an` gets augmented, would be a good thing. Or maybe that's just because I spent a lot of time thinking and working on that so I'm biased :)
Let me know if you like that helper function and I'll start working on that, to show you how it looks outside of the blueprints :) If you don't... I'm not sure there is a solution here.

astorije:
Hi @keithamus, friendly ping? I'd really like this one to get done... :-)

keithamus:
Hey @astorije sorry for not replying quickly enough.

Overloading the `an` assertion is a no go for me, within the scope of this PR - especially to get it wrapped up quickly. We can bikeshed this later, but for now if you move that functionality to something like `util.expectedTypes` as suggested then I'd be happy to merge :smile:

astorije:
Hi @keithamus, thanks for your comment, I am now done with changes.

I have moved things to a util called `expectTypes`. I am using infinitive tense to be consistent with all other utils. Of course, English is not my native language so I'll let you choose better wording to your liking.

Also, you will note that one must call the function with `_.expectTypes(this, ['type1', ...]);` instead of `_.expectTypes(obj, ['type1', ...]);`. There are several reasons to this, two of them being that it prevents users/plugins to misuse this helper (for example for general type checking instead of using `a`/`an`), and it also gives us more freedom if in the future we need something else from the object (like getting the `negate` flag, or any new future flags).

I have updated/added tests accordingly, and reverted the `a`/`an` assertion.

One last question: Do you think we should add support for single values (`_.expectTypes(this, 'string');`, instead of passing it an array)? I don't think we should necessarily, but it's up to you. If so, do we need to define an alias called `expectType`?

astorije:
Oh, forgot to (re-)mention, but once this makes it into `master`, I will add the check to other assertions that need it (`number` for `above`, `within`, ...).

What would be nice would be to add a docstring for that. What do you think and suggest?

astorije:
@keithamus meeeerge meeee :-)

keithamus:
@astorije sorry was on holiday. All merged now, I'll cut a release soon :smile:

astorije:
No worries!
I'll try to include that bit to other assertions before you release though.
Do you think we should add the types supported in the docstring in any way?

keithamus:
Hmmmm... yes I guess that'd be a good idea actually. Feel free to either raise a new PR or tuck it into your next one for the other assertions.

PR #507, opened 2015-08-17
Changed Assertion.NaN to a strict implementation as defined in ES6

### LINKED ISSUE #498: NaN assertion is not strict enough
NaN was recently hopefully recently added. 

But because it relies on the global [isNaN()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN) function, it has the same flaws : because isNan() converts its argument before testing, it returns true in many unexpected cases, and false in others. It is not very predictible (i.e. isNan("bla") returns true, but isNaN("") returns false).

ES6 introduces [Number.isNaN()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN) which behaves as intuitively expected : if the value is a NaN, it returns true, otherwise (strings, dates...), it returns false. A very short polyfill  for ES5 is proposed in the MDN documentation.

A strict isNaN is really needed. I can see 2 options : 
1. rewriting the current NaN property using the MDN polyfill, but this would be a breaking change
2. implementing a strict NaN testing, which could be either
  1. as a simple Property : .striclyNaN
  2. using a flag like .stricly.NaN  (expect(x).to.be.stricly.NaN;

I really think current isNaN() implementation is nearly useless, so I prefer the first option.

If the preferred way is option 2, then option 2.1 feels more like in the spirit of chai, which has different predicates for loose and strict versions of the same predicate (above/least, below/most, equal/eql).

I'm OK to do the job when this discussion will be closed.

olivr70:
Relies on an ES6 polyfill for previous Javascript versions

Fixes #498

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -333,7 +333,7 @@ module.exports = function (chai, _) {
 
   /**
    * ### .NaN
-   * Asserts that the target is a number and is exactly `NaN. No conversions are made.
+   * Asserts that the target is a number and is exactly `NaN`. No conversions are made.
    * Relies on Number.isNaN() function or a polyfill
    *
    *     expect( Math.sqrt(-1) ).to.be.NaN;
@@ -345,7 +345,9 @@ module.exports = function (chai, _) {
    */
   (function () {
     // use standard Number.isNaN (ES6) or its polyfill (as proposed on MDN)
-    var isNanFunc = Number.isNaN || function (val) { return typeof val === 'number' && val !== val; };
+    // as indicated in the Javascript documentation, NaN is the only value 
+    //   different from itself
+    var isNanFunc = Number.isNaN || function (val) { return val !== val; };
     Assertion.addProperty('NaN', function () {
       this.assert(
           isNanFunc(flag(this, 'object'))
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -333,22 +333,27 @@ module.exports = function (chai, _) {
 
   /**
    * ### .NaN
-   * Asserts that the target is `NaN`.
+   * Asserts that the target is a number and is exactly `NaN. No conversions are made.
+   * Relies on Number.isNaN() function or a polyfill
    *
-   *     expect('foo').to.be.NaN;
+   *     expect( Math.sqrt(-1) ).to.be.NaN;
    *     expect(4).not.to.be.NaN;
+   *     expect('foo').not.to.be.NaN;
    *
    * @name NaN
    * @api public
    */
-
-  Assertion.addProperty('NaN', function () {
-    this.assert(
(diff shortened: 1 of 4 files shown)


lib/chai/core/assertions.js
```diff
@@ -333,22 +333,27 @@ module.exports = function (chai, _) {
 
   /**
    * ### .NaN
-   * Asserts that the target is `NaN`.
+   * Asserts that the target is a number and is exactly `NaN. No conversions are made.
```
keithamus: Missing a backtick here? `` `NaN`` should be `` `NaN` ``

lib/chai/core/assertions.js
```diff
@@ -333,22 +333,27 @@ module.exports = function (chai, _) {
...
-  Assertion.addProperty('NaN', function () {
-    this.assert(
-        isNaN(flag(this, 'object'))
-        , 'expected #{this} to be NaN'
-        , 'expected #{this} not to be NaN'
-    );
-  });
+  (function () {
+    // use standard Number.isNaN (ES6) or its polyfill (as proposed on MDN)
+    var isNanFunc = Number.isNaN || function (val) { return typeof val === 'number' && val !== val; };
```
thefourtheye: `typeof val === 'number'` is not necessary.

keithamus: Are you sure? The [spec seems to strongly suggest that it should return early if the type is not `"number"`](http://www.ecma-international.org/ecma-262/6.0/index.html#sec-number.isnan).

keithamus: Actually, you're right, NaN is the only value that does not equal itself in a strict equality operation.

thefourtheye: @keithamus Yup, refer the [`Note` in this section](http://www.ecma-international.org/ecma-262/6.0/index.html#sec-isnan-number)

olivr70: @thefourtheye 
You're right. I left the type check proposed on MDN in case we would be using ES6 Number.isNan(), but the [spec](http://www.ecma-international.org/ecma-262/6.0/#sec-number.isnan) for it states that it does a typecheck on its argument.

Are you OK with the following change on line 348
```javascript
var isNanFunc = Number.isNaN || function (val) { return val !== val; };
```

thefourtheye: Actually I also made PR, as I was already working on it #508

BTW, the function looks fine to me.

keithamus:
Sorry @thefourtheye this PR slipped the net for a while. It's a bit outdated now, could you pretty please rebase and I'll look toward merging it :smile:

thefourtheye:
@keithamus #508 has been landed already

keithamus:
Doi, you're totally right. Thanks for that. I'll close this one.

PR #509, opened 2015-08-28
Cause deep.have.same.members() to print a diff.

LyonesGamer:
As it is, when the above assertion fails, one gets the rather unhelpful error message "expected [ Array(n) ] to have the same members as [ Array(m) ]". If the arrays are the same size, but some value in them is different, this isn't very helpful. It usually ends with me printing out the two arrays to the console and attempting to do a diff myself anyway.

By simply passing ``showDiff = true`` to assert() , it indicates that the arrays should be diffed when the error is printed. This functionality is tested in mocha 2.2.5 for both cases where either array is missing a value in the other and where the value differs between the arrays.

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -1513,6 +1513,7 @@ module.exports = function (chai, _) {
         , 'expected #{this} to not be a superset of #{act}'
         , obj
         , subset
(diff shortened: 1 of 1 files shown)


keithamus:
:+1:

Issue #425, opened 2015-04-13
how can I know which assertion failed ?


mderazon:
Hi guys, don't know if it's a Chai question / Mocha, but I am facing a very basic issue that I feel like i'm missing something.

Some of my tests have more than a few assertions and when some of them fail all I get is something like `AssertionError: expected true to be false`.

Is there any way to know which assertion failed ? Line number / stack trace perhaps ?

I know I can add a text description to each assertion but I don't want to clutter my test files with unnecessary text just to mark which assertion failed when a line number would be sufficient.

Right now I am just commenting out assertion and re-running tests just to find the faulty one. I feel stupid when I do that...

Thanks :+1:

mderazon:
possible duplicate of #377

keithamus:
Hey @mderazon thanks for the issue.

Mocha should present you with a stack trace of the error, including the line which threw the assertion error.

Could you please paste the output of both your tests, and the failing log - and if it does not give you a stack trace I'll hopefully be able to point you as to why not.

mderazon:
Thanks

I actually do see that in some cases Mocha gives me a line number (greyed out below the assertion). 
For example when running the assertion
```
expect(res.body.code).to.equal('something');
```
I get

```
AssertionError: expected 'success' to equal 'something'
      + expected - actual

      +"something"
      -"success"

      at Context.<anonymous> (order-test.js:239:30)
```

But in other cases when I run for example the assertion
```
expect(res.body.data.user.can_buy).to.be.true;
```

I simply get

```
1) should charge successfully


  0 passing (8s)
  1 failing

  1) charge should charge successfully:
     AssertionError: expected false to be true
```

keithamus:
@mderazon what happens if you set `chai.config.includeStack = true;` at the top of the file?

mderazon:
@keithamus it shows the stack!

 Is there any reason why this shouldn't be the default behaviour ?

keithamus:
Honestly, I don't know. I'm sure there is a good reason. @logicalparadox may be able to tell you.

diversario:
I upgraded to chai 2.x and node v0.12.x and now I don't get stack traces anymore, even with `chai.config.includeStack = true`.

keithamus:
@diversario what test runner are you using? What versions? Do you have some sample code? It might be better to file a new issue.

j13z:
[...]

Issue #355, opened 2015-01-29
doesNotThrow swallows stacktrace info


olsonpm:
Given a method that possibly throws an exception, I get the following stacktrace by wrapping it inside doesNotThrow:

```
AssertionError: expected [Function] to not throw an error but 'AssertionError: expected [<...>] to be a function' was thrown
at Context.<method> (<file>:<line>:<char>)
```

if I just run the function without wrapping it

```
<Class> <method>:
AssertionError: expected [<...>] to be a function
  at <Class>.<method> (<file>:<line>:<char>) <----------This line is missing from above
  at <Context>.<method> (<file>:<line>:<char>)
```

keithamus:
Thanks for the report @olsonpm! Do you have any idea what might be causing it? It'd be sweet if you could set up a reduced test case to prove it out. If you have an idea of what is causing it, feel free to make a PR to fix it.

olsonpm:
I appreciate the response - but honestly I won't have time to look into the problem.

dasilvacontin:
I see what happened. The output is actually correct. I reproduced it with this simple code:
```
require('chai').should()

var f = function throwingFunction() {
    throw new Error
}

f.should.not.throw() // comment this line
f()
```
In the first case you are seeing the stack trace of the AssertionError, which is generated (in my case) when executing `.throw()`.

In the second case you are seeing the stack trace of the actual error thrown by the tested function.

I believe that showing the stack trace of the actual thrown error would be more useful, since you are also getting info about where the actual error was thrown.

If you want to change this behaviour, I'd be willing to prepare a PR. Cheers!

olsonpm:
Good to know.  Thanks for taking the time to look into that and sharing

dasilvacontin:
You're welcome!

keithamus:
Many thanks @dasilvacontin for taking the time to figure this issue out. Your idea sounds like something that will benefit chai users so I'd love to see an attempt at a PR!

[...]

PR #514, opened 2015-09-08
Fix stack trace tracking for property asserts

### LINKED ISSUE #358: Stack traces on to.be.true/false
Apologies if this is covered somewhere but I've had no luck with finding any information.

```
expect(var).to.equal(true)
```

throws an error with a full call stack including the location of the statement, but

```
expect(var).to.be.true
```
throws an error where the call stack is all inside chai code.  Is there something that can be done about this, or do I just need to use the first form?

kpdecker:
The ssfi parameter was not being updated for these methods, which ended up causing the trace to be omitted.

Fixes #358

### DIFF
--- a/lib/chai/utils/addProperty.js
+++ b/lib/chai/utils/addProperty.js
@@ -4,6 +4,9 @@
  * MIT Licensed
  */
 
+var config = require('../config');
+var flag = require('./flag');
+
 /**
  * ### addProperty (ctx, name, getter)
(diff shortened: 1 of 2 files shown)


keithamus:
Ahh @kpdecker you're my hero! I'm merging this and we'll look to close #358, #425 and #355 from it.

pschuegr:
:+1:

Issue #525, opened 2015-10-02
should.approximately support


axelpale:
I propose ShouldJS's [`approximately(value, delta, [description])` assertion](https://shouldjs.github.io/#assertion-approximately) should be a part of the core API of Chai. One could use `.within(above, below)` to do the job but writing the both limits easily causes code similar to:

    var epsilon = 0.01;
    ...
    var val = 10;
    (9.99).should.be.within(val - epsilon, val + epsilon);

versus drier:

    var epsilon = 0.01;
    ...
    `(9.99).should.be.approximately(10, epsilon);`

Furthermore because [Chai claims to have Should interface](http://chaijs.com/), having all ShouldJS's methods is expected. Therefore to have a separate Chai plugin does not feel practical.

axelpale:
To underline the inconvenience, in my case (PhantomJS + Mocha + Chai + chai-jquery) going back to original ShouldJS instead of Chai is not an option because the benefits from chai-jquery.

keithamus:
Hey @axelpale thanks for the issue!

We have [`.closeTo`](http://chaijs.com/api/bdd/#-closeto-expected-delta-) which I think it what you're after, to quote the docs:

> ### .closeTo(expected, delta)
> 
> - @param{ Number }expected
> - @param{ Number }delta
> - @param{ String }message_optional_
> Asserts that the target is equal expected, to within a +/- delta range.

> ```js
expect(1.5).to.be.closeTo(1, 0.5);
```

If you'd like, I'll happily accept a PR which aliases `approximately` to `closeTo`. [Heres the code for `.closeTo`](https://github.com/chaijs/chai/blob/master/lib/chai/core/assertions.js#L1440-L1478), and [above that is an example of how we create multiple aliases for a method](https://github.com/chaijs/chai/blob/master/lib/chai/core/assertions.js#L1410-L1438). Simply refactor the closeTo code to use the alias syntax, and add some [new tests for the alias](https://github.com/chaijs/chai/blob/master/test/should.js#L865-L883), and I'll merge :smile:

berkerpeksag:
The alias is added in https://github.com/chaijs/chai/commit/7fe23e9799d2440ffd45541b6e120bab7ad43e36. This can be closed now.

keithamus:
:+1:

axelpale:
Oh, nice! Thank you guys.

PR #527, opened 2015-10-04
Added approximately alias to close to

cjqed:
Solves #525 , barring some questions.

1. Should approximately be added to the assert interface as well? This PR does, but this can be removed if you feel it bloats the interface.

2. The aliasing syntax doesn't allow for custom error messaging. This is fine for things like satisfy vs satisfies, but it gets a bit weird when it's something like closeTo and approximately. I edited the internal error messages to be more explicit ('the arguments to closeTo or approximately must be numbers'), but left the assertion messaging the same as to not break anything that might rely on that specific messaging. I don't really like this, because the closeTo function now knows how it's being used... but not having it can be really confusing for people trying to use `approximately` in a wrong way. What do you think of this? We could also just leave all messaging the same and treat approximately as a true alias.

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -1445,19 +1445,20 @@ module.exports = function (chai, _) {
    *     expect(1.5).to.be.closeTo(1, 0.5);
    *
    * @name closeTo
+   * @alias approximately
    * @param {Number} expected
    * @param {Number} delta
    * @param {String} message _optional_
    * @api public
    */
 
-  Assertion.addMethod('closeTo', function (expected, delta, msg) {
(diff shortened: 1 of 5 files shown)


keithamus:
1. `assert` should be a first class citizen - ideally it would have every method the other two interfaces have, including aliases. So we can keep this, good job!

2. The [precedent has been set with other assertion methods](https://github.com/cjqed/chai/blob/should-approximately/lib/chai/core/assertions.js#L173-L227) - we provide aliases but the messaging always leans on the original method name, but this could be documented better on our website, which is in much need of love anyway. Having said that, I think we can keep the early error regarding type info.

All is good, thanks for the hard work :smile:

Issue #117, opened 2012-12-06
Make assert work in oldIE


domenic:
The assert interface depends on the expect and should interfaces. This is a real shame because it stops us from working in IE<=8. We should make it work!

mightyiam:
OK. How hard is this, please?

Does anyone have a vague estimation of the amount of work this will take?

How deep does this go? Which files is this confined to? @keithamus?

I would like a go at this, please.

mightyiam:
I see that if I desire a part in this I should participate in the [above linked thread](https://groups.google.com/forum/?fromgroups=#!topic/chaijs/TXtkOAGyetU).

Thank you.

wrumsby:
@mightyiam have a look at [ristretto](https://github.com/wrumsby/ristretto) - it provides a subset of the Chai API and we used it to run 1,000s of tests until we no longer needed to support older versions of IE.

mightyiam:
Thank you @wrumsby.

I think that my likely candidates are [Jasmine](http://jasmine.github.io/) and [Unexpected](https://github.com/sunesimonsen/unexpected).

Since I desire to use [Karma](http://karma-runner.github.io) and Jasmine has already a plugin for Karma, I'm leaning towards it more.

Ristretto seems to be AMD, while I'm using CommonJS. I know that gap is fillable but, you know...

mightyiam:
For those looking for an assertion library for IE < 9, [proclaim](https://github.com/rowanmanning/proclaim), at least in my case, was a drop-in replacement for Chai's assert.

jokeyrhyme:
@logicalparadox @vesln @domenic : would you be interested in a pull request that uses [enricomarino/is](https://github.com/enricomarino/is) (or something like it), dropping / extracting the comparison logic in https://github.com/chaijs/chai/blob/master/lib/chai/core/assertions.js ? The Assert API could then be rewritten to use those `Boolean`-returning functions more directly, short-circuiting the setter/getter stuff that breaks in IE8.

keithamus:
@mightyiam sorry for not commenting earlier. I somehow missed this notifications. 

I don't think this will be solved overnight - I was maybe a bit hasty to add the `pull-request-wanted` label. I think we should work towards specifying a design mentioned in @logicalparadox's comment https://groups.google.com/d/msg/chaijs/TXtkOAGyetU/tE0iC_WJxboJ.

[...]

PR #528, opened 2015-10-04
Throw when calling _super on overwriteMethod if method is undefined

### LINKED ISSUE #467: Overwrite method default _super does not throw
It appears [that this is intentional](https://github.com/chaijs/chai/blob/084a419d81338b4559fb56a53ec48bb552f2be82/lib/chai/utils/overwriteMethod.js#L42), but it seems that the following plugin:

```js
Assertion.addProperty('callCenter', function() {
  utils.flag(this, 'callCenter', true);
  this.assert(this._obj instanceof CallCenter);
});

Assertion.overwriteMethod('callCount', function(_super) {
  return function assertCallCount(n) {
    if (utils.flag(this, 'callCenter')) {
      /* custom assertion logic */
    }
    else {
      _super.apply(this, arguments);
    }
  };
});
```

Should work as so:

```js
// this results in the assertion running & should pass/fail based on the call center plugin logic
var callcenter = new CallCenter();
callcenter.should.be.callCenter.and.have.callCount(0);

// using another plugin
var spy = sinon.spy();
spy.should.have.callCount(0);

// this results in a calling through to super every time. it doesn't make sense & should probably always fail
var array = [];
array.should.have.callCount(5);
```

The goal here would be to make a plugin that respects other plugins that may define a `callCount` method. I just wanted to check to see if this was the intended behavior of Chai. If it is, is there another way to achieve this type of result?

wbyoung:
This fixes #467.

### DIFF
--- a/lib/chai/utils/overwriteMethod.js
+++ b/lib/chai/utils/overwriteMethod.js
@@ -39,7 +39,9 @@
 
 module.exports = function (ctx, name, method) {
   var _method = ctx[name]
-    , _super = function () { return this; };
+    , _super = function () {
+      throw new Error(name + ' is not a function');
+    };
 
   if (_method && 'function' === typeof _method)
     _super = _method;
--- a/test/utilities.js
+++ b/test/utilities.js
@@ -259,13 +259,30 @@ describe('utilities', function () {
         expect(_super).to.be.a('function');
         return function () {
           _.flag(this, 'doesnt', true);
-          _super.apply(this, arguments);
         }
       });
     });
 
     var dne = expect('something').to.doesnotexist();
     expect(dne.__flags).to.have.property('doesnt');
+
+    chai.use(function (_chai, _) {
+      expect(_chai.Assertion).to.not.respondTo('doesnotexistfail');
+      _chai.Assertion.overwriteMethod('doesnotexistfail', function (_super) {
+        expect(_super).to.be.a('function');
+        return function () {
+          _.flag(this, 'doesnt', true);
+          _super.apply(this, arguments);
+        }
+      });
+    });
+
+    var dneFail = expect('something');
+    var dneError;
+    try { dneFail.to.doesnotexistfail(); }
+    catch (e) { dneError = e; }
+    expect(dneFail.__flags).to.have.property('doesnt');
+    expect(dneError.message).to.eql('doesnotexistfail is not a function');
   });
 
   it('overwriteMethod returning result', function () {


keithamus:
:+1: thanks @wbyoung

wbyoung:
@keithamus sure thing. Sorry for such a long delay between submitting the issue & sending a PR — life's been busy.

keithamus:
Can totally respect that. Thank you for taking the time to do it! I'll let you know when it is released :smile:

electricmonk:
This is awesome.
I'd very much like to see a change in documentation to encourage plugin authors to respect and share namespaces, like @wbyoung mentions in #467

keithamus:
@electricmonk yeah, that is a great point. Care to make a PR for it? All of the plugin [guides in Chaijs.com](http://chaijs.com/guide/) are markdown files [here on the chai-docs repo](https://github.com/chaijs/chai-docs/tree/master/data/guide). Please, if you have any insight to put in to the guides (which I'm sure you do), then I encourage you to make a PR :smile:

electricmonk:
@keithamus Shall I make the PR to the `4.x.x` branch?
In general, I would suggest deprecating `addMethod` and `addProperty` and only allowing a single interface for registering matchers. But this is a breaking change that would have to be gradually introduced.

keithamus:
@electricmonk 
> Shall I make the PR to the 4.x.x branch?
Yes please!

> In general, I would suggest deprecating addMethod and addProperty and only allowing a single interface for registering matchers. But this is a breaking change that would have to be gradually introduced.

Yup, totally. Have a look at #457 which makes small mention of that, you can also see https://github.com/chaijs/chai/issues/117#issuecomment-90508208 for some sketching on what a simpler plugin system would look like. I have some more code on my machine but I want to hash out a sensible design before we properly dive into that kind of refactoring.

electricmonk:
Hmm. @keithamus, the chai-docs repo doesn't seem to have a `4.x.x` branch, shall I create it?

keithamus:
@electricmonk sorry, for the chai-docs repo if you could please take a look at the `gh-pages` branch, which is effectively inline with 4.x.x (or, at least, they'll probably be released around the same time).

electricmonk:
hey, @keithamus -
Currently the documentation for building a plugin starts from describing `addProperty` and `addMethod`, and then describes the `overwrite...` methods. My thoughts are to deprecate the `add...` methods in the documentation, so that they are mentioned at the end as backwards-compatibility relics, and focus on `overwriteMethod` and `overwriteProperty`. 

However, since this means a major overhaul of the `helpers` page, I thought I'd better ok it with you first. So - are you ok with this? :)

keithamus:
@electricmonk I'd rather focus energy on #585. Also I'm not sure `overwriteProperty` is the one canonical property; it's more complex than `addMethod`, which will still work for the majority of use cases.

electricmonk:
@keithamus so what should be the scope for my docs PR? maybe a note at the top of the page, in the spirit of
> Note: the plugin API is pending a major overhaul in future versions of Chai. In the meantime, please prefer `overwriteMethod` to `addMethod` for the sake of sharing plugin namespaces.

keithamus:
yes that might be a good idea, thanks.

electricmonk:
@keithamus It appears that the [helpers](https://github.com/chaijs/chai-docs/blob/44fa4ac15f4e2a058efb0e3e0731ec78e363787f/_guides/helpers.md) page in the `gh-pages` branch doesn't have the `addMethod` section (although it appears in the ToC). Is that intentional?

keithamus:
That certainly isn't intentional, the `gh-pages` site is a work in progress, so its likely something that got missed. If you could so kindly add it back in that'd be swell :smile:

PR #529, opened 2015-10-05
Fix assert lengthOf example

dxuehu:


keithamus:
Very well noticed, thanks @dxuehu! :+1:

dxuehu:
No prob :)

PR #538, opened 2015-10-21
Revert .empty assertion change from PR #499

### LINKED ISSUE #537: .empty fails for mongoose arrays
Hey,

The new `.empty` assertion that was introduced in #499 fails for mongoose arrays.

You can reproduce with the following:

```js
var mongoose = require('mongoose');

var schema = new mongoose.Schema({
    array: [{
        prop: String
    }]
});

var Foo = mongoose.model('Foo', schema);
var foo = new Foo();

it('should be empty', function () {
    expect(foo.array).to.be.empty;
});
```

It fails with a `AssertionError: expected [] to be empty`.

I am using mongoose @ `4.1.12` but I’m not sure that’s relevant to the issue.

tusbar:
The change introduced in #499 breaks compatibility with Array-like objects (e.g. mongoose arrays).
It should be applied in the next major version.

Fix #537 

cc @keithamus

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -396,8 +396,17 @@ module.exports = function (chai, _) {
    */
 
   Assertion.addProperty('empty', function () {
+    var obj = flag(this, 'object')
+      , expected = obj;
+
+    if (Array.isArray(obj) || 'string' === typeof object) {
+      expected = obj.length;
+    } else if (typeof obj === 'object') {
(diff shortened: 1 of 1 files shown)


keithamus:
I'll cut this as 3.4.0 as new features have been introduced since the last merge. If I can find the time I'll also try to rustle up a 3.3.1 which patches just this against 3.3.0.

tusbar:
:+1:

Issue #545, opened 2015-10-28
should.throw and expect().to.throw assertions broken


sgen:
Both of the following assertions fail 

> 'use strict';
> 
> var chai = require('chai');
> var expect = chai.expect;
> chai.should();
> 
> function throwError() {
>   throw Error('an error');
> }
> 
> it('should doesnt consider equivalent errors equivalent?', function() {
>   (function() {
>     throwError();
>   }).should.throw(Error('an error'));
> });
> 
> it('expect doesnt consider equivalent errors equivalent?', function() {
>   expect(function() { 
>     throwError();
>   }).to.throw(Error('an error'));
> });

with 

>  1)  should doesnt consider equivalent errors equivalent?:
>     AssertionError: expected [Function] to throw 'Error: an error' but 'Error: an error' was thrown
>      at Context.<anonymous> (index.js:14:18)
>
>  2)  expect doesnt consider equivalent errors equivalent?:
>     AssertionError: expected [Function] to throw 'Error: an error' but 'Error: an error' was thrown
>      at Context.<anonymous> (index.js:20:14)

Both assertions should pass according to the docs
http://chaijs.com/api/bdd/#-throw-constructor-
http://chaijs.com/guide/styles/#should-extras

keithamus:
Hey @sgen thanks for the issue!

Sadly the docs are slightly misleading here. We should definitely fix that.

If you pass in an instance (either `new Error('foo')` or `Error('foo')`) then the error is checked for referential equality (see [here](https://github.com/chaijs/chai/blob/master/lib/chai/core/assertions.js#L1254-L1255)). The doc example:

> ```js
expect(fn).to.not.throw(new RangeError('Out of range.'));
```

Is wrong - because `fn` throws a different RangeError (just because they have the same type and message doesn't matter, they're two different instances). The right way to do this would be, as demonstrated in the docs just above that:

> ```js
expect(fn).to.throw(ReferenceError, /bad function/);
```

I'm really sorry that the docs are wrong here - I can imagine that must be frustrating :angry:. Let's get the docs fixed up and remove the incorrect example. If you'd like to make the change, and take the credit for it - then you just need to submit a PR deleting [this line](https://github.com/chaijs/chai/blob/master/lib/chai/core/assertions.js#L1217) and we'll get round to updating the website soon :wink:

PR #546, opened 2015-11-02
Update assertions.js

Pklong:
removed 'out of range' reference error, issue #545


keithamus:
Great work @Pklong! High five :hand:!

Charlotteis:
Nice one! :wave: Thanks from Your First PR (if that's how you found this bug!)

PR #552, opened 2015-11-15
(docs) Add namespace to all public methods

keithamus:
This will make generating the new sites' docs much easier.

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -35,6 +35,7 @@ module.exports = function (chai, _) {
    * - same
    *
    * @name language chains
+   * @namespace BDD
    * @api public
    */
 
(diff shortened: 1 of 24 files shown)


astorije:
@keithamus, I am planning to add this as well in https://github.com/astorije/chai-immutable/issues/39, at least for consistency with Chai, as I try to match the docstrings of my plugin as closely as possible to Chai's docstrings.
Just wondering, does this and the other docstring data have any impact on the new site docs for plugins in general?

keithamus:
It won't for plugins - this was just for core to generate docs. Feel free to add docstrings to help you maintain your own docs though. 

Also, you should take a look at https://github.com/chaijs/chai/issues/585, which will hopefully lead to having an easy way to generate docs, among other good things.

Issue #560, opened 2015-12-02
expect('word$').to.equal('xxx') failure message broken


jurko-gospodnetic:
Executing `expect('word$').to.equal('xxx')` will give a broken failure message:

`AssertionError: expected 'word to equal 'xxx' to equal 'xxx'`

The final `$` sign seems to be used like some internal placeholder and expanded incorrectly.

keithamus:
Hey @jurko-gospodnetic thanks for the issue.

Well, this was a tricky one to figure out! The offending lines are here; https://github.com/chaijs/chai/blob/master/lib/chai/utils/getMessage.js#L45-L48. The reason this is happening is because `String.prototype.replace` gives special meaning to `$` within its second argument.  Specifically, `$1` - `$99`, `$'`, `$` `, `$&` and `$$` are all special substrings which reflect parts of the Regular Expression. When we inspect a String (using `chai.util.inspect`) it returns it back wrapped in quotes:

```js
> chai.util.inspect('foo$')
'`\foo$\'"
```

Of course, as I mentioned before, `$'` is a special substring, which reflects part of the Regular Expression. Here is an example of the problem fully reduced:

```js
'x'.replace('x', '$\'')
```

You'd expect that to equal `$'` but instead it returns `` (an empty string).

#### The Fix

Okay, so I think the easiest way to fix this is to subvert the String behaviour, and give the replace function a different argument. We can give it a function that returns a String, and everything works exactly as you'd expect. So I'd recommend changing [L45-48](https://github.com/chaijs/chai/blob/master/lib/chai/utils/getMessage.js#L45-L48) from:

```js
  msg = msg
    .replace(/#{this}/g, objDisplay(val))
    .replace(/#{act}/g, objDisplay(actual))
    .replace(/#{exp}/g, objDisplay(expected));
```

To:

```js
  msg = msg
    .replace(/#{this}/g, function () { return objDisplay(val) })
    .replace(/#{act}/g, function () { objDisplay(actual) })
    .replace(/#{exp}/g, function () { objDisplay(expected) });
```

This should fix the problem outright. As part of this, I would expect to see tests as well - our [test suite for `getMessage` is here](https://github.com/chaijs/chai/blob/master/test/utilities.js#L337-L353); I'd expect to see the developer who PRs this add the extra test cases demonstrating that the resulting message works well with `$`s in the expected/actual strings.

[...]

PR #570, opened 2015-12-14
Fix error message string substitutions

jurko-gospodnetic:
This is a fix for issue #560.

Tests included.

I notices the `getMessage` tests currently seem to be missing several `getMessage()` implementation features (e.g. how different objects get serialized to strings), but I did not go into implementing those here.


keithamus:
Great work @jurko-gospodnetic :smile:

Issue #515, opened 2015-09-11
No showDiff set on assert interface


timruffles:
Just wondering why? Happy to PR this in. It's a bit of a shame this behaviour magically stopped working.

e.g

```javascript
    test.assert(
        exp == flag(test, 'object')
      , 'expected #{this} to equal #{exp}'
      , 'expected #{this} to not equal #{act}'
      , exp
      , act
// <- showDiff true not passed
    );
```

keithamus:
Hey @timruffles thanks for the issue.

I assume you are saying that at one point you didn't need to pass in `true` as the `showDiff` argument? As far as I can see [the behaviour has been the same for over a year](https://github.com/chaijs/chai/blame/master/lib/chai/assertion.js#L101-L102) - unless something changed somewhere else which effected this? 

However (if Im right in what you're saying), I do agree that passing `exp` and `act` should be enough to make `showDiff` be `true`. So feel free to raise a PR, but since it's a breaking change could you please raise it under the `4.x.x` branch, and we'll get it merged in for the next major release.

Thanks :smile:

keithamus:
To summarise what should change, rather than [L101](https://github.com/chaijs/chai/blob/e4286f81f4a1edef7dee74f79f1883be7ab3d0e4/lib/chai/assertion.js#L101) currently being:

 ```js
if (true !== showDiff) showDiff = false;
```

It should be changed to 

```js
if (false !== showDiff) showDiff = true;
if (undefined === expected && undefined === _actual) showDiff = false;
```

And you'll need to add tests reflecting this change - in the [chai/test/assert.js file](https://github.com/chaijs/chai/blob/a42ac436bbb5e2f8a13faf5e572627d902a93fb6/test/assert.js)

lucasfcosta:
Hi @keithamus, I've been reading your code today and I'm not sure on how to test this.
Would mind taking a look at what I've thought?

```js
  it.only('passing exp and act makes showDiff be true', function() {
    chai.use(function(chai) {
      var test = new chai.Assertion();

      try {
        test.assert(
            'one' === 'two'
          , 'message'
          , 'negateMessage'
          , 'two'
          , 'one'
        );
      } catch(e) {
        //console.log(e);
        assert.isTrue(e.showDiff);
      }
    });
  });
```
This works fine for Node but not on Karma. I have no idea why.
[...]

PR #574, opened 2015-12-19
New showDiff behaviour also depending on expected and actual values

lucasfcosta:
This is related to the #515 issue.

I've used try/catch blocks on the test because adding a new parameter to the `err()` function seems too ugly.

I'm also directly using chai.Assertion because this way we guarantee this test won't be too tightly coupled with other assertion methods implementation.

PS.: Thank you guys for the awesome work you've been doing on Chai.


keithamus:
Thanks for the awesome work @lucasfcosta :smile:

PR #575, opened 2015-12-21
Clean browser before karma

lucasfcosta:
Hi @keithamus, according to what we've been talking on #515, this is the Makefile I was using when making my changes. I avoided using `lib/**` as you said, instead I added the `clean-browser` task as a dependency of every browser's tests.

I've also edited `CONTRIBUTING.md` with clearer instructions according to the new Makefile.

**I also have a doubt:** are you guys planning on using Gulp? Personally, I think it would be more friendly both for different platforms (Windows doesn't have Make available, for example) and for JS programmers (which are used to this kind of task automation).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -121,12 +121,9 @@ git checkout -b <topic-branch-name>
 
 4. Commit your changes in logical chunks. Use Git's [interactive rebase](https://help.github.com/articles/interactive-rebase) feature to tidy up your commits before making them public.
 
-5. Run you code to make sure it works.
+5. Run you code to make sure it works. If you're still having problems please try to run `make clean` and then test your code again.
 
 ```bash
-npm i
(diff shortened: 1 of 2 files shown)


keithamus:
Thanks for this @lucasfcosta :smile:. I hope to get some time over the holidays to do some refactoring in a `5.x.x` branch and will probably take a look at the build tasks at the same time. I'm not [exactly the biggest fan of Gulp et al](http://blog.keithcirkel.co.uk/why-we-should-stop-using-grunt/) but I'd like to move to [npm scripts](http://blog.keithcirkel.co.uk/how-to-use-npm-as-a-build-tool/) for straightforwardness and windows compatibility.

lucasfcosta:
Ah, sure, I totally understand your point of view and it seems very reasonable.
If you need any help with these things you can count on me!

Thank you too!

Issue #441, opened 2015-05-08
Typed Arrays not truncated


jwagner:
Chai does currently not handle typed arrays very well:
http://jsfiddle.net/7LbaveL6/

[Typed Arrays][1] have various uses but are usually big blobs of data (for example from canvas getImageData).

The problem with the way they are currently handled is that it causes the page to freeze when assertions are run on objects containing bigger typed arrays. 

I looked at the code but wasn't sure where the right place for a fix would be.

In my opinion it would be ok to just output the to String representation of the typed array (by treating it like a primitive), but it could also use the array code path with proper truncation.

[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Typed_arrays

keithamus:
Thanks for the issue @jwagner, I also completely agree.

Right now, all code for object inspection lives in [lib/chai/utils/inspect.js](https://github.com/chaijs/chai/blob/master/lib/chai/utils/inspect.js) (I say "right now", because there are plans to finally migrate it over into https://github.com/chaijs/loupe).

To add support for Typed Arrays should be reasonably easy to fix. If I were to fix this, I'd simply create a function called `isTypedArray(value)` which returns true if it is an instance of one of the Typed Array constructors. Then just [above the `isArray` call (L126)](https://github.com/chaijs/chai/blob/master/lib/chai/utils/inspect.js#L126), run the check using pretty much the same if statement as the `isArray` one. Maybe have a special `formatTypedArray` too - which can truncate really long typed arrays.

Of course you'd need to add some good tests to ensure good coverage of this, and add guards to make sure engines without typed arrays do not blow up when trying to reference them. Tests for the inspect util are found in [test/utilities.js#l355-375](https://github.com/chaijs/chai/blob/master/test/utilities.js#L355-L375)

lucasfcosta:
Hi @keithamus, I'm planning on tackling this one.
Should I raise my PR under 4.x.x or master?

Would you mind if I opened a pull request tagged as [WIP] so you guys can review, approve/disapprove my changes?

keithamus:
@lucasfcosta absolutely feel free to raise it under a [WIP] PR! `4.x.x` seems like a good branch to put it under.

lucasfcosta:
Hi @keithamus, I've done some progress and I'd like to confirm if my PR has acceptable code to solve this before refactoring it.
**I've also pointed out some considerations/questions related to this issue**.

[...]

PR #576, opened 2015-12-22
[WIP] Truncate Typed Arrays (#441)

lucasfcosta:
*Please don't merge yet, I'm still doing some experiments*

As I said on the issue itself (#441), I'm gonna leave this PR here while working on this change so you guys can review my code and guarantee everything's fine.

**Notes/Questions:**
1. It's not possible to check for a single `TypedArray` constructor, we have to check each one of them.
2. Which `TypedArrays` should be considered *"too long"* and how are we going to truncate them?
3. Shouldn't we truncate long `Arrays` too? If the answer for this is "yes" I feel like this code could be improved.

**Progress:** 
- [x] Tests
- [x] `isTypedArray()`
- [x] `formatTypedArray()`
- [x] Truncating TypedArray
- [x] Refactor

### DIFF
--- a/lib/chai/utils/inspect.js
+++ b/lib/chai/utils/inspect.js
@@ -4,6 +4,7 @@
 var getName = require('./getName');
 var getProperties = require('./getProperties');
 var getEnumerableProperties = require('./getEnumerableProperties');
+var config = require('../config');
 
 module.exports = inspect;
 
@@ -244,8 +245,9 @@ function formatArray(ctx, value, recurseTimes, visibleKeys, keys) {
 function formatTypedArray(ctx, value, recurseTimes, visibleKeys, keys) {
   var output = [];
 
-  // Truncating TypedArray if it has more than 1000 elements
-  var l = value.length > 1000 ? 5 : value.length;
+  // Truncating TypedArray if it has more elements than the configured threshold
+  var t = config.truncateThreshold;
+  var l = value.length > t ? t : value.length;
 
   for (var i = 0; i < l; ++i) {
     if (Object.prototype.hasOwnProperty.call(value, String(i))) {
--- a/lib/chai/utils/inspect.js
+++ b/lib/chai/utils/inspect.js
@@ -175,6 +175,8 @@ function formatValue(ctx, value, recurseTimes) {
   var output;
   if (array) {
     output = formatArray(ctx, value, recurseTimes, visibleKeys, keys);
+  } else if (typedArray) {
+    output = formatTypedArray(ctx, value, recurseTimes, visibleKeys, keys);
   } else {
     output = keys.map(function(key) {
       return formatProperty(ctx, value, recurseTimes, visibleKeys, key, array);
@@ -238,6 +240,20 @@ function formatArray(ctx, value, recurseTimes, visibleKeys, keys) {
   return output;
 }
 
+function formatTypedArray(ctx, value, recurseTimes, visibleKeys, keys) {
+  var output = [];
+
+  for (var i = 0; i < l; ++i) {
+    if (Object.prototype.hasOwnProperty.call(value, String(i))) {
+      output.push(formatProperty(ctx, value, recurseTimes, visibleKeys,
+          String(i), true));
+    } else {
+      output.push('');
+    }
+  }
+
+  return output;
+}
 
 function formatProperty(ctx, value, recurseTimes, visibleKeys, key, array) {
   var name, str;
--- a/lib/chai/utils/inspect.js
+++ b/lib/chai/utils/inspect.js
@@ -120,7 +120,15 @@ function formatValue(ctx, value, recurseTimes) {
     }
   }
 
-  var base = '', array = false, braces = ['{', '}'];
+  var base = ''
+    , array = false
+    , typedArray = false
+    , braces = ['{', '}'];
+
+  if (isTypedArray(value)) {
+    typedArray = true;
+    braces = ['[', ']'];
+  }
 
   // Make Array say that they are Array
   if (isArray(value)) {
@@ -309,6 +317,22 @@ function reduceToSingleString(output, base, braces) {
   }
 
   return braces[0] + base + ' ' + output.join(', ') + ' ' + braces[1];
+}
+
+function isTypedArray(ar) {
+  // Unfortunately there's no way to check if an object is a TypedArray
+  // We have to check if it's one of these types
+  return (typeof ar === 'object' &&
+          objectToString(ar) === '[object Int8Array]' ||
+          objectToString(ar) === '[object Uint8Array]' ||
+          objectToString(ar) === '[object Uint8ClampedArray]' ||
+          objectToString(ar) === '[object Int16Array]' ||
+          objectToString(ar) === '[object Uint16Array]' ||
+          objectToString(ar) === '[object Int32Array]' ||
+          objectToString(ar) === '[object Uint32Array]' ||
+          objectToString(ar) === '[object Float32Array]' ||
+          objectToString(ar) === '[object Float64Array]'
+        )
 }
 
 function isArray(ar) {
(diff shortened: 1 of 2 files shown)


lucasfcosta:
This is it for now.
The basics are done, please take a look at the notes I left on the PR text.

I'm going to refactor it soon in order to get cleaner and more elegant code.

lib/chai/utils/inspect.js
```diff
@@ -311,6 +344,21 @@ function reduceToSingleString(output, base, braces) {
...
+          objectToString(ar) === '[object Int8Array]' ||
+          objectToString(ar) === '[object Uint8Array]' ||
+          objectToString(ar) === '[object Uint8ClampedArray]' ||
+          objectToString(ar) === '[object Int16Array]' ||
+          objectToString(ar) === '[object Uint16Array]' ||
+          objectToString(ar) === '[object Int32Array]' ||
+          objectToString(ar) === '[object Uint32Array]' ||
+          objectToString(ar) === '[object Float32Array]' ||
+          objectToString(ar) === '[object Float64Array]');
+}
```
keithamus: I suppose it is open to abuse, but you could just use:
```js
/\w+Array]$/.test(objectToString(ar))
```

lib/chai/utils/inspect.js
```diff
@@ -230,6 +241,28 @@ function formatArray(ctx, value, recurseTimes, visibleKeys, keys) {
   return output;
 }
 
+function formatTypedArray(ctx, value, recurseTimes, visibleKeys, keys) {
+  var output = [];
+
+  // Truncating TypedArray if it has more than 1000 elements
+  var l = value.length > 1000 ? 5 : value.length;
```
keithamus: This seems strange - if I had an array with 999 elements it would show all 999, but with 1000 it would only show 5?

lucasfcosta: Yes I thought this was a little bit strange too, but I had no idea how you guys would like the representation of the `Array` to be, are there any guidelines?

lib/chai/utils/inspect.js
```diff
@@ -230,6 +241,28 @@ function formatArray(ctx, value, recurseTimes, visibleKeys, keys) {
...
+
+  for (var i = 0; i < l; ++i) {
+    if (Object.prototype.hasOwnProperty.call(value, String(i))) {
+      output.push(formatProperty(ctx, value, recurseTimes, visibleKeys,
+          String(i), true));
```
keithamus: To be honest I haven't used Typed Arrays at all so don't know the best way to show two in a comparison - is showing the start a good idea? For files expressed as typed arrays the first few bytes are going to be magic bytes right, so will largely always be identical?

keithamus:
@lucasfcosta the general direction you're going with this is very good. Despite the two comments above, I'm happy with this code.

lucasfcosta:
Thank you!
I'm gonna take a look at your comments and make it as best as I can!

I'm a fairly new programmer, so your orientation helps me a lot.
Thank you again!

lib/chai/utils/inspect.js
keithamus: We have `chai.config.truncateTreshold` which could be used as a metric - but I suppose my point was less about the cutoff figure, but more that the cutoff makes a dramatic change. If it was `var l = value.length > 1000 ? 1000 : value.length;` it would make much more sense (although 1000 is maybe a bit too high still, perhaps a number like 20 or so?)

lucasfcosta:
Hi @keithamus, sorry for this time without commits (christmas and everything postponed my GitHub plans hahahaha)

Well, I've made the changes you pointed on this PR and tests are running smoothly. But as you've already questioned: what's the best way to show two in a comparison? I too am not aware of the most common TypedArray's use cases therefore I'm not sure what the best format would be.

About arrays: shouldn't they be truncated or not according to the `config.truncateThreshold` too?

keithamus:
@lucasfcosta thanks for sticking with this, I completely appreciate that you can't dedicate 100% of your time to Chai, and in fact I would certainly not want you to. Hope you enjoyed your christmas holidays :smile:.

I don't think I was particularly clear about truncateThreshold, and that was my fault - something which I will correct now:

The way truncateThreshold works right now, is that if the Stringified value above a certain length, then instead of displaying the stringified value, we display a truncated value (see [objDisplay.js](https://github.com/chaijs/chai/blob/34a2f1715df4b7856a55585732818a8e2595e7f8/lib/chai/utils/objDisplay.js#L31-L49)). For example if you have `[ 'hello world', 'goodbye world', 'something else' ]` (stingified, this is over 40 characters) we'll instead display the truncated `[ Array(3) ]`. Importantly, truncateThreshold is configurable with a sensible default because you might think 40 chars is too short for your particular project, which is fine. Obviously this is a rather crude way to limit the amount of data displayed in an error message, but for the most part if kind of works.

For TypedArrays, where we know all values are ints of a fixed upper length (they're only numbers after all), we can reasonably predict the length of the string during parse time and therefore offer better truncation heuristics, so, using the truncateThreshold we can keep useful info about the array but also keep this threshold for reasonable output.

To demonstrate this with code, I'd expect some code a bit like this:

```js
str = '[ ';
for (var i in output) {
  if (str.length >= config.truncateThreshold - 7) {
    str += '...';
    break;
  }
  str += output[i] + ', ';
}
str += ' ]';
```

The above code stringifies the array from within the loop, breaking as soon as the string is about to go over the threshold. Importantly: this doesn't rely on an arbitrary array length - but the configurable truncateTreshold property.

Hopefully this makes sense.

lucasfcosta:
Ah, thanks @keithamus, everything is clear now.

But I was thinking about the better way to do this and I think we may not mix display logic with inspection logic, since we've got `objDisplay.js` I think it could be something like:

```js
module.exports = function (obj) {
  var str = inspect(obj);
  var type = Object.prototype.toString.call(obj);
  var arrayMatch = / (.*Array)]$/.exec(type);

  if (config.truncateThreshold && str.length >= config.truncateThreshold) {
    if (type === '[object Function]') {
      return !obj.name || obj.name === ''
        ? '[Function]'
        : '[Function: ' + obj.name + ']';
    } else if (arrayMatch !== null) {
      return '[ ' + arrayMatch[1] + '(' + obj.length + ') ]';   
    } else if (type === '[object Object]') {
      var keys = Object.keys(obj);
      var kstr = keys.length > 2
          ? keys.splice(0, 2).join(', ') + ', ...'
          : keys.join(', ');
      return '{ Object (' + kstr + ') }';
    } else {
      return str;
    }
  } else {
    return str;
  }
};
```

This way we can show both an `Array` and a `TypedArray` in a similar manner.

On `inspect.js` the code for `formatTypedArray()` could be:

```js
function formatTypedArray(value) {
  str = '[ ';
  
  for (var i = 0; i < value.length; i++) {
    if (str.length >= config.truncateThreshold - 7) {
      return Object.create(Object.getPrototypeOf(value), 
        {length: {value: value.length}});
    }
    str += value[i] + ', ';
  }

  // Removing trailing `, `
  str = str.slice(0, -2) + ' ]';

  return str;
}
```

What do you think about it? Too see all these changes please take a look at the last commit. If you want to keep the logic inside `inspect.js` I won't mind rolling back my last commit and doing everything as you said on the last comment.

Thanks for the help!

lucasfcosta:
Thanks, @keithamus, everything makes sense now!
Would you like to show the `TypedArrays` the same way it happens with `Arrays`, for example: `[ Uint8Array(3) ] or would you like to use `...` as I've thought before?

keithamus:
@lucasfcosta Nope, I think the way Arrays are currently displayed is not great. Ideally we'd have all objects represent themselves truncated by ellipsis and `[ Array(n) ]` style notation as a worst-case scenario. I'm going to put some time in soon to get all other values behaving exactly like this PR - but this is a fantastic start to where we want to go.

I'll merge this now as it's looking great.

keithamus:
@lucasfcosta if you feel like you need to put any more work into this - feel free to put another `4.x.x` PR up. Personally I feel this is exactly where it should be though, so kudos :tada:!

lucasfcosta:
@keithamus Ok then, thanks for the help!

Issue #594, opened 2016-01-21
Add an ES6-friendly way to register "should"


callumlocke:
I suggest adding a file named `should.js` at the root of this project that auto-registers 'should', i.e. `require('./index').should();`.

```js
/* CURRENT SITUATION */

  // commonjs style looks nice
  require('chai').should();

  // but es6 isn't so nice
  import chai from 'chai';
  chai.should();

  // or you can do this, but it's still 2 lines
  import {should} from 'chai';
  should();

/* PROPOSAL */

  // single-line way to register 'should' in ES6
  import `chai/should`;

  // nb. other libraries use this convention too
  import 'babel-core/register';
  import 'coffee-script/register';
```

Note: people sometimes think side-effecting imports feel dirty, but it's arguably less dirty than a globally side-effecting function call like `.should()`. When you have a 'fromless' import that doesn't assign any variables, the syntax is explicitly intended for side-effects.

keithamus:
Something I'd like to address in the roadmap (https://github.com/chaijs/chai/issues/457). Not sure what the right way to address this is though, as each has its downsides, especially when using plugins:

```js
/* CURRENT SITUATION */

  // commonjs style looks nice
  require('chai').use(require('some-plugin')).should();

  // but es6 isn't so nice
  import chai from 'chai';
  import somePlugin from 'some-plugin';
  chai.use(somePlugin).should();

  // or you can do this, but it's still 2 lines
  import chai, {should} from 'chai';
  import somePlugin from 'some-plugin';
  chai.use(somePlugin);
  should();

/* PROPOSAL */

  // single-line way to register 'should' in ES6
  import chai from 'chai';
  import somePlugin from 'some-plugin';
  chai.use(somePlugin);
  import `chai/should`;

  // nb. other libraries use this convention too
  import 'babel-core/register';
  import 'coffee-script/register';
```

keithamus:
I suppose keeping `chai.should()`, `chai.expect`, `chai.assert` would be fine, but adding `import 'chai/should'` would help for expressiveness for those who don't use plugins. So the trifecta becomes:

```js
// Plugin-less
import { expect } from 'chai'
import { assert } from 'chai'
import 'chai/should'
// if you need should.exist/should.fail etc:
import should from 'chai/should'

// With plugins
import chai from 'chai';
import somePlugin from 'some-plugin';
const expect = chai.use(somePlugin).expect

import chai from 'chai';
import somePlugin from 'some-plugin';
[...]

PR #604, opened 2016-01-29
Add script that registers should as a side-effect

callumlocke:
See #594.

Allows a cleaner way to register `should` when using ES6 import syntax.

```js
import 'chai/should';

// as opposed to
import {should} from 'chai';
should();
```


keithamus:
Great work :bowtie:

Issue #544, opened 2015-10-27
Feature request: .change/.increases etc with a function for value


elado:
// `chai-change` plugin doesn't seem to work anymore (https://github.com/timruffles/chai-change/issues/4)

Seems like `assertChanges` in `assertions.js` expects `object, prop`. It limits the usage of this kind of assertion, as I can't do:

```js
assert.increases(
  someOperation,
  () => getSomething().length
)
```

It's good in cases I use immutable data, and `getSomething` returns different object completely.

RSpec's `change` matcher has this feature, it can get either an object and a method name or a function for more arbitrary operations.

https://www.relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/change-matcher

elado:
Currently monkey patching:

```js

assert.changes = function (fn, valueFn, options={}) {
  var valueBefore = valueFn()
  fn()
  var valueAfter = valueFn()

  if (options.by) {
    assert.equal(valueAfter, valueBefore + options.by)
  }
}

assert.changes(
  someOperation,
  () => getSomething().length,
  { by: 1 }
)
```

keithamus:
Hey @elado thanks for the issue.

If you'd like to make a PR for this feature I'd happily review it with a view to merging :smile:. You'll need to modify the [assertChanges](https://github.com/chaijs/chai/blob/master/lib/chai/core/assertions.js#L1601-L1615) assertion, as well as [assertIncreases](https://github.com/chaijs/chai/blob/master/lib/chai/core/assertions.js#L1638-L1652), and [assertDecreases](https://github.com/chaijs/chai/blob/master/lib/chai/core/assertions.js#L1675-L1689). Documentation is above each method ([here's the assertchanges docs](https://github.com/chaijs/chai/blob/master/lib/chai/core/assertions.js#L1581-L1599)) - that will need updating with examples. Finally, extra tests will be needed to cover this, in all three interfaces: [assert](https://github.com/chaijs/chai/blob/master/test/assert.js#L858-L880), [expect](https://github.com/chaijs/chai/blob/master/test/expect.js#L1122-L1147), and [should](https://github.com/chaijs/chai/blob/master/test/should.js#L951-L977). 

I look forwarding to seeing what you come up with :grinning:

lucasfcosta:
Hi @keithamus, I'm ready to tackle this one, can I or is anyone else already working on this?
If the answer is yes, should I raise it under the `4.x.x` branch?

keithamus:
[...]

PR #606, opened 2016-01-31
Fix missing msg argument for change related assertions

lucasfcosta:
As I was reading the code to solve #544 I found out the message parameter for change related methods (increases, decreases, etc...) on the assert interface wasn't being used at all.

I also took a look at [chai's docs](http://chaijs.com/api/assert/#changes) just to make sure it should be used.

### DIFF
--- a/lib/chai/interface/assert.js
+++ b/lib/chai/interface/assert.js
@@ -1344,7 +1344,7 @@ module.exports = function (chai, util) {
   }
 
    /**
-   * ### .changes(function, object, property)
+   * ### .changes(function, object, property, [message])
    *
    * Asserts that a function changes the value of a property
    *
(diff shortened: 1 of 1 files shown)


keithamus:
Looks great, but could you also please change the doc strings above each method to reflect the change? The doc strings above are used to generate the site documentation so it'd be good to advertise this feature if we're going to have it :smile:

lucasfcosta:
Oh, of course, I thought about doing it when I've found this out but then I totally forgot it
Gonna do it :smiley:

lucasfcosta:
Hi @keithamus, I thought I've forgotten it, but actually this was already in the JSDoc comments, it just wasn't implemented.

You can take a look at [this line](https://github.com/chaijs/chai/pull/606/files#diff-3097782fd7a1ba72828599a4e764de19R1359) for example. [The docs on Chai's website](http://chaijs.com/api/bdd/) also say we can have a message argument for this assertion.

lucasfcosta:
Done!
I think all the doc strings are okay now.

PS.: The `message` argument was already being listed as a `@param` and it was [documented on chai's website too](http://chaijs.com/api/bdd/#change), it just wasn't implemented and the doc strings needed to be updated properly.

keithamus:
💥

Issue #193, opened 2013-09-05
Chain .property() assertions on objects


bodokaiser:
Hello,

is there a workaround to allow something like this:

```js
chai.expect({ foo: 'bar', bar: 'foo' })
  .to.have.property('foo', 'bar')
  .and.to.have.property('bar', 'foo')
  ;
```

wbyoung:
I find that this would be especially nice when used in conjunction with @domenic's [`chai-as-promised`](https://github.com/domenic/chai-as-promised/) to avoid repetitive `.should.eventually`. Currently you have to:

```js
var profile = getProfile();
profile.should.eventually.have.property('id', 1);
profile.should.eventually.have.property('username', 'wbyoung');
```

But I think this would be a bit nicer chained:

```js
getProfile().should.eventually.have.property('id', 1)
  .and.have.property('username', 'wbyoung');
```

In some ways, this could be solved via a `properties` assertion (which was discussed & understandably rejected in #72).

I like the _outdent_ idea proposed by @logicalparadox best. Perhaps `also` could serve as this word:

```js
getProfile().should.eventually.have.property('id', 1)
  .and.also.have.property('username', 'wbyoung');
```

wbyoung:
I made a small plugin, called [`chai-also`](https://github.com/wbyoung/chai-also) for this & just pushed it to npm. No browser support, but that wouldn't be hard to add if anyone is interested.

keithamus:
I quite like the idea of `also`... but it needs some more discussion to fill gaps; for example I can see people expecting to traverse deep objects like so:

```js
var multiLayerObject = {
    foo: {
        bar: 1,
        baz: 2,
    },
    bing: {
        bong: 3
        bash: 4
    }
};

// Does it just go up a level?
expect(multiLayerObject).to.have.property('foo')
    .and.have.property('bar', 1)
    .and.also.have.property('baz', 2)
  .and.also.also.have.property('bing') // ???
    .and.have.property('bong', 3)
    .and.also.have.property('bash', 4);

// Or does it reset to the originally object within expect()?
expect(multiLayerObject).to.have.deep.property('foo.bar', 1)
    .and.also.have.deep.property('foo.baz', 2)
    .and.also.have.deep.property('bing.bong', 3)
    .and.also.have.deep.property('bing.bash', 4);
```

At the risk of sounding cynical - I can see either implementation getting a few issues springing up discussing semantics and trying to alter them to the users use-case. 

[...]

PR #610, opened 2016-02-08
Update include assertion docs for object properties

wbyoung:
Originally added in #230, asserting that an object contains a subset of properties and values was never documented. Questions like #193 have popped up asking for this feature even though it's been present since 1.9.0.

The discussion in #193 focused mostly around getting something like this to work via the `property` assertion, but that discussion should be considered moot by the existing functionality already present in `include` that was simply overlooked.


keithamus:
:+1:

Issue #339, opened 2015-01-02
Assert an increase/decrease by an amount (delta)


keithamus:
As per #333 and #330, we should be able to assert on the delta of a `.changed`/`.increased`/`.decreased`.

So far (in #333) the proposed syntaxes are:

```js
foo.should.change('bar').by(5);
foo.should.increase('bar').by(5);
foo.should.decrease('bar').by(5);
```
or
```js
foo.should.change('bar', 5);
foo.should.increase('bar', 5);
foo.should.decrease('bar', 5);
```

/cc @oveddan @cmpolis @logicalparadox

cjqed:
The former is closer to the rest of the syntax in chai. In my opinion we should always aim for tests to be human readable, and the 5 in the second set just seem like magic numbers.

timruffles:
both @matthijsgroen's [chai-changes](https://github.com/matthijsgroen/chai-changes) and  [chai-change](https://github.com/timruffles/chai-change) (disclaimer: mine) plugins support this, along with async assertions (promises in Matthijs' and callbacks in mine).

I guess the change of maintainer explains why the changes feature didn't land in 2013 in #218, I'm not bitter ;)

keithamus:
Hey @timruffles, sorry about the overlap - wasn't aware of either of these plugins at the time of merging the original PR. As you ant @matthijsgroen both have plugins which cater for this - do you have any input on this issue? Even if that input is "don't do it" :smile:.

timruffles:
I really like the `by` feature! It's in both the linked plugins.

If I was to make a suggestion, it's that the current implementation in core doesn't support async use-cases, or functional style:

e.g the below - which is bread and butter node webdev - isn't possible in the current implementation:

```javascript
it('creates a version', function(done) {

	assert.change(function(cb) {
		create({
			type: "plan",
			name: "v2",
		}, cb)
	}, function(cb) {
			Version.count({}, cb);
	}, { 
		callback: done, by: 1,
	});
		
});

```

matthijsgroen:
https://github.com/matthijsgroen/chai-changes#changebydelta

The change library supports async, automatically if the return value of the `when(function() { })` is a promise.

lucasfcosta:
+1 for the `by` feature.
I really think having `by` improves readability, which is one of the main things that makes Chai awesome.

Also, I can't think of a better way to implement it other than having an assertion just for it.
Even if we have it just because of `changes`, `increases` and `decreases` I still think this is a price worth paying.

[...]

PR #621, opened 2016-02-12
[WIP] Change by delta (New Assertion)

lucasfcosta:
As discussed on #339 this is the implementation of the `by` assertion.

I've tagged this as _WIP_ because I want to confirm with you guys (especially @keithamus, since he is the one involved issue #339) how this is going to get implemented on the `assert` interface.

I thought about two possible ways to do it, here they are:

1. Create six new methods, one for each of the change related methods of this interface.

  Doing this seems redundant and cumbersome, but this seems to follow the pattern of the rest of the interface.

  If we did this, we would have these new methods: `changesBy`, `doesNotChangeBy`, `increasesBy`, `doesNotIncreaseBy`, `decreasesBy` and `doesNotDecreaseBy`.

  **I also think this is the most adequate approach.**

2. Return the created assertion and allow the user to chain the `.by()` assertion.
  The interface's methods would be something like:

  ```js
  assert.changes = function (fn, obj, prop, msg) {
    if (arguments.length === 3 && typeof obj === 'function') {
      msg = prop;
      prop = null;
    }

    return new Assertion(fn, msg).to.change(obj, prop);
  }

  // This allows the user to write (for example):
  assert.changes(aFunction, anObject, 'aProperty').by(6);
  ```

  Personally, I don't like this because it adds a new kind of behavior to this interface. It seems to be cleaner but actually it causes a mess because it does not follow the other method's patterns

You can also feel free to suggest other improvements and point mistakes, I will always be open to others' opinions.

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -1634,30 +1634,30 @@ module.exports = function (chai, _) {
    * @name change
    * @alias changes
    * @alias Change
-   * @param {String} object
+   * @param {String} target
    * @param {String} property name _optional_
    * @param {String} message _optional_
    * @namespace BDD
    * @api public
    */
 
-  function assertChanges (object, prop, msg) {
+  function assertChanges (target, prop, msg) {
     if (msg) flag(this, 'message', msg);
     var fn = flag(this, 'object');
     new Assertion(fn).is.a('function');
 
     var initial;
     if (!prop) {
-      new Assertion(object).is.a('function');
-      initial = object();
+      new Assertion(target).is.a('function');
+      initial = target();
     } else {
-      new Assertion(object, msg).to.have.property(prop);
-      initial = object[prop];
+      new Assertion(target, msg).to.have.property(prop);
+      initial = target[prop];
     }
 
     fn();
 
-    var final = prop === undefined ? object() : object[prop];
+    var final = prop === undefined ? target() : target[prop];
     var msgObj = prop === undefined ? initial : '.' + prop;
 
     this.assert(
@@ -1682,30 +1682,30 @@ module.exports = function (chai, _) {
    * @name increase
    * @alias increases
    * @alias Increase
-   * @param {String} object
+   * @param {String} target
    * @param {String} property name
    * @param {String} message _optional_
    * @namespace BDD
    * @api public
    */
 
-  function assertIncreases (object, prop, msg) {
+  function assertIncreases (target, prop, msg) {
     if (msg) flag(this, 'message', msg);
     var fn = flag(this, 'object');
     new Assertion(fn).is.a('function');
 
     var initial;
     if (!prop) {
-      new Assertion(object).is.a('function');
-      initial = object();
+      new Assertion(target).is.a('function');
+      initial = target();
     } else {
-      new Assertion(object, msg).to.have.property(prop);
-      initial = object[prop];
+      new Assertion(target, msg).to.have.property(prop);
+      initial = target[prop];
     }
 
     fn();
 
-    var final = prop === undefined ? object() : object[prop];
+    var final = prop === undefined ? target() : target[prop];
     var msgObj = prop === undefined ? initial : '.' + prop;
 
     this.assert(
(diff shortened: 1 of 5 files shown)


keithamus:
I think Option 1 sounds like the best idea :smile:.

lib/chai/core/assertions.js
```diff
@@ -1634,32 +1634,38 @@ module.exports = function (chai, _) {
...
     fn();
 
-    var final = prop === undefined ? object() : object[prop];
+    var final = prop === undefined ? target() : target[prop];
     var msgObj = prop === undefined ? initial : '.' + prop;
 
+    // This gets flagged because of the .by(delta) assertion
+    flag(this, 'msgObj', msgObj);
+    flag(this, 'initial', initial);
+    flag(this, 'final', final);
```
keithamus: I'm not sold on the names of these flags. They're very likely only going to be used by the increase/decrease/change set - having then more specific to that would be good.

test/expect.js
```diff
@@ -1129,11 +1129,16 @@ describe('expect', function () {
...
     expect(sameFn).to.not.change(obj, 'value');
+    expect(sameFn).to.not.change(obj, 'value').by(1);
+
     expect(sameFn).to.not.change(obj, 'str');
     expect(bangFn).to.change(obj, 'str');
-    expect(batFn).to.change(lenFn);
-    expect(sameFn).to.not.change(lenFn);
+
+    expect(batFn).to.change(lenFn).by(1);
+    expect(sameFn).to.not.change(lenFn).by(1);
```
keithamus: I'd like to see some test cases confirming when the deltas end up being negative, and also some error cases.

lib/chai/core/assertions.js
lucasfcosta: Agreed :+1: 
I'm gonna do it

test/expect.js
lucasfcosta: Really seems like a situation we should get covered.

About the error cases: you mean trying to use `by` alongside an assertion other than `change`, `increases` and `decreases` or testing the error it throws when the delta doesn't match?

lucasfcosta:
Thanks for your feedback! It was very valuable (as always)

lucasfcosta:
Hi @keithamus, sorry for the delay on this, well, I've got a doubt here about the behavior of `.by` involving `.not`.

When someone uses, for example, `anyFunction.should.not.change(anObject, 'aKey').by(5)`, should it raise an exception only if the change had a delta value of `5`? Currently if `anyFunction` changed the value of `anObject` by 10, for example, it would throw an Exception because it has changed and this sentence had `should.not.change` in it, I think this may not be the desired behavior.

If so, how can we detect there is a `by` assertion after `change` before even running the assertion?
I've been thinking about a way to do this but I couldn't find one.

keithamus:
Presumably `change()` would end up throwing an error before `by()` got a chance to execute?

lucasfcosta:
Yes, exactly, currently I'm making `change`, `increase` and `decrease` call the `by` assertion, which will run the tests.

`change`, `increase` and `decrease` will only create flags that will be used by the `by` assertion.

I'll make sure to explain everythin when adding it to this PR.

lucasfcosta:
Okay, so, after days analyzing this I've come to a dead end, so let me explain the problem we have here and the possible solution:

**Q: What is the problem?**
**A: ** When running assertions like: `expect(fn).to.not.change(object, 'key').by(2)` the `negate` flag gets applied both to `change` and `by`, so if `fn` changes 'key' (even if it's not by 2) it will fail before it even gets to the by.

**Q: Why does this happen?**
**A: ** We currently run the `change`, `increase` and `decrease` assertions as soon as they get called. When we type `expect(fn).to.not.change(object, 'key')` the `change` assertion runs immediately, therefore when we run `expect(fn).to.not.change(object, 'key').by(2)` we should be saying "fn should not change 'key' by 2, every other change (or no change at all) is accepted" but we're actually saying "fn should not change 'key'" and then saying "it should not change 'key' by 2".
Got it?

**Q: And how can we solve this?**
**A: ** The best idea I've had until now is to change the syntax to:

```
expect(fn).to.change(obj, 'key').but.not.by(2)
```

Using it this way the `not` getter would turn on the negate flag only for the `by` assertion.
Adding a 'but' word just for chain purposes also seems like a good idea.

Please, @keithamus and others, feel free to give feedback/ideas.
If you accept this syntax I'm ready to commit it on this branch (and don't worry, I'll rebase to keep the commit log as clean as possible).

keithamus:
@lucasfcosta thanks for the detailed description of the issue, however I'm not sure it is that much of a problem. If `expect(fn).to.not.change(object, 'key')` fails, then I would expect `expect(fn).to.not.change(object, 'key').by(2)` to also fail. Chai's chaining comes with an implicit agreement that the first failing assertion will fail the whole assertion.

You mention `expect(fn).to.change(obj, 'key').but.not.by(2)` as a workaround, which I think is a good workaround and the kind of syntax we should be aiming for. This should work out of the box right?

lucasfcosta:
@keithamus, Yes, it works right out of the box, because `not` gets the `negate` flag turned on only for the by assertion.

I just need to add `but` as a "chaining" only word and change the assertions on the `assert` interface.

I will make sure to make this PR as clean as possible so you can take a look at the final version of these changes.

Thanks for your feedback! I'll probably commit the code for this until tomorrow.

lucasfcosta:
Well, it's done.
I also did some minor changes to variable names and property checks related to `change` methods.
Please tell me if I've forgotten something. :smile:

keithamus:
Great work @lucasfcosta. Happy to merge this in :smile:

PR #626, opened 2016-02-17
Convert keys to strings

stpeng:
Property keys should be treated as strings if they aren't already (e.g. integers, floats, objects). It should work with any type, but testing for integers is sufficient.

Single property assertions already work because they use standard Object methods, but I've added some test cases for completeness.

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -1171,6 +1171,8 @@ module.exports = function (chai, _) {
 
     if (!keys.length) throw new Error('keys required');
 
+    keys = keys.map(String);
+
     var actual = Object.keys(obj)
       , expected = keys
       , len = keys.length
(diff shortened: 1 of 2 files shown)


keithamus:
Thanks for the PR @stpeng :smile: 

Something we haven't got in our tests is looking for Symbol properties - these cannot be coerced into Strings (it will throw) but can be used as properties for Objects. As far as I can see this behaviour would break code like:

```js
var ob = {}; ob[Symbol.for('foo')] = 'foo';
chai.expect(ob).to.have.property(Symbol.for('foo'))
```
([See working example](https://tonicdev.com/56b8cbbab828ee0d0099a569/56c45ca76f7c150d00ae8303))

keithamus:
Actually scratch that comment, after re-reading I see this only effects a path which uses `Object.keys` which doesn't return Symbols, so this should be fine.

keithamus:
Okay, LGTM :smile: Thanks for the work @stpeng

Issue #632, opened 2016-03-04
Add assertions for ES6 Maps and Sets


tiemevanveen:
@keithamus:
> @tiemevanveen you should raise a new issue - we can discuss the design of this and work towards adding it as a feature :smile:

------------------ 

It would be cool if there would be assertions for ES6 Maps (and Sets).

For example something like:

    expect(myMap).to.have.key('some_key');

Currently one could use:

    expect(myMap.has('some_key'), 'map should have some_key').to.be.true;

But that doesn't work great..

keithamus:
We already have a [`.keys` assertion](https://github.com/chaijs/chai/blob/master/lib/chai/core/assertions.js#L1115-L1235) so it makes sense to extend this to detect maps and sets. You would need to wrap the existing [key extraction code](https://github.com/chaijs/chai/blob/master/lib/chai/core/assertions.js#L1174-L1180) into the `else` of an `if` that type detects maps and sets. So the result would look a little like:

```js
var len = keys.length,
    any = flag(this, 'any'),
    all = flag(this, 'all'),
    actual,
    expected;
if (_.type(obj) === 'map') {
  actual = Array.from(obj.keys());
  expected = keys;
} else {
  actual = Object.keys(obj);
  expected = keys.map(String);
}
```

A good PR would also include some test coverage, and extend the existing documentation (found in the comments above the method).

lucasfcosta:
Hi everyone,
As I was working on this I had a doubt about the type of keys: **on Sets and Maps we can have virtually anything as a key**, so are we going to accept these types to be passed as arguments?

TL;DR
Should this be valid?
```js
var myMap = new Map().set({}, 'value');
expect(myMap).to.have.key({});
```

keithamus:
@lucasfcosta awesome! Look forward to the PR :smile:. If it is a breaking change, make it against `4.x.x` - but it seems as though it'd be a feature without any breakages, as trying to assert if a map has a key will currently only cause an error; so PRing against `master` sounds like a good idea.

bnord01:
@lucasfcosta 
> Should this, for example, be allowed?
```javascript
var myMap = new Map().set({}, 'value').set({lol: 'lel'}, 'laugh');
expect(myMap).to.have.all.keys({}, {lol: 'lel'});
```

I would expect this to fail. Keys in maps and values in sets are by reference and checked that way using `has`
```javascript
var k1 = {}, k2={}, myMap = new Map().set(k1,1)
[...]

PR #633, opened 2016-03-07
Keys for maps and sets

lucasfcosta:
Hello everyone, this PR aims to solve #632.

Here's a summary of the changes I made:

1. I am using `.some()` instead of `.filter()` when checking for `any` keys. There is no need to calculate the full size of intersections (`intersection > 0`), if we have one it's already bigger than `0` therefore we don't need to iterate through everything.
2. I had to change the `mixedArgs` message because **`keys` assertion now accepts objects as keys** (as I've said on #632).
3. When checking maps and sets' keys we need to compare objects, that's why I'm using `_.eql(expectedKey, actualKey)`.
4. I noticed that the `assert` interface had no `keys` related methods, so I added every possible combination (any/all/contains/have - Please notice that some of these combinations have the same meaning, as explained on the `keys` method description, so I added only the ones with different effects)
5. I included tests for these changes on all three interface's test files.
6. Every method I've touched got it's JSDoc comments updated (examples were included too).

PS.: When creating these tests I've had to use `if (Maps !== undefined)` and `if (Sets !== undefined)` because apparently PhantomJS 1.9 uses an engine that hasn't implemented these specs yet.

Feel free to point possible mistakes and give feedback or new ideas.

Thanks everyone

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -1111,7 +1111,6 @@ module.exports = function (chai, _) {
     );
   });
 
-
   /**
    * ### .keys(key1, [key2], [...])
    *
@@ -1133,6 +1132,10 @@ module.exports = function (chai, _) {
    * match the number of keys passed in (in other words, a target object must
    * have all and only all of the passed-in keys).
    *
+   * You can also use this on Sets and Maps. Please notice that these two types
+   * can have objects as keys, so, in this case, you can pass multiple objects as arguments if
+   * you want to.
+   *
    *     expect({ foo: 1, bar: 2 }).to.have.any.keys('foo', 'baz');
    *     expect({ foo: 1, bar: 2 }).to.have.any.keys('foo');
    *     expect({ foo: 1, bar: 2 }).to.contain.any.keys('bar', 'baz');
@@ -1142,7 +1145,10 @@ module.exports = function (chai, _) {
    *     expect({ foo: 1, bar: 2 }).to.have.all.keys({'bar': 6, 'foo': 7});
    *     expect({ foo: 1, bar: 2, baz: 3 }).to.contain.all.keys(['bar', 'foo']);
    *     expect({ foo: 1, bar: 2, baz: 3 }).to.contain.all.keys({'bar': 6});
-   *
+   *     expect(new Map([[{objKey: 'value'}, 'value'], [1, 2]])).to.contain.key({objKey: 'value'});
+   *     expect(new Map([[{objKey: 'value'}, 'value'], [1, 2]])).to.contain.any.keys([{objKey: 'value'}, {anotherKey: 'anotherValue'}]);
+   *     expect(new Map([['firstKey', 'firstValue'], [1, 2]])).to.contain.all.keys('firstKey', 1);
+   *     expect(new Set([['foo', 'bar'], ['example', 1]])).to.have.any.keys('foo');
    *
    * @name keys
    * @alias key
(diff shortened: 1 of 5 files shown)


lib/chai/core/assertions.js
```diff
@@ -1142,6 +1145,9 @@ module.exports = function (chai, _) {
    *     expect({ foo: 1, bar: 2 }).to.have.all.keys({'bar': 6, 'foo': 7});
    *     expect({ foo: 1, bar: 2, baz: 3 }).to.contain.all.keys(['bar', 'foo']);
    *     expect({ foo: 1, bar: 2, baz: 3 }).to.contain.all.keys({'bar': 6});
+   *     expect(new Map([[{objKey: 'value'}, 'value'], [1, 2]])).to.contain.key({objKey: 'value'});
+   *     expect(new Map([['firstKey', 'firstValue'], [1, 2]])).to.contain.all.keys('firstKey', 1);
+   *     expect(new Set([['foo', 'bar'], ['example', 1]])).to.have.any.keys('foo');
```
keithamus: Would be good to have an example of passing multiple objects as keys here.

lib/chai/core/assertions.js
```diff
@@ -1185,17 +1204,29 @@ module.exports = function (chai, _) {
 
     // Has any
     if (any) {
-      var intersection = expected.filter(function(key) {
-        return ~actual.indexOf(key);
+      ok = expected.some(function(expectedKey) {
+        if (isMapOrSet) {
+          return actual.some(function(actualKey) {
+            return _.eql(expectedKey, actualKey);
```
keithamus: If I'm not mistaken; _.eql will use deep equality here. I'm not sure we want deep equality without the `.deep` qualifier. A `===` might work better here.

lib/chai/core/assertions.js
```diff
@@ -1185,17 +1204,29 @@ module.exports = function (chai, _) {
...
     }
 
     // Has all
     if (all) {
-      ok = keys.every(function(key){
-        return ~actual.indexOf(key);
+      ok = expected.every(function(expectedKey) {
+        if (isMapOrSet) {
+          return actual.some(function(actualKey) {
+            return _.eql(expectedKey, actualKey);
```
keithamus: Same here (`_.eql`)

keithamus:
Only a couple of minor points above. Otherwise this looks _awesome_ @lucasfcosta. Great work, as usual :smile:

lib/chai/core/assertions.js
lucasfcosta: Yes, it would be good!
Totally forgot it, thanks for pointing this

lucasfcosta:
Thanks, @keithamus!
Your feedback is always valuable. :smile: 

Well, I'm aware `_.eql` uses deep equality and I've used it because when, for example, a Map has an object as Key and I compare the objects (even if they're equal) I will get `false` as result because `===` checks for referential equality, therefore this, for example, wouldn't work:

```js
expect(new Map([[{objKey: 'value'}, 'value'], [1, 2]])).to.contain.key([{objKey: 'value'}]);
```

This will fail, because although the object used as key on the map is depply equal to the one passed to the `.key` assertion they're not the same so `===` will return false. If you want to check for referential equality only then I feel like we should use `===`.

keithamus:
@lucasfcosta sure, I expected you to say as much - I should have elaborated more in my original comment, but got a little side tracked with other things.

Deep equality is going to be _sometimes_ desirable within Maps, but _sometimes_ referential equality will be desirable. If we start with using referential equality, we can always move to deep equality later - by adding the `.deep` qualifier. We won't be able to (easily) do the same the other way around.

lucasfcosta:
Ah, I understand now and I totally agree, it really seems to be the right step. I haven't thought about it before.
I can also implement the deep check when there is a `deep` flag if you want me to.

Thanks for the explanation, I'll make sure to update this PR with the changes we've talked about ASAP :smiley:

keithamus:
We can tackle the deep flag in another PR, to keep this one simple :smile:

lucasfcosta:
This one is done! :tada: 
I look forward to implementing the `deep` flag support on another PR later, if you don't mind. I feel like it would be very useful when it comes to Maps and Sets.

And thanks again for your support!

brutalcrozt:
@lucasfcosta awesome :smile:

keithamus:
Fantastic work as usual @lucasfcosta. Feel free to make a PR adding support for deep Map/Set keys :smile:

lucasfcosta:
Thanks @keithamus, I'll certainly do it :smile:

brutalcrozt:
close #632 ?

Issue #634, opened 2016-03-08
`doesNotThrow()` doesn't match error messages


brunops:
Hi, according to the documentation, `.doesNotThrow` should match error messages, but it's currently not doing so. See the comments in the example.

```js
var assert = require('chai').assert;

var fn = function () {
    throw new Error('foo');
}

describe('`throw` and `doesNotThrow`', function () {

    // this fails - should PASS
    it('does not throw "bar" error', function () {
        assert.doesNotThrow(fn, 'bar');
    });

    // this passes - OK
    it('throws "foo" error', function () {
        assert.throws(fn, 'foo');
    });

    // this fails - OK
    it('throws "foo" error', function () {
        assert.throws(fn, 'bar');
    });
});
```

lucasfcosta:
Since we've got #683 merged, this one can now be closed.
Now we match both `Error` and `message` if both args exist.

PR #635, opened 2016-03-08
Apply the same logic in `.throws()` to `.doesNotThrow()` so it matche…

brunops:
…it matches error messages

- `.doesNotThrow()` is currently not matching the error message, although its description says so
- Changed `.Throw` to `.throw` so its consistent with `.throws()`
- Added test
- See #634 for examples

### DIFF
--- a/lib/chai/interface/assert.js
+++ b/lib/chai/interface/assert.js
@@ -1140,13 +1140,13 @@ module.exports = function (chai, util) {
    * @api public
    */
 
-  assert.doesNotThrow = function (fn, type, msg) {
-    if ('string' === typeof type) {
-      msg = type;
-      type = null;
+  assert.doesNotThrow = function (fn, errt, errs, msg) {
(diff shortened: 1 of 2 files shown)


brunops:
Just saw #551, this issue might be fixed there, feel free to close this if that's the case.

lucasfcosta:
@keithamus Hmmm, this one is interesting, I think we have a problem with the assert interface.
It could also accept an error mesage.
The wrong behavior has been fixed, but the `assert` interface needs some polishment.
I'll take my time this afternoon to finish everything related to #683, so I might send PR for this too.
Let's keep this one open for now just keep this on our sight.

meeber:
@lucasfcosta I've lost track where we stand on this. Do we still have work to do with the error stuff?

lucasfcosta:
Thanks for noticing, @meeber!
This one is done. #742 solves it.

Issue #562, opened 2015-12-07
`an` language chain affects `length` language chain?


benmosher:
Chai 3.4.1, Node 4.2.2:
```js
expect([1, 2, 3]).to.be.an.instanceOf(Array).and.to.have.length.below(1000)
```
... should work, but fails because `length` returns `0`.

Removing `an` from the chain, leaving:
```js
expect([1, 2, 3]).to.be.instanceOf(Array).and.to.have.length.below(1000)
```
works as expected.

Seems similar to an issue reported late in #427 by valscion.

keithamus:
Hey @benmosher sorry for the late reply.

You're totally right this does seem to be an issue. It looks as though `an`, being an actual assertion function, overrides the `length` property (because its a function). And so `length` is `0` in the first example, while in the second it is the assertion.

This seems to be to do with how the chain is formed. [`addChainableMethod`](https://github.com/chaijs/chai/blob/master/lib/chai/utils/addChainableMethod.js) does some special magic to return _both_ the method and the chain.

All normal methods just return `this` - which obviously changes between `addChainableMethod` and `addMethod`. To illustrate:

```js
> chai.expect([1, 2, 3]).to.be.an.instanceof(Array)
[Function: assert]
> chai.expect([1, 2, 3]).to.be.instanceof(Array)
Assertion {
  __flags: { ssfi: [Function], object: [ 1, 2, 3 ], message: undefined } }
```

I'm not entirely sure how we can fix this, certainly without sweeping architectural changes. We could change how [`addMethod`](https://github.com/chaijs/chai/blob/master/lib/chai/utils/addMethod.js#L36-L44) works by, instead of returning `this`, returning a new Assertion chain with the flags copied over - for example [L42](https://github.com/chaijs/chai/blob/master/lib/chai/utils/addMethod.js#L42) would have to become:

```js
return result === undefined ? transferFlags(new chai.Assertion()) : result
```

I'm not entirely sure that would work though. It'd certainly need lots of extra tests to back it up.

I'll mark this as `pull-request-wanted` and `hard-fix` for now - if you want to explore the above solution in a PR, it'd very much be welcome. If it doesn't work then we can revisit this issue and look for other solutions.

benmosher:
Fair enough. I'm not sure if I'll have bandwidth to attack this, especially given the simplicity of the workaround (just don't chain `a`/`an` with `length`).

Mostly wanted to get it documented for others, as it was pretty surprising when one of my teammates stumbled onto it.

[...]

PR #642, opened 2016-03-16
addMethod returns a new assertion with flags copied over instead of this

lucasfcosta:
This aims to solve #562.

Special thanks to @keithamus for the whole idea here, I just made some minor adjustments.

This PR should include:
- [x] `addMethod` returning a `new Assertion` with flags copied over
- [x] Tests for it
- [x] Fix [this](https://github.com/chaijs/chai/pull/642#discussion_r56754878) - EDIT: Unfortunately there is one case we won't be able to cover, so this is only *partially* resolved. [Please read this](https://github.com/chaijs/chai/pull/642#issuecomment-198979726).

I'm currently writing the tests and I'm not sure if they're complete enough.
I was thinking about testing chaining some assertions as described on #562, testing the type of their return and if methods are really returning a new Assertion with the flags which should have been copied. Is this enough? If not, what else should be verified?

EDIT: This needs a review, please see [this](https://github.com/chaijs/chai/pull/642#discussion_r56754878). I'll take a further look at it.

### DIFF
--- a/lib/chai/utils/addMethod.js
+++ b/lib/chai/utils/addMethod.js
@@ -5,6 +5,7 @@
  */
 
 var config = require('../config');
+var transferFlags = require('./transferFlags');
 
 /**
  * ### .addMethod (ctx, name, method)
@@ -39,6 +40,10 @@ module.exports = function (ctx, name, method) {
     if (old_ssfi && config.includeStack === false)
       flag(this, 'ssfi', ctx[name]);
     var result = method.apply(this, arguments);
-    return result === undefined ? this : result;
+
+    var newAssertion = new chai.Assertion();
+    transferFlags(this, newAssertion);
+
+    return result === undefined ? newAssertion: result;
   };
 };
--- a/lib/chai/utils/addProperty.js
+++ b/lib/chai/utils/addProperty.js
@@ -35,6 +35,8 @@ var transferFlags = require('./transferFlags');
  */
 
 module.exports = function (ctx, name, getter) {
+  getter = getter === undefined ? new Function() : getter;
+
   Object.defineProperty(ctx, name,
     { get: function addProperty() {
         var old_ssfi = flag(this, 'ssfi');
--- a/lib/chai/utils/addProperty.js
+++ b/lib/chai/utils/addProperty.js
@@ -6,6 +6,7 @@
 
 var config = require('../config');
 var flag = require('./flag');
+var transferFlags = require('./transferFlags');
 
 /**
  * ### addProperty (ctx, name, getter)
@@ -41,7 +42,10 @@ module.exports = function (ctx, name, getter) {
           flag(this, 'ssfi', addProperty);
 
         var result = getter.call(this);
-        return result === undefined ? this : result;
+        var newAssertion = new chai.Assertion();
+        transferFlags(this, newAssertion);
+
+        return result === undefined ? newAssertion: result;
       }
     , configurable: true
   });
--- a/test/utilities.js
+++ b/test/utilities.js
@@ -232,6 +232,46 @@ describe('utilities', function () {
     expect(expect('foo').result()).to.equal('result');
   });
 
+  it('addMethod returns new assertion with flags copied over', function () {
+    var assertionConstructor;
+
+    chai.use(function(_chai, utils) {
+      assertionConstructor = _chai.Assertion;
+      _chai.Assertion.addMethod('returnNewAssertion', function () {
+        utils.flag(this, 'mySpecificFlag', 'value1');
+        utils.flag(this, 'ultraSpecificFlag', 'value2');
+      });
+
+      _chai.Assertion.addMethod('checkFlags', function() {
+        this.assert(
+            utils.flag(this, 'mySpecificFlag') === 'value1' &&
+            utils.flag(this, 'ultraSpecificFlag') === 'value2'
+          , 'expected assertion to have specific flags'
+          , "this doesn't matter"
+        );
+      });
+    });
+
+    assertion1 = expect('foo');
+    assertion2 = assertion1.to.returnNewAssertion();
+
+    // Checking if a new assertion was returned
+    expect(assertion1).to.not.be.equal(assertion2);
+
+    // Check if flags were copied
+    assertion2.checkFlags();
+
+    // Checking if it's really an instance of an Assertion
+    expect(assertion2).to.be.instanceOf(assertionConstructor);
+
+    // Test chaining `.length` after a method to guarantee it is not a function's `length`
+    var anAssertion = expect([1, 2, 3]).to.be.an.instanceof(Array);
+    expect(anAssertion.length).to.be.an.instanceof(assertionConstructor);
+
+    var anotherAssertion = expect([1, 2, 3]).to.have.a.lengthOf(3).and.to.be.ok;
+    expect(anotherAssertion.length).to.be.an.instanceof(assertionConstructor);
+  });
+
   it('overwriteMethod', function () {
     chai.use(function (_chai, _) {
       expect(_chai.Assertion).to.respondTo('eqqqual');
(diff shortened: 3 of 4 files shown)


lucasfcosta:
Here are the tests I thought were right for this.
If you guys think of more things that should be covered by this test, please share.

test/utilities.js
```diff
@@ -232,6 +232,42 @@ describe('utilities', function () {
...
+
+    // Checking if a new assertion was returned
+    expect(assertion1).to.not.be.equal(assertion2);
+
+    // Checking if it's really an instance of an Assertion
+    expect(assertion2).to.be.instanceOf(assertionConstructor);
+
+    // If it is, calling length on it should return an assertion, not a function
+    var anAssertion = expect([1, 2, 3]).to.be.an.instanceof(Array);
+    expect(anAssertion.length).to.be.an.instanceof(assertionConstructor);
```
keithamus: Could you please add some assertions that feature the `an` keyword, and also feature the `length` keyword, similar to the original issue.

```js
expect([1, 2, 3]).to.be.an.instanceOf(Array).and.to.have.length.below(1000)
```

lucasfcosta: Well thought @keithamus.
This fix appears to work with the expression you posted, but I went to the issue to test other examples and this appears to fail:

```js
expect(my_object).to.have.a.deep.property("foo.bar").that.is.an("array").with.a.length.of.at.least(1);
```

The reason is the same you explained on the issue, we are getting back an `assert` function instead of an `Assertion` object.

I'll take a further look at this to find the exact reason.

Thanks again for your feedback, it was really useful :smiley:

keithamus: Hey @lucasfcosta thanks for doing that, and also awesome work adding some more tests and discovering another issue :smile: 

It looks like in that case, the `.a.` part turns the Assertion into a function, from which it doesn't recover. To wit:

```js
> chai.expect({foo: [1,2,3]}).to
Assertion {
  __flags:
   { ssfi: [Function: addProperty],
     object: { foo: [Object] },
     message: undefined } }
> chai.expect({foo: [1,2,3]}).to.have
Assertion {
  __flags:
   { ssfi: [Function: addProperty],
     object: { foo: [Object] },
     message: undefined } }
> chai.expect({foo: [1,2,3]}).to.have.a
[Function: assert]
> chai.expect({foo: [1,2,3]}).to.have.a.deep
[Function: assert]
```

Luckily, it looks like the fix will be pretty much the same thing. [L44 of addProperty](https://github.com/chaijs/chai/blob/master/lib/chai/utils/addProperty.js#L44) has the same ternary.

lucasfcosta: Ah, that makes sense.
Thanks for the insight here.

I was also doing some research on [this](https://github.com/chaijs/chai/blob/master/lib/chai/utils/addChainableMethod.js#L85) too, shouldn't we create another assertion and transfer flags to it instead of returning `this`?

I'm not sure this is needed, I will do some further experiments with this, but it may be good to prevent further errors related to this.

lucasfcosta:
Well, I've done some further studies on this and I've discovered that the explanation for why this (`expect(my_object).to.have.a.deep.property("foo.bar").that.is.an("array").with.a.length.of.at.least(1)`) is not working is the following:

We've got [this line into our `addChainableMethod.js` file](https://github.com/chaijs/chai/blob/master/lib/chai/utils/addChainableMethod.js#L108) which returns a `function` as the getter for the property being added, therefore, when calling `.length` right after using the `a` getter we get back a function (the function which executes the `chainingBehavior`, [as you can see here](https://github.com/chaijs/chai/blob/master/lib/chai/utils/addChainableMethod.js#L80), so it has got its own `.length` property, instead of the `.length` property of the assertion.

What I've done on this PR partially solves our problem because when we return a `new Assertion` instead of this, it means we're returning an object which is not a `function` and therefore does not have the `length` property, that is why `an.<something_here_will_return_new_Assertion>.length` works (because we are "transforming" the assert function into another object which is not a function) and `a.length` does not (because `a` immediately returns a function, which has a `length` property).

--------------------------------------------

## ** How we could solve this: **

This would be a very, very easy fix if we could just do:  `delete length` before returning `assert` [here](https://github.com/chaijs/chai/blob/master/lib/chai/utils/addChainableMethod.js#L108). This would cause the `.length` property to be called from the `prototype`, which means it will call the assertion instead of the function's length property.

**Unfortunately this is not available on all browsers** because the `.length` property is configurable only on Firefox(Gecko) >= 37 and on Node.js.

**Being able to delete `.length` would enable us to solve this with a single line of code**, unfortunately this is not possible on older environments (or even on PhantomJS 2.1).

Currently we can only solve this for some cases, like the ones described above, in which we return a `new Assertion` when using `addMethod` or `addProperty`.

I hope I was clear, but if I wasn't please ask.

keithamus:
Hey @lucasfcosta thanks for the awesome deep dive! Here's my commentary on this:

> Unfortunately this is not available on all browsers because the .length property is configurable only on Firefox(Gecko) >= 37 and on Node.js.

This is because `Function.length` (and `Function.name`) are non-configurable as per ES5 spec ([ES5$15.3.3.1](https://es5.github.io/#x15.3.3.1) & [ES5$15.3.3.2](https://es5.github.io/#x15.3.3.2)) but in ES6 it was decided that they should be configurable ([$19.2.4.1](http://www.ecma-international.org/ecma-262/6.0/index.html#sec-function-instances-name) & [$19.2.4.1](http://www.ecma-international.org/ecma-262/6.0/index.html#sec-function-instances-length)). Being configurable also means they are delete-able ([$9.1.10](http://www.ecma-international.org/ecma-262/6.0/index.html#sec-ordinary-object-internal-methods-and-internal-slots-delete-p) if you're curious).

According to [Kangax](https://kangax.github.io/compat-table/es6/) (here is the [name is configurable test](https://kangax.github.io/compat-table/es6/#test-function_name_property_isn't_writable,_is_configurable) and here is the [length is configurable test](https://kangax.github.io/compat-table/es6/#test-miscellaneous_function_'length'_is_configurable)), Edge, Firefox, Chrome and Node (4 & 5) support this, but older browsers (IE<=11, older Android versions not using Chrome, all Safaris including iOS) do not.

> For now....

Older IEs and Safari are basically immovable at this point. Its disappointing but it is what it is. Fixing this by ensuring we don't end up returning functions in the chain, as you've done in this PR - is useful of its own merit. 

If we looked at deleting the properties (`length` and `name`) from the functions to help with newer browsers, we open ourselves up to issues from older browsers failing on these assertions. Consistent behaviour (even if it is consistently broken) is probably preferable over inconsistent behaviour - especially in a test framework. It should be our top priority that users should never have to debug Chai - they're already too busy debugging their code!

lucasfcosta:
@keithamus thanks for the info on the `Function.length` and `Function.name` props! All the links to the spec were very useful. Awesome job!

I totally agree with you when you say that consistent behaviour is preferable, that's why this PR has only non-breaking and consistent changes, but, as I said, there's no way to solve chaining `.length` right after `.a` without problems on older browsers, so I kept it out of these commits.

This **partially** solves the problem, but it's still (in my opinion) better than nothing. What do you think about it?

Thanks for your feedback and for your help, without you this PR wouldn't have been possible :smile:

keithamus:
LGTM :smile:

lib/chai/utils/addMethod.js
```diff
@@ -39,6 +40,10 @@ module.exports = function (ctx, name, method) {
     if (old_ssfi && config.includeStack === false)
       flag(this, 'ssfi', ctx[name]);
     var result = method.apply(this, arguments);
-    return result === undefined ? this : result;
+
+    var newAssertion = new chai.Assertion();
```
Turbo87: I'm getting `ReferenceError: chai is not defined` thrown here

lib/chai/utils/addProperty.js
```diff
@@ -41,7 +44,10 @@ module.exports = function (ctx, name, getter) {
           flag(this, 'ssfi', addProperty);
 
         var result = getter.call(this);
-        return result === undefined ? this : result;
+        var newAssertion = new chai.Assertion();
```
Turbo87: same here: `ReferenceError: chai is not defined`

lib/chai/utils/addMethod.js
Turbo87: I guess this problem has been shadowed by `global.chai = require('../..');` in all tests...

keithamus: Good catch, thanks @Turbo87. Fancy making a PR for this?

keithamus: Awesome, I see you already have :tada: :+1:

PR #651, opened 2016-03-27
Update expect.js (test)

unional:
The err() takes only two params


keithamus:
Looks like this was added in #96. Not sure why - but this LGTM. @lucasfcosta want to double check this?

lucasfcosta:
I think someone at #96 may just have tought this was about the `.throw` assertion, since that PR was about error messages, but, anyway, I just took a look at [these files](https://github.com/chaijs/chai/tree/master/test/bootstrap) and @unional seems right.

This LGTM too.

PR #661, opened 2016-04-01
Fix stacktraces for overwritten properties and methods

### LINKED ISSUE #659: stack trace wrong for overridden properties
let's say I use the following assertion:

```js
expect(true).to.be.undefined;
```

I get the following error with the expected stack trace:

```
AssertionError: expected true to be undefined
 at Context.<anonymous> (test/test.js:18:25)
```

---

If I overwrite the `undefined` property like this:

```js
Assertion.overwriteProperty('undefined', function(_super) {
  return function() {
    _super.call(this);
  };
});
```

I'm getting the following error with an extended stack trace:

```
AssertionError: expected true to be undefined
 at Assertion.<anonymous> (src/chai-files.js:52:12)
 at Assertion.Object.defineProperty.get (node_modules/chai/lib/chai/utils/overwriteProperty.js:50:37)
 at Context.<anonymous> (test/test.js:18:25)
```

---

If I overwrite the `undefined` property like this:

```js
Assertion.overwriteProperty('undefined', function(_super) {
  return function() {
    this.assert(false, 'foo', 'bar');
  };
});
```

I'm getting the following error with a missing stack trace:

```
AssertionError: foo
```

Turbo87:
This PR resolves #659 

Since testing overwritten properties/methods doesn't seems to be possible at the moment I tested this manually and it seems to work now. It would be good if this could be verified by someone else before merging though!

/cc @keithamus

### DIFF
--- a/lib/chai/utils/addChainableMethod.js
+++ b/lib/chai/utils/addChainableMethod.js
@@ -10,7 +10,6 @@
 
 var transferFlags = require('./transferFlags');
 var flag = require('./flag');
-var config = require('../config');
 
 /*!
  * Module variables
@@ -79,7 +78,7 @@ module.exports = function (ctx, name, method, chainingBehavior) {
 
         var assert = function assert() {
           var old_ssfi = flag(this, 'ssfi');
-          if (old_ssfi && config.includeStack === false)
+          if (old_ssfi)
             flag(this, 'ssfi', assert);
           var result = chainableBehavior.method.apply(this, arguments);
           return result === undefined ? this : result;
--- a/lib/chai/utils/addProperty.js
+++ b/lib/chai/utils/addProperty.js
@@ -5,7 +5,6 @@
  */
 
 var chai = require('../../chai');
-var config = require('../config');
 var flag = require('./flag');
 var transferFlags = require('./transferFlags');
 
@@ -41,7 +40,7 @@ module.exports = function (ctx, name, getter) {
   Object.defineProperty(ctx, name,
     { get: function addProperty() {
         var old_ssfi = flag(this, 'ssfi');
-        if (old_ssfi && config.includeStack === false)
+        if (old_ssfi)
           flag(this, 'ssfi', addProperty);
 
         var result = getter.call(this);
--- a/lib/chai/utils/addProperty.js
+++ b/lib/chai/utils/addProperty.js
@@ -39,8 +39,9 @@ module.exports = function (ctx, name, getter) {
 
   Object.defineProperty(ctx, name,
     { get: function addProperty() {
+        var keep_ssfi = flag(this, 'keep_ssfi');
         var old_ssfi = flag(this, 'ssfi');
-        if (old_ssfi)
+        if (!keep_ssfi && old_ssfi)
           flag(this, 'ssfi', addProperty);
 
         var result = getter.call(this);
--- a/lib/chai/utils/overwriteMethod.js
+++ b/lib/chai/utils/overwriteMethod.js
@@ -4,6 +4,8 @@
  * MIT Licensed
  */
 
+var flag = require('./flag');
+
 /**
  * ### overwriteMethod (ctx, name, fn)
  *
@@ -46,7 +48,15 @@ module.exports = function (ctx, name, method) {
     _super = _method;
 
   ctx[name] = function () {
+    var keep_ssfi = flag(this, 'keep_ssfi');
+    var old_ssfi = flag(this, 'ssfi');
+    if (!keep_ssfi && old_ssfi)
+      flag(this, 'ssfi', ctx[name]);
+
+    flag(this, 'keep_ssfi', true);
     var result = method(_super).apply(this, arguments);
+    flag(this, 'keep_ssfi', false);
+
     return result === undefined ? this : result;
   }
 };
--- a/test/utilities.js
+++ b/test/utilities.js
@@ -320,6 +320,60 @@ describe('utilities', function () {
     expect(expect('foo').result()).to.equal('result');
   });
 
+  describe('overwriteMethod', function () {
+    before(function() {
+      chai.config.includeStack = false;
+
+      chai.use(function(_chai, _) {
+        _chai.Assertion.addMethod('four', function() {
+          this.assert(this._obj === 4, 'expected #{this} to be 4', 'expected #{this} to not be 4', 4);
+        });
+
+        _chai.Assertion.overwriteMethod('four', function(_super) {
+          return function() {
+            if (typeof this._obj === 'string') {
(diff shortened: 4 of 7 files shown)


Turbo87:
@keithamus any comments on this?

lib/chai/utils/addChainableMethod.js
```diff
@@ -79,7 +78,7 @@ module.exports = function (ctx, name, method, chainingBehavior) {
 
         var assert = function assert() {
           var old_ssfi = flag(this, 'ssfi');
-          if (old_ssfi && config.includeStack === false)
+          if (old_ssfi)
```
keithamus: Could you please explain why you've removed the `config.includeStack` part here?

lib/chai/utils/addProperty.js
```diff
@@ -40,8 +39,9 @@ module.exports = function (ctx, name, getter) {
 
   Object.defineProperty(ctx, name,
     { get: function addProperty() {
+        var keep_ssfi = flag(this, 'keep_ssfi');
```
keithamus: Could you also explain why we need the `keep_ssfi` flag. As far as I can see it is used to to override ssfi when we chain on properties - is this right?

lib/chai/utils/addChainableMethod.js
Turbo87: That flag is already checked in the assert() method and checking it here has no real benefit aside from making the logic more complicated

lib/chai/utils/addProperty.js
Turbo87: imagine expect(true).to.be.false with false overwritten:

- expect sets ssfi
- to overwrites ssfi
- be overwrites ssfi
- custom false overwrites ssfi
- _super false overwrites ssfi

but actually that last part shouldn't happen because you don't want the _super in the stack trace. setting keep_ssfi during the _super prevents that from happening.

keithamus:
LGTM, I'd like @lucasfcosta to review this one also though.

Turbo87:
Sure, it needs a bit of brain twisting to understand what's going on :wink:

lib/chai/utils/overwriteMethod.js
```diff
@@ -46,7 +48,15 @@ module.exports = function (ctx, name, method) {
     _super = _method;
 
   ctx[name] = function () {
+    var keep_ssfi = flag(this, 'keep_ssfi');
+    var old_ssfi = flag(this, 'ssfi');
+    if (!keep_ssfi && old_ssfi)
+      flag(this, 'ssfi', ctx[name]);
+
+    flag(this, 'keep_ssfi', true);
```
lucasfcosta: Hi @Turbo87, I just want to confirm I understood the whole thing here:

Whenever you overwrite a method you set `keep_ssfi` to true before calling `_super` so you won't get that call on the stack due to those checks you added to `addMethod`, `addChainableMethod` and `addProperty`, am I right?

Turbo87: correct

lib/chai/utils/addProperty.js
lucasfcosta: Shouldn't we test if `_super` is really in the stack trace when we've got properties/methods overwritten?
What do you guys think about this?

Tests are okay, now, but maybe we could have this too.

lucasfcosta:
Hello everyone, first of all, thanks for the PR @Turbo87, great job here :smile: !

I needed to pay a lot of attention to really understand how to handle the Stack Trace properly, but I think I've got it. This is looking good, in general, anyway, I just added two notes to the code just to ensure we're going in the right direction here.

My two notes are [here](https://github.com/chaijs/chai/pull/661#discussion_r58314409) and [here](https://github.com/chaijs/chai/pull/661#discussion_r58314662).

lib/chai/utils/addProperty.js
keithamus: Yup, great idea :+1: :100:

Turbo87: @lucasfcosta but sure I understand what you meant?

lucasfcosta: @Turbo87, sorry but I didn't understand what you just said :(

Tests for this should be similar to what we already have, we just need to overwrite the property and them check the stack trace to make sure it does not include that `_super` call. Basically you've got to ensure the second example you posted on [this issue](https://github.com/chaijs/chai/issues/659) won't happen.

Turbo87: damn autocorrect... that was "not sure I understand what you meant" 😉

Turbo87:
@lucasfcosta I've added a few tests to verify that the stack traces work correctly when overwriting properties and methods

test/utilities.js
```diff
@@ -320,6 +320,60 @@ describe('utilities', function () {
...
+    it('calling _super has correct stack trace', function() {
+      try {
+        expect(5).to.be.four();
+        expect(false, 'should not get here because error thrown').to.be.ok;
+      } catch (err) {
+        // not all browsers support err.stack
+        // Phantom does not include function names for getter exec
+        if ('undefined' !== typeof err.stack && 'undefined' !== typeof Error.captureStackTrace) {
+          expect(err.stack).to.include('utilities.js');
+          expect(err.stack).to.not.include('overwriteProperty');
```
lucasfcosta: I believe you should be checking if it does not include `overwriteMethod` instead of `overwriteProperty` here, shouldn't you?

test/utilities.js
```diff
@@ -320,6 +320,60 @@ describe('utilities', function () {
...
+    it('overwritten behavior has correct stack trace', function() {
+      try {
+        expect('five').to.be.four();
+        expect(false, 'should not get here because error thrown').to.be.ok;
+      } catch (err) {
+        // not all browsers support err.stack
+        // Phantom does not include function names for getter exec
+        if ('undefined' !== typeof err.stack && 'undefined' !== typeof Error.captureStackTrace) {
+          expect(err.stack).to.include('utilities.js');
+          expect(err.stack).to.not.include('overwriteProperty');
```
lucasfcosta: The same applies to this line.
(`overwriteMethod` instead of `overwriteProperty`)

lucasfcosta:
Great job @Turbo87, I just think you accidentally looked for `overwriteProperty` on the stack-trace for `overwriteMethod` tests.

test/utilities.js
Turbo87: whoops, copy-paste issue. sorry!

Turbo87:
@lucasfcosta fixed

test/utilities.js
lucasfcosta: hahaha, np, happens to everyone sometimes :laughing:

lucasfcosta:
This LGTM!
Nice job!

PR #668, opened 2016-04-06
Deep flag support for keys assertion

### LINKED ISSUE #662: Inspect doesn't work properly for Maps and Sets - It displays an empty object
As I was implementing the `.deep` flag support on the `keys` assertion I noticed that `utils.inspect` isn't working properly when it comes to Maps and Sets. Instead of returning the values inside the Map or Set it returns the following string: `{}`.

You can reproduce this behavior by doing, for example:

``` js
chai.use(function (_chai, utils) {
  var inspectSet = _utilsinspect(new Set([{iAmAn: 'item'}, {thisIs: 'anotherItem'}]));
  var inspectMap = utils.inspect(new Map([[{iAmA: 'key'}, 'aValue']]))
  console.log(inspectSet); // --> {}
  console.log(inspectMap); // --> {}
});
```

What do you guys think should be the desired output when inspecting maps or sets?

IMO the ideal output would be:

``` js
chai.use(function (_chai, utils) {
  var inspectSet = _utilsinspect(new Set([{iAmAn: 'item' }, {thisIs: 'anotherItem'}]));
  var inspectMap = utils.inspect(new Map([[{iAmA: 'key' }, 'aValue']]))
  console.log(inspectSet); // Desired --> [ {iAmAn: 'item' }, { thisIs: 'anotherItem' } ]
   console.log(inspectMap); // Desired --> [ [ { iAmA: 'key' }, 'aValue' ] ]
});
```

---

I think this wouldn't be a difficult fix. Maybe adding a check alongside [these other checks](https://github.com/chaijs/chai/blob/4.x.x/lib/chai/utils/inspect.js#L177) and then calling a function to iterate through every entry and format the Map/Set members should be enough. This will be very similiar to what we already do in [`formatArray`](https://github.com/chaijs/chai/blob/4.x.x/lib/chai/utils/inspect.js#L225).

I will make sure to submit a PR for this before implementing the `deep` flag for Maps/Sets because this will be very useful to create good test cases.

Please let me know if you've got any idea or opinion about this :wink:

lucasfcosta:
This adds support for the `deep` flag on the `keys` assertion (for Maps/Sets) as I've talked to @keithamus at #633.

This PR should include:
- [x] Deep flag support for the keys assertion (Maps/Sets)
- [x] Updated doc comments with examples and a general explanation about the `deep` flag on this assertion
- [x] Tests on the `should` interface
- [x] Tests on the `expect` interface
- [x] Bindings for these on the `assert` interface
- [x] Tests on the `assert` interface

Currently the only thing missing are the bindings for these assertions on the assert interface, I didn't do it yet because I'm not sure if adding lots of bindings for this to the `assert` interface would be cool. If we added bindings for every possible combination of the `keys` assertion flags we would have:

1. hasAnyKeys
2. hasAllKeys
3. containsAllKeys
4. doesNotHaveAnyKeys
5. doesNotHaveAllKeys
6. hasAnyDeepKeys
7. hasAllDeepKeys
8. containsAllDeepKeys
9. doesNotHaveAnyDeepKeys
10. doesNotHaveAllDeepKeys

IMO we should have all of them on the `assert` interface because I think it should have the same features as the other interfaces do, but I just wanted to make sure you guys agree with me before adding all these bindings.

----------------------------------------

### Important Detail: 

Currently we need to fix #662 in order to enable all tests, so I think we may keep this open but on hold until we've got `inspect` to work properly for `maps` and `sets` so we can fully test this.
What's your opinion on this?

------------------------------------------

Thanks in advance, please feel free to point any possible mistakes/improvements :smile:

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -71,7 +71,7 @@ module.exports = function (chai, _) {
   /**
    * ### .deep
    *
-   * Sets the `deep` flag, later used by the `equal`, `include`, `members`, and
+   * Sets the `deep` flag, later used by the `equal`, `include`, `members`, `keys` and
    * `property` assertions.
    *
    *     const obj = {a: 1};
@@ -1328,6 +1328,10 @@ module.exports = function (chai, _) {
    * can have objects as keys, so, in this case, you can pass multiple objects as arguments if
    * you want to.
    *
+   * The default behavior when it comes to Maps and Sets is to use strict equality
+   * (===) to compare values. You can also use the `deep` flag if you want a deep comparison
+   * instead of a strict comparison.
+   *
    *     expect({ foo: 1, bar: 2 }).to.have.any.keys('foo', 'baz');
    *     expect({ foo: 1, bar: 2 }).to.have.any.keys('foo');
    *     expect({ foo: 1, bar: 2 }).to.contain.any.keys('bar', 'baz');
@@ -1338,9 +1342,13 @@ module.exports = function (chai, _) {
    *     expect({ foo: 1, bar: 2, baz: 3 }).to.contain.all.keys(['bar', 'foo']);
    *     expect({ foo: 1, bar: 2, baz: 3 }).to.contain.all.keys({'bar': 6});
    *     expect(new Map([[{objKey: 'value'}, 'value'], [1, 2]])).to.contain.key({objKey: 'value'});
+   *     expect(new Map([[{objKey: 'value'}, 'value'], [1, 2]])).to.contain.deep.key({objKey: 'value'});
    *     expect(new Map([[{objKey: 'value'}, 'value'], [1, 2]])).to.contain.any.keys([{objKey: 'value'}, {anotherKey: 'anotherValue'}]);
+   *     expect(new Map([[{objKey: 'value'}, 'value'], [1, 2]])).to.contain.any.deep.keys([{objKey: 'value'}, {anotherKey: 'anotherValue'}]);
    *     expect(new Map([['firstKey', 'firstValue'], [1, 2]])).to.contain.all.keys('firstKey', 1);
+   *     expect(new Map([['firstKey', 'firstValue'], [1, 2]])).to.contain.all.deep.keys('firstKey', 1);
    *     expect(new Set([['foo', 'bar'], ['example', 1]])).to.have.any.keys('foo');
+   *     expect(new Set([['foo', 'bar'], ['example', 1]])).to.have.any.deep.keys('foo');
    *
    * @name keys
    * @alias key
@@ -1351,11 +1359,14 @@ module.exports = function (chai, _) {
 
   function assertKeys (keys) {
     var obj = flag(this, 'object')
+      , isDeep = flag(this, 'deep')
       , str
+      , deepStr = ''
       , ok = true
       , mixedArgsMsg = 'when testing keys against an object or an array you must give a single Array|Object|String argument or multiple String arguments';
 
     if (_.type(obj) === 'map' || _.type(obj) === 'set') {
+      deepStr = isDeep ? 'deeply ' : '';
       actual = [];
 
       // Map and Set '.keys' aren't supported in IE 11. Therefore, use .forEach.
@@ -1402,7 +1413,11 @@ module.exports = function (chai, _) {
     if (any) {
       ok = expected.some(function(expectedKey) {
         return actual.some(function(actualKey) {
-          return expectedKey === actualKey;
+          if (isDeep) {
+            return _.eql(expectedKey, actualKey);
+          } else {
+            return expectedKey === actualKey;
+          }
         });
       });
     }
@@ -1411,7 +1426,11 @@ module.exports = function (chai, _) {
     if (all) {
       ok = expected.every(function(expectedKey) {
         return actual.some(function(actualKey) {
-          return expectedKey === actualKey;
+          if (isDeep) {
+            return _.eql(expectedKey, actualKey);
+          } else {
+            return expectedKey === actualKey;
+          }
         });
       });
 
@@ -1445,8 +1464,8 @@ module.exports = function (chai, _) {
     // Assertion
     this.assert(
         ok
-      , 'expected #{this} to ' + str
-      , 'expected #{this} to not ' + str
+      , 'expected #{this} to ' + deepStr + str
+      , 'expected #{this} to not ' + deepStr + str
       , expected.slice(0).sort(_.compareByInspect)
       , actual.sort(_.compareByInspect)
       , true
--- a/lib/chai/interface/assert.js
+++ b/lib/chai/interface/assert.js
@@ -1420,10 +1420,6 @@ module.exports = function (chai, util) {
    */
 
   assert.hasAnyKeys = function (obj, keys, msg) {
-    if (keys === undefined) {
-      new Assertion(obj, msg).to.not.have.all.keys();
-    }
-
     new Assertion(obj, msg).to.have.any.keys(keys);
   }
 
@@ -1448,10 +1444,6 @@ module.exports = function (chai, util) {
    */
 
   assert.hasAllKeys = function (obj, keys, msg) {
-    if (keys === undefined) {
-      new Assertion(obj, msg).to.not.have.all.keys();
-    }
-
     new Assertion(obj, msg).to.have.all.keys(keys);
   }
 
@@ -1480,10 +1472,6 @@ module.exports = function (chai, util) {
    */
 
   assert.containsAllKeys = function (obj, keys, msg) {
-    if (keys === undefined) {
-      new Assertion(obj, msg).to.not.have.all.keys();
-    }
-
     new Assertion(obj, msg).to.contain.all.keys(keys);
   }
 
@@ -1508,10 +1496,6 @@ module.exports = function (chai, util) {
    */
 
   assert.doesNotHaveAnyKeys = function (obj, keys, msg) {
-    if (keys === undefined) {
-      new Assertion(obj, msg).to.not.have.all.keys();
-    }
-
     new Assertion(obj, msg).to.not.have.any.keys(keys);
   }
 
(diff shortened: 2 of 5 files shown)


lucasfcosta:
Hello everyone, I'll finish the items left on this one since we're getting close to releasing `4.x.x`.
But before getting this merged I'd like to have #662 fixed just so we can have complete tests for it.
What do you think?

keithamus:
@lucasfcosta my work with refactoring loupe is quite a ways off. We can probably apple a simple fix for #662 though, to get it working with Maps and Sets for now.

lucasfcosta:
@keithamus seems like a good idea. Let's solve this one that way then so we can get this merged.
I'll fix the `inspect` embedded to Chai's core and include it into this PR too.

lucasfcosta:
This one is done!
Feel free to review and merge this if you think it's good to go.
Thanks everyone! 😄

test/assert.js
```diff
@@ -778,7 +794,7 @@ describe('assert', function () {
       var errMap = new Map();
 
       errMap.set({1: 20}, 'number');
-
+
```
vieiralucas: I see trailing spaces 😱

lucasfcosta: My vim does only removes it when there's an `.editorconfig` file in the root folder of where the file is 😢

lucasfcosta: I've just fixed that, thanks for noticing 😄

vieiralucas: haha

I added a `.editorconfig` file on my home with some rules that I always want.
I think it goes looking upwards until it finds one

vieiralucas:
Awesome work!
Specially with the docs and tests 🎉 🎉 🎉 

What could we do to remember enabling these commented tests when #662 gets solved?
Maybe solving that issue should require enabling this tests?

test/assert.js
lucasfcosta: Yes it does, great idea.

lucasfcosta:
@vieiralucas Yes, I think that would be ideal.
I'll leave a comment there so we won't forget it, thanks for noticing 😄

lib/chai/core/assertions.js
```diff
@@ -1351,11 +1359,13 @@ module.exports = function (chai, _) {
 
   function assertKeys (keys) {
     var obj = flag(this, 'object')
+      , isDeep = flag(this, 'deep')
       , str
       , ok = true
       , mixedArgsMsg = 'when testing keys against an object or an array you must give a single Array|Object|String argument or multiple String arguments';
 
     if (_.type(obj) === 'map' || _.type(obj) === 'set') {
+      str = isDeep ? 'deeply ' : '';
```
meeber: I realize messages aren't being tested yet due to an issue with inspect and sets/maps, but just noting that I believe `str` will get overwritten [here](https://github.com/chaijs/chai/blob/01963b74389f35a6f13739383015ac945987485c/lib/chai/core/assertions.js#L1430-L1436).

lib/chai/interface/assert.js
```diff
@@ -1544,6 +1544,158 @@ module.exports = function (chai, util) {
...
+   * @param {...String|Array|Object} keys
+   * @param {String} message
+   * @namespace Assert
+   * @api public
+   */
+
+  assert.hasAnyDeepKeys = function (obj, keys, msg) {
+     if (keys === undefined) {
+      new Assertion(obj, msg).to.have.any.deep.keys();
+    }
```
meeber: What if the `undefined` check was at the top of the actual `assertKeys` function instead?

lib/chai/interface/assert.js
```diff
@@ -1544,6 +1544,158 @@ module.exports = function (chai, util) {
...
+   *     assert.hasAnyDeepKeys(new Map([[{one: 'one'}, 'valueOne'], [1, 2]]), {one: 'one'});
+   *     assert.hasAnyDeepKeys(new Map([[{one: 'one'}, 'valueOne'], [1, 2]]), [{one: 'one'}, {two: 'two'}]);
+   *     assert.hasAnyDeepKeys(new Map([[{one: 'one'}, 'valueOne'], [{two: 'two'}, 'valueTwo']]), [{one: 'one'}, {two: 'two'}]);
+   *     assert.hasAnyDeepKeys(new Set([{one: 'one'}, {two: 'two'}]), {one: 'one'});
+   *     assert.hasAnyDeepKeys(new Set([{one: 'one'}, {two: 'two'}]), [{one: 'one'}, {three: 'three'}]);
+   *     assert.hasAnyDeepKeys(new Set([{one: 'one'}, {two: 'two'}]), [{one: 'one'}, {two: 'two'}]);
+   *
+   * @name doesNotHaveAllKeys
+   * @param {Mixed} object
+   * @param {...String|Array|Object} keys
```
meeber: It doesn't look to me like the `...String` style of the `keys` argument is supported here.

lib/chai/interface/assert.js
```diff
@@ -1544,6 +1544,158 @@ module.exports = function (chai, util) {
...
+   * will be used as the expected set of keys.
+   *
+   *     assert.hasAllDeepKeys(new Map([[{one: 'one'}, 'valueOne']]), {one: 'one'});
+   *     assert.hasAllDeepKeys(new Map([[{one: 'one'}, 'valueOne'], [{two: 'two'}, 'valueTwo']]), [{one: 'one'}, {two: 'two'}]);
+   *     assert.hasAllDeepKeys(new Set([{one: 'one'}]), {one: 'one'});
+   *     assert.hasAllDeepKeys(new Set([{one: 'one'}, {two: 'two'}]), [{one: 'one'}, {two: 'two'}]);
+   *
+   * @name hasAllDeepKeys
+   * @param {Mixed} object
+   * @param {...String|Array|Object} keys
```
meeber: It doesn't look to me like the `...String` style of the `keys` argument is supported here.

lib/chai/interface/assert.js
```diff
@@ -1544,6 +1544,158 @@ module.exports = function (chai, util) {
...
+      new Assertion(obj, msg).to.have.all.deep.keys();
+    }
+
+    new Assertion(obj, msg).to.have.all.deep.keys(keys);
+  }
+
+ /**
+   * ### .containsAllDeepKeys(object, [keys], [message])
+   *
+   * Asserts that `object` contains all of the `keys` provided.
```
meeber: Even though it's shown in the example, probably worth mentioning explicitly that the object can have more keys than those provided, just to contrast with `hasAll`

lib/chai/interface/assert.js
```diff
@@ -1544,6 +1544,158 @@ module.exports = function (chai, util) {
...
+   * will be used as the expected set of keys.
+   *
+   *     assert.containsAllDeepKeys(new Map([[{one: 'one'}, 'valueOne'], [1, 2]]), {one: 'one'});
+   *     assert.containsAllDeepKeys(new Map([[{one: 'one'}, 'valueOne'], [{two: 'two'}, 'valueTwo']]), [{one: 'one'}, {two: 'two'}]);
+   *     assert.containsAllDeepKeys(new Set([{one: 'one'}, {two: 'two'}]), {one: 'one'});
+   *     assert.containsAllDeepKeys(new Set([{one: 'one'}, {two: 'two'}]), [{one: 'one'}, {two: 'two'}]);
+   *
+   * @name containsAllDeepKeys
+   * @param {Mixed} object
+   * @param {...String|Array|Object} keys
```
meeber: It doesn't look to me like the `...String` style of the `keys` argument is supported here.

lib/chai/interface/assert.js
```diff
@@ -1544,6 +1544,158 @@ module.exports = function (chai, util) {
...
+   * will be used as the expected set of keys.
+   *
+   *     assert.doesNotHaveAnyDeepKeys(new Map([[{one: 'one'}, 'valueOne'], [1, 2]]), {thisDoesNot: 'exist'});
+   *     assert.doesNotHaveAnyDeepKeys(new Map([[{one: 'one'}, 'valueOne'], [{two: 'two'}, 'valueTwo']]), [{twenty: 'twenty'}, {fifty: 'fifty'}]);
+   *     assert.doesNotHaveAnyDeepKeys(new Set([{one: 'one'}, {two: 'two'}]), {twenty: 'twenty'});
+   *     assert.doesNotHaveAnyDeepKeys(new Set([{one: 'one'}, {two: 'two'}]), [{twenty: 'twenty'}, {fifty: 'fifty'}]);
+   *
+   * @name doesNotHaveAnyDeepKeys
+   * @param {Mixed} object
+   * @param {...String|Array|Object} keys
```
meeber: It doesn't look to me like the `...String` style of the `keys` argument is supported here.

lib/chai/interface/assert.js
```diff
@@ -1544,6 +1544,158 @@ module.exports = function (chai, util) {
...
+   * will be used as the expected set of keys.
+   *
+   *     assert.doesNotHaveAllDeepKeys(new Map([[{one: 'one'}, 'valueOne'], [1, 2]]), {thisDoesNot: 'exist'});
+   *     assert.doesNotHaveAllDeepKeys(new Map([[{one: 'one'}, 'valueOne'], [{two: 'two'}, 'valueTwo']]), [{twenty: 'twenty'}, {one: 'one'}]);
+   *     assert.doesNotHaveAllDeepKeys(new Set([{one: 'one'}, {two: 'two'}]), {twenty: 'twenty'});
+   *     assert.doesNotHaveAllDeepKeys(new Set([{one: 'one'}, {two: 'two'}]), [{one: 'one'}, {fifty: 'fifty'}]);
+   *
+   * @name doesNotHaveAllDeepKeys
+   * @param {Mixed} object
+   * @param {...String|Array|Object} keys
```
meeber: It doesn't look to me like the `...String` style of the `keys` argument is supported here.

lib/chai/interface/assert.js
```diff
@@ -1544,6 +1544,158 @@ module.exports = function (chai, util) {
...
+   * @param {String} message
+   * @namespace Assert
+   * @api public
+   */
+
+  assert.hasAllDeepKeys = function (obj, keys, msg) {
+     if (keys === undefined) {
+      new Assertion(obj, msg).to.have.all.deep.keys();
+    }
```
meeber: What if the `undefined` check was at the top of the actual `assertKeys` function instead?

lib/chai/interface/assert.js
```diff
@@ -1544,6 +1544,158 @@ module.exports = function (chai, util) {
...
+   * @param {...String|Array|Object} keys
+   * @param {String} message
+   * @namespace Assert
+   * @api public
+   */
+
+  assert.containsAllDeepKeys = function (obj, keys, msg) {
+     if (keys === undefined) {
+      new Assertion(obj, msg).to.contain.all.deep.keys();
+    }
```
meeber: What if the `undefined` check was at the top of the actual `assertKeys` function instead?

lib/chai/interface/assert.js
```diff
@@ -1544,6 +1544,158 @@ module.exports = function (chai, util) {
...
+   * @param {...String|Array|Object} keys
+   * @param {String} message
+   * @namespace Assert
+   * @api public
+   */
+
+  assert.doesNotHaveAnyDeepKeys = function (obj, keys, msg) {
+     if (keys === undefined) {
+      new Assertion(obj, msg).to.not.have.any.deep.keys();
+    }
```
meeber: What if the `undefined` check was at the top of the actual `assertKeys` function instead?

lib/chai/interface/assert.js
```diff
@@ -1544,6 +1544,158 @@ module.exports = function (chai, util) {
...
+   * @param {...String|Array|Object} keys
+   * @param {String} message
+   * @namespace Assert
+   * @api public
+   */
+
+  assert.doesNotHaveAllDeepKeys = function (obj, keys, msg) {
+     if (keys === undefined) {
+      new Assertion(obj, msg).to.not.have.all.deep.keys();
+    }
```
meeber: What if the `undefined` check was at the top of the actual `assertKeys` function instead?

test/assert.js
```diff
@@ -747,6 +747,22 @@ describe('assert', function () {
       assert.doesNotHaveAnyKeys(testMap, [ 'thisDoesNotExist', 'thisToo', {iDoNot: 'exist'} ]);
       assert.doesNotHaveAllKeys(testMap, [ aKey, {iDoNot: 'exist'} ]);
 
+      // Tests for the deep variations of the keys assertion
```
meeber: I'm thinking a test should be added to the non-deep variety of assertion above demonstrating that it uses strict equality instead of deep. For example:
```js
assert.doesNotHaveAnyKeys(testMap, {thisIs: 'anExampleObject'});
```

test/assert.js
```diff
@@ -877,6 +903,22 @@ describe('assert', function () {
       errSet.add({1: 20});
       errSet.add('number');
 
+      // Tests for the deep variations of the keys assertion
```
meeber: I'm thinking a test should be added to the non-deep variety of assertion above demonstrating that it uses strict equality instead of deep. For example:

```js
assert.doesNotHaveAnyKeys(testSet, {thisIs: 'anExampleObject'});
```

test/expect.js
```diff
@@ -1205,6 +1205,27 @@ describe('expect', function () {
       expect(testMap).to.not.have.any.keys([20, 1, {13: 37}]);
       expect(testMap).to.not.have.all.keys([aKey, {'iDoNot': 'exist'}]);
 
+      // Using the same assertions as above but with `.deep` flag instead of using referential equality
```
meeber: Same thing as mentioned in the assert interface tests about adding a non-deep test to demonstrate strict equality.

test/expect.js
```diff
@@ -1285,6 +1316,26 @@ describe('expect', function () {
       expect(testSet).to.not.have.any.keys([20, 1, {13: 37}]);
       expect(testSet).to.not.have.all.keys([aKey, {'iDoNot': 'exist'}]);
 
+      // Using the same assertions as above but with `.deep` flag instead of using referential equality
```
meeber: Same thing as mentioned in the assert interface tests about adding a non-deep test to demonstrate strict equality.

test/should.js
```diff
@@ -1051,6 +1051,23 @@ describe('should', function() {
       testMap.should.not.have.any.keys([20, 1, {13: 37}]);
       testMap.should.not.have.all.keys([aKey, {'iDoNot': 'exist'}]);
 
+      // Using the same assertions as above but with `.deep` flag instead of using referential equality
```
meeber: Same thing as mentioned in the assert interface tests about adding a non-deep test to demonstrate strict equality.

test/should.js
```diff
@@ -1131,6 +1158,26 @@ describe('should', function() {
       testSet.should.not.have.any.keys([20, 1, {13: 37}]);
       testSet.should.not.have.all.keys([aKey, {'iDoNot': 'exist'}]);
 
+      // Using the same assertions as above but with `.deep` flag instead of using referential equality
```
meeber: Same thing as mentioned in the assert interface tests about adding a non-deep test to demonstrate strict equality.

lib/chai/core/assertions.js
lucasfcosta: You are right!
I'll create a different variable to store that text.

lucasfcosta: You are right!
I'm going to use a different variable to store that String, thanks 😄

lib/chai/interface/assert.js
lucasfcosta: That makes sense to me, since we're doing that for every `key` related assertion I see no problem in doing that inside `assertKeys`, I think it would be more accurate and consistent.
Good idea!

lucasfcosta: Hmm, I just thought about it and it seemed a good idea at first, but I'm not sure we should have this behavior at all, especially because of [this line](https://github.com/chaijs/chai/pull/668/files/e8bde1c45a4e744b8844804710317c8fb9bc6b95#diff-6f8be24f0197a0249fde03c55e7397dbR1399). The assertion itself expects the user to pass some keys, IMO it would not be nice to override this behavior by adding a previous check inside the assert interface and thus making this interface have a behavior different than the others.
However, I feel like these assertions should be valid:

```js
var obj = {};
var objWithKeys = {key: 'val'};
expect(obj).to.not.have.any.keys();
expect(objWithKeys).to.have.any.keys();
```
What do you guys think?
@keithamus @meeber @vieiralucas 

For now I have removed these checks on the assert interface, just for the sake of consistency.

meeber: I think I see why those checks were there in the first place. Without them, if someone called one of these assertions using the assert interface, and didn't provide the `keys` argument, then `undefined` would get explicitly passed in and treated as one of the keys due to the use of `arguments` [here](https://github.com/chaijs/chai/blob/01963b74389f35a6f13739383015ac945987485c/lib/chai/core/assertions.js#L1365) and [here](https://github.com/chaijs/chai/blob/01963b74389f35a6f13739383015ac945987485c/lib/chai/core/assertions.js#L1380).

Also the fact that someone might actually explicitly pass in `undefined` when testing the keys of a map or set because `undefined` is a valid key further muddies the waters.  Therefore, a distinction has to be made either way between no keys being provided and `undefined` explicitly being passed in. And the assert interface needs to make sure it honors whatever the user did in terms of not providing any keys or explicitly passing in `undefined`.

This is pretty messy stuff and a bunch of new tests would be needed to ensure everything is happening correctly with both explicit and implicit `undefined` across the interfaces.

I agree that implicit `undefined` (aka, no `keys` argument specified) should be valid in the way you exemplified above, even though the `empty` assertion does the same thing.

I wouldn't be opposed to ignoring this issue in this PR and tabling it for later.

lucasfcosta: @meeber agreed!

lucasfcosta:
Here it is!
So, **with the removal of the `if` conditions from the `assert` interface this becomes a breaking change**.

I'd also like to add a note here for my future self: it would be nice to refactor these methods in order for them to support what I've said in [this comment](https://github.com/chaijs/chai/pull/668#discussion_r79943456) and also throw an error indicating keys are required when no keys are passed for the other combination of keys assertions other than the ones I mentioned there.

meeber:
@lucasfcosta I like the extra tests you added since the last commit on the `assert` interface that demonstrate that strict equality is used by default, but I think similar tests should be added on the `expect` and `should` interfaces, just for consistency.

lucasfcosta:
@meeber, seems adequate!
Good idea :smile: 
I'm gonna do it later today, thanks for noticing.

lucasfcosta:
@meeber here it is!
I think we've got everything covered by now.
I added tests to see if every combination of `have`, `contain`, `all` and `any` is using `strict` equality.

meeber:
@lucasfcosta Well done sir! LGTM! The only way this could be better is if we add another 5 or 6 interfaces so you can add all of the same tests on those interfaces too! Wouldn't that be fun?! :D

Pinging @keithamus and @vieiralucas, but I recommend that we merge #808 first and verify that it switches the build to passing before we merge this or any other PR.

vieiralucas:
As I have said previously, this is awesome work.

LGTM

Issue #669, opened 2016-04-07
[ES6] should.equal fails for Symbols


meeber:
Hallo :D

I ran into this small issue while testing Symbol equality using the `should` syntax:

```js
let prince = Symbol("Love");

expect(prince).to.equal(prince);  // Passes
prince.should.equal(prince);  // AssertionError: expected {} to equal {}
```

Adding ` || this instanceof Symbol` to the shouldGetter resolved the issue. I'm not sure if a PR is desired since it's ES6-specific?

Thanks!

keithamus:
Thanks for the issue @meeber!

Where do you propose this change should go? Which line? We should make sure everything `expect` supports `should` also supports, so this is a valid bug.

meeber:
Looks like this can be resolved in...

https://github.com/chaijs/chai/blob/816d526a0ea460271c3725f79f4f4c178b0a4a8d/lib/chai/interface/should.js#L12-L17

...by changing...

```js
    function shouldGetter() {
      if (this instanceof String || this instanceof Number || this instanceof Boolean ) {
        return new Assertion(this.valueOf(), null, shouldGetter);
      }
      return new Assertion(this, null, shouldGetter);
    }
```

...to...

```js
    function shouldGetter() {
      if (this instanceof String || this instanceof Number || this instanceof Boolean || this instanceof Symbol) {
        return new Assertion(this.valueOf(), null, shouldGetter);
      }
      return new Assertion(this, null, shouldGetter);
    }
```

Although that would throw a ReferenceError in pre-ES6 Node/browsers without Symbol support. And it's a really long line. So maybe:

```js
    function shouldGetter() {
      if (this instanceof String
          || this instanceof Number
          || this instanceof Boolean
          || 'function' === typeof Symbol && this instanceof Symbol) {
        return new Assertion(this.valueOf(), null, shouldGetter);
      }
      return new Assertion(this, null, shouldGetter);
    }
```

keithamus:
I wonder if we should just ducktype that line instead?

```js
function shouldGetter() {
  if (typeof this.valueOf === 'function') {
    return new Assertion(this.valueOf(), null, shouldGetter);
  }
  return new Assertion(this, null, shouldGetter);
}
```

Might be an interesting solution to explore.

meeber:
Love that approach. `Object.prototype.valueOf` will just end up returning `this` for most non-primitives anyway unless the user shadowed `valueOf` or used `Object.create(null)`.

[...]

PR #672, opened 2016-04-08
Add ES6 Symbol support

meeber:
Changes:

- Fix bug when testing Symbol equality with should syntax (#669)
- Add Symbol tests for assert, expect, and should

Notes:

- No symbol tests were added to the key-checking block because of a separate bug. Will open a new issue on it soon. The bug occurs because both the `expected` and `actual` array of keys were being sorted in the `assertKeys` function using `Array.prototype.sort` but this throws a TypeError when one of the keys is a Symbol in a Map or a Set.

- No Symbol tests were added to the `NaN` block because JavaScript's isNaN() throws a TypeError when passed a Symbol. I don't know if Chai is supposed to mimic this behavior with its NaN test or instead return true?

- No Symbol tests that intentionally cause an error (and then test the string) were added. This is because the currently published version of Mocha doesn't correctly display AssertionError messages containing Symbols. However this issue appears to be resolved in Mocha's master branch; at least it displays "{}" to represent a Symbol.

- I was unable to use the more succinct `typeof this === 'symbol'` approach mentioned in #669 because `this` is wrapped in an object and thus returns 'object' as its typeof.

### DIFF
--- a/lib/chai/interface/should.js
+++ b/lib/chai/interface/should.js
@@ -10,7 +10,10 @@ module.exports = function (chai, util) {
   function loadShould () {
     // explicitly define this method as function as to have it's name to include as `ssfi`
     function shouldGetter() {
-      if (this instanceof String || this instanceof Number || this instanceof Boolean ) {
+      if (this instanceof String
+          || this instanceof Number
+          || this instanceof Boolean
+          || typeof Symbol === 'function' && this instanceof Symbol) {
         return new Assertion(this.valueOf(), null, shouldGetter);
       }
       return new Assertion(this, null, shouldGetter);
--- a/test/assert.js
+++ b/test/assert.js
@@ -107,12 +107,21 @@ describe('assert', function () {
   it('equal', function () {
     var foo;
     assert.equal(foo, undefined);
+
+    if (typeof Symbol === 'function') {
+      var sym = Symbol();
+      assert.equal(sym, sym);
+    }
   });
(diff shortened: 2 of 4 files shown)


meeber:
On second thought, this Map/Set fix might be problematic.

If a user already has a keys-related test written that uses Map or Set (without any Symbol keys), and if it relies on the parameters in the AssertionError message to be sorted via `Array.prototype.sort`, then I believe their test would stop passing.

It's gonna be a little awkward to respect the existing behavior of sorting the expect/actual in the AssertionError message for Map and Set key checks but not sort them if either contains keys that are Symbols.

keithamus:
Hey @meeber thanks for the PR!

In the nicest way possible; try not to do too much in one PR. If you dial this back to just the fix discussed in #669 and then we can work on how to address the keys issue separately - without holding up the fix to Symbol equality. Sound good?

meeber:
Agree 100%. Was actually doing that when you posted.

The changes have been made, only #669 fix and tests are in now.

keithamus:
In that case, fantastic work @meeber :smile:

keithamus:
LGTM. @lucasfcosta?

meeber:
Gotta run to work now so I'll open issues for the other stuff later. Thanks!

lucasfcosta:
LGTM.
Nice work and thanks for pointing out other issues.

Issue #674, opened 2016-04-08
assertKeys cannot handle some valid ES6 keys


meeber:
There seems to be two ES6 issues with assertKeys:

**Issue 1**

If `expected` is a regular object with a key that is a Symbol, then assertKeys will not behave as expected. For example, the following assertion fails:

```js
var key1 = Symbol()
  , key2 = 42
  , obj = {};
obj[key1] = 'val1';
obj[key2] = 'val2';
expect(obj).to.have.all.keys([key1, key2]);
// AssertionError: expected { '42': 'val2' } to have keys 'Symbol()', and '42'
```

val1 is missing from the AssertionError's `actual` because the behavior of `Object.keys` is to not  include Symbols in its result set. (https://github.com/chaijs/chai/blob/6b640f71451fa575f24b302e56b99379c86d8bd5/lib/chai/core/assertions.js#L1173)

val2 is showing up as the string "Symbol()" in the AssertionError's `expected` because it's explicitly converted to a string here: (https://github.com/chaijs/chai/blob/6b640f71451fa575f24b302e56b99379c86d8bd5/lib/chai/core/assertions.js#L1187)

**Issue 2**

A `TypeError` is thrown if any of the `expected` or `actual` keys in a keys assertion cannot be implicitly converted to a string by `Array.prototype.sort` here: https://github.com/chaijs/chai/blob/6b640f71451fa575f24b302e56b99379c86d8bd5/lib/chai/core/assertions.js#L1251-L1252

Examples that fall victim to this issue but not the previous issue:

```js
// key1 has no toString
var key1 = Object.create(null)
  , key2 = 42;
expect(new Map([[key1, 'val1'], [key2, 'val2']])).to.have.all.keys([key1, key2]);
// TypeError: Cannot convert object to primitive value
```
```js
// key1 has non-function toString
var key1 = {toString: null}
  , key2 = 42;
expect(new Map([[key1, 'val1'], [key2, 'val2']])).to.have.all.keys([key1, key2]);
// TypeError: Cannot convert object to primitive value
```
```js
// key1 has Symbol's special toString which rejects implicit conversion
var key1 = Symbol()
  , key2 = 42;
expect(new Map([[key1, 'val1'], [key2, 'val2']])).to.have.all.keys([key1, key2]);
// TypeError: Cannot convert a Symbol value to a string
```

**Possible Resolutions**

[...]

PR #676, opened 2016-04-09
Fix bug when asserting some valid ES6 keys

meeber:
- Resolution of #674
- Add compareByInspect utility for use with assertKeys sorts
- Add getOwnEnumerableProperties utility
- Add getOwnEnumerablePropertySymbols utility
- Add Symbol support to the inspect utility
- Add tests to utilities, should, expect, and assert

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -1170,7 +1170,7 @@ module.exports = function (chai, _) {
       }
 
     } else {
-      actual = Object.keys(obj);
+      actual = _.getOwnEnumerableProperties(obj);
 
       switch (_.type(keys)) {
         case 'array':
@@ -1184,7 +1184,10 @@ module.exports = function (chai, _) {
           keys = Array.prototype.slice.call(arguments);
       }
 
-      keys = keys.map(String);
+      // Only stringify non-Symbols because Symbols would become "Symbol()"
+      keys = keys.map(function (val) {
+        return typeof val === 'symbol' ? val : String(val);
+      });
     }
 
     if (!keys.length) throw new Error('keys required');
@@ -1248,8 +1251,8 @@ module.exports = function (chai, _) {
         ok
       , 'expected #{this} to ' + str
       , 'expected #{this} to not ' + str
-      , expected.slice(0).sort()
-      , actual.sort()
+      , expected.slice(0).sort(_.compareByInspect)
+      , actual.sort(_.compareByInspect)
       , true
     );
   }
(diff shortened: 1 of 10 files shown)


lib/chai/core/assertions.js
```diff
@@ -1172,6 +1172,11 @@ module.exports = function (chai, _) {
     } else {
       actual = Object.keys(obj);
 
+      if (typeof Object.getOwnPropertySymbols === 'function') {
+        // Object.keys excludes Symbols so we need this too
+        actual = actual.concat(Object.getOwnPropertySymbols(obj));
```
keithamus: Slight concern with this - `Object.getOwnPropertySymbols` returns enumerable _and_ non-enumerable keys - somewhat changing the semantics of what this assertion already does. Should we filter out the non-enumerable keys? Something like:


```js
actual = actual.concat(Object.getOwnPropertySymbols(obj).filter(function (symbol) {
  return Object.getOwnPropertyDescriptor(obj, symbol).enumerable;
}));
````

**Comment by meeber** [comment]
Whoops! Great catch. I'm on it :)

File: lib/chai/core/assertions.js (hunk 1 above)
Comment (meeber) [inline lib/chai/core/assertions.js]: Whoops! Great catch. I'm on it :)

**Comment by meeber** [comment]
Fixed the issue with the non-enumerable symbols, and also modified a couple of the keys tests to validate that non-enumerable strings and symbols aren't included.

File: lib/chai/utils/getOwnEnumerablePropertySymbols.js [hunk 2]
```diff
@@ -0,0 +1,27 @@
...
+ * Copyright(c) 2011-2016 Jake Luer <jake@alogicalparadox.com>
+ * MIT Licensed
+ */
+
+/**
+ * ### .getOwnEnumerablePropertySymbols(object)
+ *
+ * This allows the retrieval of directly-owned enumerable property symbols of an
+ * object. This function is necessary because Object.getOwnPropertySymbols
+ * returns both enumerable and non-enumerable property symbols.
```
lucasfcosta: This description seems wrong since this function returns only enumerable symbol properties and not all of them (enumerable and non-enumerable)

lucasfcosta:
I just went through it all and it LGTM!
Awesome work, thanks for including complete tests for the features you've just implemented.
I really, really liked it.

meeber:
Great, thank you! Just noticed a typo in the comments; gonna fix it real quick.

keithamus:
LGTM 👍

Issue #648, opened 2016-03-27
is mocha test not being run anymore?


unional:
Notice this:
https://github.com/chaijs/chai/blob/master/test/bootstrap/index.js#L14
would fail as `AssertionError` does not take an option bag as first argument.

Issue #675, opened 2016-04-09
Helper function for testing AssertionError actual and expected


meeber:
Currently, there's the `global.err` helper function for testing that an AssertionError is thrown with a specific message: https://github.com/chaijs/chai/blob/4f7284662ced18700008ba491735ab4629e6f7bc/test/bootstrap/index.js#L11-L22

I see some value in creating a new helper function (or modifying the existing `global.err`) to facilitate the testing of an AssertionError's `expected` and `actual` properties. It seems important to validate these values since some assertions (such as `assertKeys`) go out of their way to sort `expected` and `actual` in a particular order.

Such validation could be done without a helper function but it'd be a lot of repeated code.

Thoughts? :D

keithamus:
Hey @meeber thanks for the issue.

Seems like you could pretty easily extend `global.err` to also feature `actual` and `expected`. Thoughts?

meeber:
It's nice that `global.err` expects only a string or a regex as it's second argument, allowing it to be extended to accept `undefined` as an indicator that the msg validation can be skipped, without breaking current functionality. And then it could accept optional arguments for testing `actual` and/or `expected`.

But if not passing a value for the `actual` and `expected` arguments means don't validate them against the `actual` and `expected` AssertionError properties, how would one go about validating if the `actual` and `expected` AssertionError properties were equal to `undefined`?

I suppose `arguments.length` could be checked, which should indicate if `undefined` was explicitly passed in as an argument.

Actually... How about this: `global.err` still only accepts 2 arguments, but if the second one is an object instead of a string/regex, then it validates each of the object's key/value pairs against the AssertionError's key/values?

keithamus:
Sounds completely acceptable to me. Look forward to seeing the PR 😄

PR #677, opened 2016-04-12
Refactor global.err to test error properties

meeber:
- Add feature https://github.com/chaijs/chai/issues/675
- Fix bug https://github.com/chaijs/chai/issues/648
- Allow object overload of global.err's second argument
- Combine global.err and window.err into one file and function
- Add tests for global.err


meeber:
One of the tests is passing in node.js but failing in karma due to an added "...".

```js
err(function () {
  err(function () { throw new Err('cat') }, {text: 'cat'});
}, 'expected { Object (message, showDiff) } to have a property \'text\'');
```
```
expected 'expected { Object (message, showDiff, ...) } to have a property \'text\'' to equal 'expected { Object (message, showDiff) } to have a property \'text\''
```

Haven't been able to investigate deeply into it yet. Played around briefly with `truncateThreshold` with no luck.

keithamus:
You could use a regexp to get around that 😄

meeber:
:D I switched that test to a regex and it passes now. Is the engine-specific message inconsistency something we should be concerned about though as a separate issue?

Couple notes about this PR:

- I didn't realize at first that the `err` function had two identical implementations in separate files, one for node and one for karma. I tried a few different approaches for combining them and this felt like the least-awkward of the bunch. Thoughts?

- I originally wanted to use `includes` for the object property testing, but the lack of deep equality was problematic. Some of the assertions create new objects as their `expected` and `actual`, so I think deep equality is the only way to go.

Ended up doing this:
```js
case 'object': return Object.keys(val).forEach(function (key) {
  chai.expect(err).to.have.property(key).and.to.deep.equal(val[key]);
});
```

But I wanted to do this:
```js
case 'object': return chai.expect(err).to.deep.include(val);
```

keithamus:
> Is the engine-specific message inconsistency something we should be concerned about though as a separate issue?

Speaking long term - lots of this plumbing will slowly be refactored out and so there is no need to be concerned about it now.

> I didn't realize at first that the err function had two identical implementations in separate files, one for node and one for karma. I tried a few different approaches for combining them and this felt like the least-awkward of the bunch. Thoughts?

What you've done here is great! Good work!

> I originally wanted to use includes for the object property testing, but the lack of deep equality was problematic. Some of the assertions create new objects as their expected and actual, so I think deep equality is the only way to go.

You can use `.deep.members` I believe. 

```js
case 'object': return chai.expect(err).to.deep.include.members(val);
```

Overall though, this is a great changeset! LGTM - I'd like @lucasfcosta to review also though 😄

test/bootstrap/index.js
```diff
@@ -1,22 +1,39 @@
...
 
-/*!
- * Provide check for fail function.
+/**
+ * Validate that the given function throws an error. Optionally validate some
+ * additional properties of the error:
+ *
+ * If val is a string, validate val equals the error's .message
+ * If val is a regex, validate val matches the error's .message
+ * If val is an object, validate val's props are included in the error object
```
lucasfcosta: Great behavior, nice idea

lucasfcosta:
Great work! :smile: 
You've been sending awesome PRs!

LGTM.

PR #680, opened 2016-04-17
Fix failure when custom inspect method is used

polomsky:
When custom inspect method is used and this method calls stylize test crashes.
Stylize method is documented for Node.js 4.x and 5.x.

Objects which use custom `inspect` method for `util.inspect` in Node.js should not cause test errors.


keithamus:
LGTM. @lucasfcosta?

lucasfcosta:
LGTM.

I wasn't fully aware of how Node's `inspect` worked. Just learned reading [this](https://nodejs.org/api/util.html#util_custom_inspect_function_on_objects).
Learning something new everyday.

Nice catch @polomsky!

PR #683, opened 2016-04-20
Throw using checkError module

lucasfcosta:
Hello everyone!
This one is pretty big and we've got plenty of special cases at the `throws` assertion, so be careful when reviewing :smile:

This one aims to solve an issue for the 4.0.0 release (as stated on our Roadmap #457) and finish #470 implementation. I thought that before moving `checkError` to a repository of its own it would be nice to have it reviewed and merged by more people here and only then we would move it to a new repo, but I'll do what you guys think it's better, this is just my opinion, please let me know if you disagree :wink: 

-----------------------------------------------------------

This PR includes:
- [x] The `checkError` module including *DocStrings*, `compatibleInstance`, `compatibleMessage`, `compatibleConstructor`, `getConstructorName` and `getMessage`
- [x] Tests for each one of the `checkError` functions
- [x] A full BC refactor of the `throws` assertion in order for it to use `checkError`

I couldn't make the `throws` assertion more simple than this, because we've got lots of edge/special cases, so this is the most generic and clean code I could get to while maintaining BC. If any of you can think of a better way to implement this please share your idea, it will certainly be welcome :smiley: 

I look forward to hearing your feedback :+1: 

-------------------------------
**EDIT**: This will be added to the [check-error repo](https://github.com/chaijs/check-error) as requested by @keithamus.
I'll write the complete repo structure and submit a PR there as soon as it's done.

Please don't merge this since it requires changes to use the external module instead of the local one.

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -1273,6 +1273,7 @@ module.exports = function (chai, _) {
    *
    *     var err = new ReferenceError('This is a bad function.');
    *     var fn = function () { throw err; }
+   *     expect(fn).to.throw();
    *     expect(fn).to.throw(ReferenceError);
    *     expect(fn).to.throw(Error);
    *     expect(fn).to.throw(/bad function/);
@@ -1289,8 +1290,8 @@ module.exports = function (chai, _) {
    * @name throw
    * @alias throws
    * @alias Throw
-   * @param {ErrorConstructor} constructor
-   * @param {String|RegExp} expected error message
+   * @param {Error|ErrorConstructor} errorLike
+   * @param {String|RegExp} errMsgMatcher error message
    * @param {String} message _optional_
    * @see https://developer.mozilla.org/en/JavaScript/Reference/Global_Objects/Error#Error_types
    * @returns error for chaining (null if no error)
@@ -1298,121 +1299,127 @@ module.exports = function (chai, _) {
    * @api public
    */
 
-  function assertThrows (constructor, errMsg, msg) {
+  function assertThrows (errorLike, errMsgMatcher, msg) {
     if (msg) flag(this, 'message', msg);
     var obj = flag(this, 'object');
+    var negate = flag(this, 'negate') || false;
     new Assertion(obj, msg).is.a('function');
 
-    var thrown = false
-      , desiredError = null
-      , name = null
-      , thrownError = null;
-
-    if (arguments.length === 0) {
-      errMsg = null;
-      constructor = null;
-    } else if (constructor && (constructor instanceof RegExp || 'string' === typeof constructor)) {
-      errMsg = constructor;
-      constructor = null;
-    } else if (constructor && constructor instanceof Error) {
-      desiredError = constructor;
-      constructor = null;
-      errMsg = null;
-    } else if (typeof constructor === 'function') {
-      name = constructor.prototype.name;
-      if (!name || (name === 'Error' && constructor !== Error)) {
-        name = constructor.name || (new constructor()).name;
-      }
-    } else {
-      constructor = null;
+    if (errorLike instanceof RegExp || typeof errorLike === 'string') {
+      errMsgMatcher = errorLike;
+      errorLike = null;
     }
 
+    var caughtErr;
     try {
       obj();
     } catch (err) {
-      // first, check desired error
-      if (desiredError) {
-        this.assert(
-            err === desiredError
-          , 'expected #{this} to throw #{exp} but #{act} was thrown'
-          , 'expected #{this} to not throw #{exp}'
-          , (desiredError instanceof Error ? desiredError.toString() : desiredError)
-          , (err instanceof Error ? err.toString() : err)
-        );
+      caughtErr = err;
+    }
 
-        flag(this, 'object', err);
-        return this;
+    // If we have the negate flag enabled and at least one valid argument it means we do expect an error
+    // but we want it to match a given set of criteria
+    var everyArgIsUndefined = errorLike === undefined && errMsgMatcher === undefined;
+
+    // If we've got the negate flag enabled and both args, we should only fail if both aren't compatible
+    // See Issue #551 and PR #683@GitHub
+    var everyArgIsDefined = Boolean(errorLike && errMsgMatcher);
+    var errorLikeFail = false;
+    var errMsgMatcherFail = false;
+
+    // Checking if error was thrown
+    if (everyArgIsUndefined || !everyArgIsUndefined && !negate) {
+      // We need this to display results correctly according to their types
+      var errorLikeString = 'an error';
+      if (errorLike instanceof Error) {
+        errorLikeString = '#{exp}';
+      } else if (errorLike) {
+        errorLikeString = errorLike.name;
       }
 
-      // next, check constructor
-      if (constructor) {
-        this.assert(
-            err instanceof constructor
-          , 'expected #{this} to throw #{exp} but #{act} was thrown'
-          , 'expected #{this} to not throw #{exp} but #{act} was thrown'
-          , name
-          , (err instanceof Error ? err.toString() : err)
-        );
+      this.assert(
+          caughtErr
+        , 'expected #{this} to throw ' + errorLikeString
+        , 'expected #{this} to not throw an error but #{act} was thrown'
+        , errorLike && errorLike.toString()
+        , (caughtErr instanceof Error ?
+            caughtErr.toString() : (typeof caughtErr === 'string' ? caughtErr : caughtErr && caughtErr.name))
+      );
+    }
 
-        if (!errMsg) {
-          flag(this, 'object', err);
-          return this;
+    if (errorLike && caughtErr) {
+      // We should compare instances only if `errorLike` is an instance of `Error`
+      if (errorLike instanceof Error) {
+        var isCompatibleInstance = _.checkError.compatibleInstance(caughtErr, errorLike);
+
+        if (isCompatibleInstance === negate) {
+          // These checks were created to ensure we won't fail too soon when we've got both args and a negate
+          // See Issue #551 and PR #683@GitHub
+          if (everyArgIsDefined && negate) {
+            errorLikeFail = true;
+          } else {
+            this.assert(
+                negate
+              , 'expected #{this} to throw #{exp} but #{act} was thrown'
+              , 'expected #{this} to not throw #{exp}' + (caughtErr && !negate ? ' but #{act} was thrown' : '')
+              , errorLike.toString()
+              , caughtErr.toString()
+            );
+          }
         }
       }
 
-      // next, check message
-      var message = 'error' === _.type(err) && "message" in err
-        ? err.message
-        : '' + err;
-
-      if ((message != null) && errMsg && errMsg instanceof RegExp) {
-        this.assert(
-            errMsg.exec(message)
-          , 'expected #{this} to throw error matching #{exp} but got #{act}'
-          , 'expected #{this} to throw error not matching #{exp}'
-          , errMsg
-          , message
-        );
-
-        flag(this, 'object', err);
-        return this;
-      } else if ((message != null) && errMsg && 'string' === typeof errMsg) {
-        this.assert(
-            ~message.indexOf(errMsg)
-          , 'expected #{this} to throw error including #{exp} but got #{act}'
-          , 'expected #{this} to throw error not including #{act}'
-          , errMsg
-          , message
-        );
-
-        flag(this, 'object', err);
-        return this;
-      } else {
-        thrown = true;
-        thrownError = err;
+      var isCompatibleConstructor = _.checkError.compatibleConstructor(caughtErr, errorLike);
+      if (isCompatibleConstructor === negate) {
+        if (everyArgIsDefined && negate) {
+            errorLikeFail = true;
+        } else {
+          this.assert(
+              negate
+            , 'expected #{this} to throw #{exp} but #{act} was thrown'
+            , 'expected #{this} to not throw #{exp}' + (caughtErr ? ' but #{act} was thrown' : '')
+            , (errorLike instanceof Error ? errorLike.toString() : errorLike && _.checkError.getConstructorName(errorLike))
+            , (caughtErr instanceof Error ? caughtErr.toString() : caughtErr && _.checkError.getConstructorName(caughtErr))
+          );
+        }
       }
     }
 
-    var actuallyGot = ''
-      , expectedThrown = name !== null
-        ? name
-        : desiredError
-          ? '#{exp}' //_.inspect(desiredError)
-          : 'an error';
+    if (errMsgMatcher) {
+      // Here we check compatible messages
+      var placeholder = 'including';
+      if (errMsgMatcher instanceof RegExp) {
+        placeholder = 'matching'
+      }
 
-    if (thrown) {
-      actuallyGot = ' but #{act} was thrown'
+      var isCompatibleMessage = _.checkError.compatibleMessage(caughtErr, errMsgMatcher);
+      if (isCompatibleMessage === negate) {
+        if (everyArgIsDefined && negate) {
+            errMsgMatcherFail = true;
+        } else {
+          this.assert(
+            negate
+            , 'expected #{this} to throw error ' + placeholder + ' #{exp} but got #{act}'
+            , 'expected #{this} to throw error not ' + placeholder + ' #{exp}'
+            ,  errMsgMatcher
+            ,  _.checkError.getMessage(caughtErr)
+          );
+        }
+      }
     }
 
-    this.assert(
-        thrown === true
-      , 'expected #{this} to throw ' + expectedThrown + actuallyGot
-      , 'expected #{this} to not throw ' + expectedThrown + actuallyGot
-      , (desiredError instanceof Error ? desiredError.toString() : desiredError)
-      , (thrownError instanceof Error ? thrownError.toString() : thrownError)
-    );
+    // If both assertions failed and both should've matched we throw an error
+    if (errorLikeFail && errMsgMatcherFail) {
+      this.assert(
+        negate
+        , 'expected #{this} to throw #{exp} but #{act} was thrown'
+        , 'expected #{this} to not throw #{exp}' + (caughtErr ? ' but #{act} was thrown' : '')
+        , (errorLike instanceof Error ? errorLike.toString() : errorLike && _.checkError.getConstructorName(errorLike))
+        , (caughtErr instanceof Error ? caughtErr.toString() : caughtErr && _.checkError.getConstructorName(caughtErr))
+      );
+    }
 
-    flag(this, 'object', thrownError);
+    flag(this, 'object', caughtErr);
   };
 
   Assertion.addMethod('throw', assertThrows);
--- a/test/utilities.js
+++ b/test/utilities.js
@@ -109,7 +109,7 @@ describe('utilities', function () {
       info.value.should.equal(obj.dimensions.units);
       info.name.should.equal('units');
       info.exists.should.be.true;
-    }); 
+    });
 
     it('should handle non-existent property', function() {
       var info = gpi('dimensions.size', obj);
@@ -118,7 +118,7 @@ describe('utilities', function () {
       expect(info.value).to.be.undefined;
       info.name.should.equal('size');
       info.exists.should.be.false;
-    }); 
+    });
 
     it('should handle array index', function() {
       var info = gpi('primes[2]', obj);
@@ -127,7 +127,7 @@ describe('utilities', function () {
       info.value.should.equal(obj.primes[2]);
       info.name.should.equal(2);
       info.exists.should.be.true;
-    }); 
+    });
 
     it('should handle dimensional array', function() {
       var info = gpi('dimensions.lengths[2][1]', obj);
@@ -136,7 +136,7 @@ describe('utilities', function () {
       info.value.should.equal(obj.dimensions.lengths[2][1]);
       info.name.should.equal(1);
       info.exists.should.be.true;
-    }); 
+    });
 
     it('should handle out of bounds array index', function() {
       var info = gpi('dimensions.lengths[3]', obj);
@@ -180,13 +180,13 @@ describe('utilities', function () {
       hp(1, arr).should.be.true;
       hp(3, arr).should.be.false;
     });
-    
+
     it('should handle literal types', function() {
       var s = 'string literal';
       hp('length', s).should.be.true;
       hp(3, s).should.be.true;
       hp(14, s).should.be.false;
-      
+
       hp('foo', 1).should.be.false;
     });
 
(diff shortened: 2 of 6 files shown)


keithamus:
Hey @lucasfcosta this is looking great. I've created a [check-error repo](https://github.com/chaijs/check-error), if you could please make a PR putting the util js in there, that'd be aaaweeesome. Then we can rework this PR to depend on that lib 😄

lucasfcosta:
Ah, of course @keithamus!
As you wish buddy, I'll do it, np :smile:

meeber:
Wow! This was quite an undertaking. Great work :D

I have a few questions which I'll add line comments for in a moment.

I also noticed that the issue mentioned in https://github.com/chaijs/chai/pull/551 still persists:

```js
expect(function() { throw new Error(); }).to.throw(Error, 'hello'); // fails
expect(function() { throw new Error(); }).to.not.throw(Error, 'hello'); // fails too
```

Is this the desired behavior?

lib/chai/core/assertions.js
```diff
@@ -1285,130 +1286,104 @@ module.exports = function (chai, _) {
...
+   * @param {String|RegExp} errMatcher error message
    * @param {String} message _optional_
    * @see https://developer.mozilla.org/en/JavaScript/Reference/Global_Objects/Error#Error_types
    * @returns error for chaining (null if no error)
    * @namespace BDD
    * @api public
    */
 
-  function assertThrows (constructor, errMsg, msg) {
+  function assertThrows (errorLike, errMatcher, msg) {
```
meeber: I'm not crazy about the terms `errorLike` and `errMatcher`, partially because one is "error" and the other "err", and partially because I feel some context is lost with "errMatcher".

What do you think about `errLike` and `errMsgMatcher`?

lib/chai/core/assertions.js
```diff
@@ -1285,130 +1286,104 @@ module.exports = function (chai, _) {
...
-      name = constructor.prototype.name;
-      if (!name || (name === 'Error' && constructor !== Error)) {
-        name = constructor.name || (new constructor()).name;
-      }
-    } else {
-      constructor = null;
+    if (errorLike instanceof RegExp || typeof errorLike === 'string' && arguments.length < 2) {
+      errMatcher = errorLike;
+      errorLike = null;
     }
```
meeber: It's my understanding that this logic allows a user to overload the first parameter when they only care about error message and not type, allowing them to do `assertThrows("some message")` instead of `assertThrows(undefined, "some message")`.

But how does the `arguments.length < 2` play into the mix? And why is it only important if the first argument is a string, but not if the first argument is a regexp?

Just for giggles I tried removing it and running tests and nothing broke, so if there is an importance to it I recommend a comment be added to the code, and a test be added to verify functionality.

lib/chai/core/assertions.js
```diff
@@ -1285,130 +1286,104 @@ module.exports = function (chai, _) {
...
-        );
+      caughtErr = err;
+    }
 
-        flag(this, 'object', err);
-        return this;
-      }
+    // If we have the negate flag enabled and at least one valid argument it means we do expect an error
+    // but we want it to match a given set of criteria
+    var everyArgIsUndefined = Array.prototype.slice.call(arguments).every(function(el) { return el === undefined });
```
meeber: Would it be more appropriate for this check to only consider the two first arguments,  `errorLike` and `errMatcher`? Returning `false` seems unexpected if the first two are `undefined` but the third argument `msg` is defined:

```js
expect(fn).to.throw(undefined, undefined, "ahhhhh")
```

test/utilities.js
```diff
@@ -841,4 +841,111 @@ describe('utilities', function () {
...
+      expect(checkError.compatibleInstance(errorInstance, sameInstance)).to.be.true;
+      expect(checkError.compatibleInstance(errorInstance, otherInstance)).to.be.false;
+      expect(checkError.compatibleInstance(errorInstance, Error)).to.be.true;
+      expect(checkError.compatibleInstance(errorInstance, aNumber)).to.be.true;
+    });
+
+    it('compatibleConstructor', function () {
+      var errorInstance = new Error('I am an instance');
+      var sameInstance = errorInstance;
+      var otherInstance = new Error('I an another instance');
```
meeber: Small typos on 869 and 857 with "an another" :D

lib/chai/utils/checkError.js
```diff
@@ -0,0 +1,132 @@
...
+ * @namespace Utils
+ * @api public
+ */
+function compatibleInstance(thrown, errorLike) {
+  if (errorLike instanceof Error) {
+    return thrown === errorLike;
+  } else {
+    return true;
+  }
+}
```
meeber: It's not obvious to me why `compatibleInstance` "Returns true if errorLike is not an instance of Error." Recommend a code comment explaining why.

lib/chai/core/assertions.js
lucasfcosta: Seems awesome!
It really looks semantically better, thanks for the insight.

lib/chai/core/assertions.js
lucasfcosta: `arguments.length < 2` is really not needed. You're right here.
I think I've got confused because it's possible to throw a `String` instead of an `Error` object, but then we should do message checks instead, since there's no way a thrown `String` will also have a message, it will be the message itself.

lib/chai/core/assertions.js
lucasfcosta: Whoa, great review buddy, I didn't see it coming.
Thanks :smile:

lib/chai/utils/checkError.js
lucasfcosta: That's because of `.throws` behavior. If this method didn't have this check we would need a check back there. Anyway it seems you're right, for the sake of modularity we should keep the check at `throws` and use only a strict comparison here.

I will also update this method's DocStrings, since it should only receive an `Error` instance to compare against and not an `ErrorConstructor`.

lucasfcosta:
Thanks for the review @meeber, great catches!
I'll do some fixes this afternoon.

I will also take a look at the behaviour you've just reported [here](https://github.com/chaijs/chai/pull/683#issuecomment-213110123), that shouldn't happen indeed, I'll make sure to add some tests to cover this.

lucasfcosta:
Hello everyone,
I've spent the whole afternoon thinking about [this](https://github.com/chaijs/chai/pull/683#issuecomment-213110123) and I came to the conclusion that there isn't a "clean" way to pull it off.

This is the code I currently have:

```js
  function assertThrows (errorLike, errMsgMatcher, msg) {
    var negate = flag(this, 'negate');

    if (errorLike instanceof RegExp || typeof errorLike === 'string') {
      errMsgMatcher = errorLike;
      errorLike = null;
    }

    var caughtErr;
    try {
      obj();
    } catch (err) {
      caughtErr = err;
    }

    // If we have the negate flag enabled and at least one valid argument it means we do expect an error
    // but we want it to match a given set of criteria
    var everyArgIsUndefined = errorLike === undefined && errMsgMatcher === undefined;

    // Checking if error was thrown
    if (everyArgIsUndefined || !everyArgIsUndefined && !negate) {
      // We need this to display results correctly according to their types
      var errorLikeString = 'an error';
      if (errorLike instanceof Error) {
        errorLikeString = '#{exp}';
      } else if (errorLike) {
        errorLikeString = errorLike.name;
      }

      this.assert(
          caughtErr
        , 'expected #{this} to throw ' + errorLikeString
        , 'expected #{this} to not throw an error but #{act} was thrown'
        , errorLike && errorLike.toString()
        , (caughtErr instanceof Error ?
            caughtErr.toString() : (typeof caughtErr === 'string' ? caughtErr : caughtErr && caughtErr.name))
      );
    }

    // If we've got the negate flag enabled and both args, we should only fail if both aren't compatible
    // See Issue #551@GitHub
    var everyArgIsDefined = errorLike && errMsgMatcher;
    var constructorError;
    var messageError;

    if (errorLike && caughtErr) {
      // We should compare instances only if `errorLike` is an instance of `Error`
      if (errorLike instanceof Error) {
        this.assert(
            _.checkError.compatibleInstance(caughtErr, errorLike)
          , 'expected #{this} to throw #{exp} but #{act} was thrown'
          , 'expected #{this} to not throw #{exp}' + (caughtErr && !negate ? ' but #{act} was thrown' : '')
          , errorLike.toString()
          , caughtErr.toString()
        );
      }

      try {
        this.assert(
            _.checkError.compatibleConstructor(caughtErr, errorLike)
          , 'expected #{this} to throw #{exp} but #{act} was thrown'
          , 'expected #{this} to not throw #{exp}' + (caughtErr ? ' but #{act} was thrown' : '')
          , (errorLike instanceof Error ? errorLike.toString() : errorLike && _.checkError.getConstructorName(errorLike))
          , (caughtErr instanceof Error ? caughtErr.toString() : caughtErr && _.checkError.getConstructorName(caughtErr))
        );
      } catch(e) {
        if (everyArgIsDefined && negate) {
          constructorError = e;
        } else {
          throw e;
        }
      }
    }

    if (errMsgMatcher) {
      // Here we check compatible messages
      var placeholder = 'including';
      if (errMsgMatcher instanceof RegExp) {
        placeholder = 'matching'
      }

      try {
        this.assert(
          _.checkError.compatibleMessage(caughtErr, errMsgMatcher)
          , 'expected #{this} to throw error ' + placeholder + ' #{exp} but got #{act}'
          , 'expected #{this} to throw error not ' + placeholder + ' #{exp}'
          ,  errMsgMatcher
          ,  _.checkError.getMessage(caughtErr)
        );
      } catch(e) {
        if (everyArgIsDefined && negate) {
          messageError = e;
        } else {
          throw e;
        }
      }
    }

    if (everyArgIsDefined && negate) {
      if (constructorError && messageError) {
        throw constructorError;
      }
    }

    flag(this, 'object', caughtErr);
  };
```

This code passes every test, including:

```js
expect(function() { throw new Error(); }).to.throw(Error, 'hello'); // this should fail
expect(function() { throw new Error(); }).to.not.throw(Error, 'hello'); // this should pass
```

**The problem is**: when running, for example, `expect(function() { throw new Error('hello'); }).to.not.throw(Error, 'hello')` the error we get back is only the error caused by the uncompatible constructors, and not both. To get a perfectly useful error message we would have to `throw` a `new AssertionError`, which I'm not sure it's a good idea.

Also, even if I do it, we will still have to change tests, since some of them use `not` combined with both `errorLike` and `errMsgMatcher` arguments, such as [this](https://github.com/chaijs/chai/blob/master/test/should.js#L55).

So, what do you guys think? Should I `throw` a `new AssertionError` with an useful message or should I leave this as it is without any changes to the tests other than the addition of what @meeber suggested above?

meeber:
What if we performed all of our validations near the top and stored the results in booleans?

For example, we already do this here:

```js
var caughtErr;
try {
  obj();
} catch (err) {
  caughtErr = err;
}
```

But what if we immediately went on to do this too, prior to any assertions:

```js
var hasCompatibleMessage = _.checkError.compatibleMessage(caughtErr, errMatcher)
    , hasCompatibleInstance = _.checkError.compatibleInstance(caughtErr, errorLike)
    , hasCompatibleConstructor = _.checkError.compatibleConstructor(caughtErr, errorLike);
```

Would that give us more freedom to organize the assertions and make sure the best error message is thrown on a failure (especially for compound tests)? It might also allow some of the `errorLike instanceof Error` type logic to move into the checkError utility functions.

lucasfcosta:
Thanks for the suggestion, @meeber, that looks great :smile: , I'll do it, but we still have a problem here.
What `Error` will we throw when the failure you've described above happens? Should we throw a `new AssertionError`?
That's what I'm still unsure about.

lucasfcosta:
Thanks for the suggestion, @meeber! :smile:
 
Unfortunately I'm not sure this would be a very good idea, because we would still need some `if`s to make the code work, because the `not` keyword inverts the `assertion` expression value and expects it to be true, so we would not be able eliminate `if` conditions and put them inside the assertion expressions.

Also, doing that prior to the assertions themselves would require us to add logic which is too tightly related to the `throw` assertion to the `check-error` module, and I think code should be as modular as possible, IMO we should think about adapting `throws` to `check-error` and not the other way round.

And after all, the main problem here still persists:
What `Error` will we throw when the failure you've described above happens? Should we throw a `new AssertionError` or only one of the previous errors?

I'll explore the code some more and see if I can find another solution for this, thanks again for sharing your thoughts, it's been great to have such a good reviewer :smile: 
That's what I'm still unsure about.

meeber:
You're right! We definitely don't want any throw-specific logic in the helper functions, and we'd end up doing many of the same `if` checks anyway. 

I think there may still be some value in lifting the `compatibleMessage`, `compatibleInstance`, and `compatibleConstructor` checks out of their `this.assert`s. Doing so might allow the logic to flow more cleanly and not require having to wrap assertions in `try` or throw a new assertion.

Before I ramble any further, lemme see if I can cook up an example to express my thoughts in code :D

meeber:
Okay, after playing around with the code, thinking about the problem, staring at my cat, and eating a chicken sammich, here's my analysis:

The reason this is so tricksy is because the negate logic is fundamentally different than the default logic.

With the default logic, you can proceed assertion-by-assertion and short-circuit the function as soon as an assertion fails. Conversely, with the negate logic, you must perform all relevant assertions ("relevant" being based on which combination of errLike and errMsgMatcher is provided); you can never short-circuit.

Because the default and negate logic are so different, you have three options:

1. Try to conform the negate logic into the default logic.
2. Try to conform the default logic into the negate logic.
3. Split default/negate into two separate logic paths, only sharing helper functions and the like.

There are issues with every option. Traditionally this function has gone with the first. I think that worked okay until trying to fix that one bug with the `.not.throws` behavior. It's my suspicion that the second approach could be significantly cleaner, avoiding the `try` blocks and having to rethrow errors. I haven't thought much about the third option, so I don't have much of an opinion of it yet, though you've kind of laid some groundwork for it with the checkError module.

Of course, refactoring to use the 2nd or 3rd option is made a bit trickier by having to respect the current failure message precedence for backward compatibility purposes, but only a minor hurdle I think.

lucasfcosta:
Thanks for the help @meeber!

The issue with the 3rd option is that it will create code repetition, because we've got to do some of the same things for both cases (`negate` and `!negate`). The second option may be viable, but I really can't guarantee that until I get to code it effectively, since the `throws` logic is a very intricate one.

I'm also not a big fan of having to rethrow errors as we're currently doing, so my first goal on the next refactor will be to find a clean way of doing that.

I will take some time this week for a new full refactor of `throws` now that I know more about its inner workings and I will take your considerations into account.

You've been very helpful, thanks! :smile: 

Oh, and just in case you want to stare at cats other than yours, [here you go](http://lucasfcosta.com/Oh-God-I-Love-Cats/), hahaha

meeber:
You've created a portal to eternal bliss. Well done :D 

I agree, the 3rd option likely isn't viable due to code repetition.

As for the 2nd option, it might not be as big of a refactor as I feared.

Basically any time we have something like this:

```js
      try {
        this.assert(
            _.checkError.compatibleConstructor(caughtErr, errorLike)
          , 'expected #{this} to throw #{exp} but #{act} was thrown'
          , 'expected #{this} to not throw #{exp}' + (caughtErr ? ' but #{act} was thrown' : '')
          , (errorLike instanceof Error ? errorLike.toString() : errorLike && _.checkError.getConstructorName(errorLike))
          , (caughtErr instanceof Error ? caughtErr.toString() : caughtErr && _.checkError.getConstructorName(caughtErr))
        );
      } catch(e) {
        if (everyArgIsDefined && negate) {
          constructorError = e;
        } else {
          throw e;
        }
      }
```

It could be refactored to something like this:

```js
        if (!_.checkError.compatibleConstructor(caughtErr, errorLike)) {
          this.assert(
            , false
            , 'expected #{this} to throw #{exp} but #{act} was thrown'
            , 'expected #{this} to not throw #{exp}' + (caughtErr ? ' but #{act} was thrown' : '')
            , (errorLike instanceof Error ? errorLike.toString() : errorLike && _.checkError.getConstructorName(errorLike))
            , (caughtErr instanceof Error ? caughtErr.toString() : caughtErr && _.checkError.getConstructorName(caughtErr))
          );
          perfectMatch = false;
        }
```

What this does is prevent the negate logic from short-circuiting if the constructor is compatible, giving us an opportunity to also check the compatibility of the message later on if we need to. But it still allows the non-negate logic to short-circuit if the constructor isn't compatible. 

Something similar would need to be done for compatibleInstance and compatibleMessage.

And then at the end, we could check if both `negate` and `perfectMatch` are set to true, and if they are then we `this.assert(true, blah, blah, blah, blah)`. And then the challenge becomes filling in the blahs with the most appropriate info ("expected {this} to not throw {errLike and/or errMsgMatcher}")

 Hope I explained that well, it's mind-twisting stuff :D

lucasfcosta:
Good idea, @meeber.
I've taken a look at it yesterday and I think I already know how to do this.
This is gonna be a busy week but I hope to refactor this taking your tips into account until the weekend.
Thanks for the help!

lucasfcosta:
It's finally done!

My brain was broken into so many pieces I can't even think straight anymore, that was an awesome puzzle!

I made some minor tweaks to @meeber's ideas and I could get it working. I had to create two new variables to indicate whether a check has failed or not and then I would mark them with `false` if both arguments were defined and we've got the `negate` turned on instead of failing an assertion.

This was the most elegant way I could get it to work, please feel free to suggest improvements if you find room for any.

I would also like to thank @meeber for the help with this PR, without you this wouldn't have been possible, you've done a great job buddy, thanks! :smile: 

------------------------

### Important:

Currently the fail message is the same as if the constructor check had failed, if you guys want this to be fully BC we can't change it, because maybe (it's a remote possibility, but we can't discard it) someone is expecting an specific error message (the constructor fail message), so changing it could break someone's tests.

What do you guys think about this?
Do you want me to submit another PR against `4.x.x` or will we merge this here and change error messages at the `4.x.x` branch?

meeber:
Well done, sir! LGTM. That was not an easy one. I concur with your approach to maintain backward compatibility for now; we can consider improving the error message for a future breaking change.

Thanks for all your hard work :D

lucasfcosta:
Thanks @meeber!
I wish I could put your name into those commits too, you were as responsible for this as I was!

Well, so @keithamus can merge this if he's happy with it :smile: 
Thanks folks!

keithamus:
@lucasfcosta @meeber is anything outstanding on https://github.com/chaijs/check-error/pull/1? We should merge that and release it as a module, so we can then rework this PR to use the external module 😄

lucasfcosta:
@keithamus [chaijs/check-error#1](https://github.com/chaijs/check-error/pull/1) is complete, it just has minor linting-related improvements for the linting task to pass without errors, but they're the same.

I will refactor this to use the external module as soon as it's published. :smile:

lucasfcosta:
The build is failing because the `check-error` module doesn't cover a special case regarding poorly-constructed errors. I have tested this with the changes I submitted at this [PR at check-error](https://github.com/chaijs/check-error/pull/5) and every test is now passing. As soon as it gets merged there we will be good to go. I will just need to update `package.json` with the new version number.

I have already:
- [x] Added `check-error` to the `package.json`
- [x] Used the external module instead of the local one
- [x] Removed tests for `check-error` module from this, since they're already being done externally at `check-error`

lucasfcosta:
I think this one is ready for a merge.
We're almost done here, thanks @meeber and @keithamus for you support 🎉🎉🎉🎉🎉🎉🎉

meeber:
Wooooooooooo! Do we wanna rebase into one commit before merge?

lucasfcosta:
@meeber Thanks for remembering, I got so excited I forgot to rebase 😆

meeber:
@lucasfcosta Hmm, still seems to be showing up as three commits.

lucasfcosta:
@meeber Ah, I thought you meant I should rebase on master, not meld/squash the commits into a single one.
Well, good idea anyway, I'm gonna do it as soon as I get home :smile:

lucasfcosta:
@meeber Done 🎉 🎉

meeber:
@lucasfcosta Amazing job! LGTM! Where are you going on vacation to celebrate??

lucasfcosta:
Thanks @meeber. There's two months left until I get my vacation, that's too much time to wait, let's throw a commit party on github 😄

Us celebrating

meeber:
@lucasfcosta I dunno that's more than a little terrifying :D

Someone should definitely double-check this but I think we'll be able to close the following once this gets merged: #470 #551 #596 #634 #635 

It doesn't seem to close #430 or #559, but I'm not sure they're using a valid form of the `.throw` assertion by doing `expect(fn).to.throw(new Error("blah"))`? If that's an invalid form, then we can suggest a different form, and close them.

It doesn't seem to close #355 either; stack trace still doesn't include original error frame.

lucasfcosta:
I think we can close all of those except #355.

Here are the reasons for the ones you were not sure:
* #430 because we didn't drop strict comparisons. The strict comparisons (`===`) will run whenever the user passes an error instance to the `.throws` assertion. That was a design decision.
* #436 because the initial example will now work correctly since when we've got 2 arguments (`Error` and `Message`) both will need to match in order for the assertion to pass. Using `new Error()` will trigger strict comparison instead
* #559 because as I said before,`new` triggers strict comparison and also because [the example here](https://github.com/chaijs/chai/issues/559#issuecomment-160193068) will now work, as you can see in [this test](https://github.com/lucasfcosta/chai/blob/b9436a9e564b7500be9f3bfafc44635db83d4bf6/test/expect.js#L1160) and [this test](https://github.com/lucasfcosta/chai/blob/b9436a9e564b7500be9f3bfafc44635db83d4bf6/test/should.js#L983) for example.

meeber:
@lucasfcosta Thanks for double checking! I agree with your assessment. You might be setting the world record for most issues and PRs closed by a single commit.

meeber:
@keithamus I forgot to ping you on this. Sorreh!

meeber:
pester pester @keithamus

meeber:
@lucasfcosta Oh hey I just thought of something. We should add check-error to README.md under "Related Projects".

Also!! I saw you ask a question on a youtube video!! :D

lucasfcosta:
@meeber Yes we should!
I'll open another PR for that since we've got other modules that are not listed into `README.md` so we can keep this one on scope.

PS.: Yes I did it! MPJ rocks, hahahaha, I love his channel 😄

keithamus:
Sorry for taking so long on this one. LGTM!

lucasfcosta:
@keithamus We all get plenty busy sometimes, don't worry about it Keith.
Thanks for merging. I also ticked the `check-error` checkbox at our [Roadmap](https://github.com/chaijs/chai/issues/457).

Thanks everyone for your support :tada: :tada: :tada: :tada:

meeber:
One last hurdle: Looks like a couple IE tests are failing.

lucasfcosta:
@meeber I'll check this out after lunch, that's strange since `check-error` itself works on every IE version.
Thanks for noticing!

lucasfcosta:
Just updating everyone:

Today I'm gonna take some time to:
* Fix tests on IE as @meeber pointed out
* Do some polishment on the `assert` interface as we talked about on #635

We will also need some more time to investigate and fix #355 in the future.
I expected these two things to be done until tomorrow.

keithamus:
#596 and #430 are also still open. Can I ask they also be checked on to see if they can be closed or at least updated with the impact of this merge.

Issue #688, opened 2016-04-24
Has own property on objects with nulll prototype


aryzing:
I recently ran a test that involved testing an object with a null prototype using `expect(obj).to.have.ownProperty('x')`. The test failed because it had no access to `Object.prototype.hasOwnProperty`. Can the `ownProperty` method be tweaked to use `call`/`apply` instead? Asi in `Object.prototype.hasOwnProperty.call(obj)`. 

Specifically, I was thinking about [line 1226](https://github.com/chaijs/chai/blob/master/chai.js#L1226)

meeber:
@aryzing Great catch! Sounds reasonable to me. It looks like your proposed solution is already used on [line 5082](https://github.com/chaijs/chai/blob/d664ef409ffdc1e0e70a27ee375aa780aa9f5c2e/chai.js#L5082) as well.

Are you interested in PRing it? This would be in `lib/chai/core/assertions.js`.

aryzing:
Sure! Thanks!

meeber:
Hmm... We should also prolly add a test that demonstrates current failure for the expect/should interfaces. Not seeing any equivalent functionality in the assert interface, though.

PR #689, opened 2016-04-25
Fix ownProperty on objects with no prototype

aryzing:
Solves #688 

Objects with no prototype can now be succesfully tested for containing properties.


meeber:
@aryzing Awesome, thanks for the PR! :D

LGTM. @lucasfcosta @keithamus ?

lucasfcosta:
Nice catch!
LGTM

PR #701, opened 2016-05-06
Add .does as a no-op assertion, Fix #700

### LINKED ISSUE #700: Requesting "does" as a no-op assertion
expect(foobar).to.have.property("baz").which.does.not.have.property("thing");

vieiralucas:
Hello guys, I saw that #700 was marked as pull-request-wanted.
So here I am 😄 

I also did not find tests addressing no-op assertions.
Maybe I could add tests for it.

What do you guys think about it?

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -34,6 +34,7 @@ module.exports = function (chai, _) {
    * - of
    * - same
    * - but
+   * - does
    *
    * @name language chains
    * @namespace BDD
@@ -43,7 +44,7 @@ module.exports = function (chai, _) {
   [ 'to', 'be', 'been'
   , 'is', 'and', 'has', 'have'
   , 'with', 'that', 'which', 'at'
-  , 'of', 'same', 'but' ].forEach(function (chain) {
+  , 'of', 'same', 'but', 'does' ].forEach(function (chain) {
     Assertion.addProperty(chain);
   });
 
--- a/test/expect.js
+++ b/test/expect.js
@@ -10,6 +10,30 @@ describe('expect', function () {
     expect('foo').to.equal('foo');
   });
 
+  it('no-op chains', function() {
+    function test(chain) {
+      // tests that chain exists
+      expect(expect(1)[chain]).not.undefined;
+
+      // tests methods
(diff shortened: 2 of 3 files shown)


lucasfcosta:
@vieiralucas is right at my side right now at college, I also watched him coding, so this LGTM.

I think it would be a great idea to have this kind of tests.
Putting them on the `expect` and `should` interfaces should be enough, since there's no way we can do this kind of chaining on the `assert` interface (and since it uses `should` behind the scenes).

meeber:
Hey guys I was thinking about the same thing earlier. I think a really simple test for each of the chain languages would definitely be good. And I agree it should appear in both the `should` and `expect` interface tests for thoroughness.

I didn't realize you two were acquaintances! This Lucas camaraderie is more dangerous than I feared... I need to find a second Meeber, and fast.

vieiralucas:
photo on 5-6-16 at 10 00 pm
Thinking about this PR

I'll do the tests, make sure to rebase this in order for the commits to make sense and keep the git logs clean.

vieiralucas:
I added the only test that I could think of, which just checks that the chain exists.
What else should be tested?

meeber:
I'd say validate that each of them is chaining correctly, even though grammatically it won't always make sense:

```js
    expect(3).to.equal(3);                                                      
    expect(3).and.equal(3);
    expect(3).of.equal(3);
    ...etc...
```

Also a nitpick: The indent-style in the test arrays is different than the `assertions.js` array.

vieiralucas:
I see, the problem is that the error will be
```javascript
TypeError: Cannot read property 'equal' of undefined
```

That's why I used `.not.undefined` instead of `.to.not.be.undefined`

But that's what you get when you use chai to test chai right?
Others tests already have this behavior.
I was wondering, have you consider using node's Assert to test chai?

Anyways, I'll add tests that validade if they are chaining correctly.

vieiralucas:
Looking closer I realized that it will be testing that `(3).equal(3)` still works.
Now I understand

meeber:
LGTM!

lucasfcosta:
Awesome, good job guys!

PR #511, opened 2015-09-01
Fix wrong order of expected and actual values for members matcher

eeroan:
This caused i.e. WebStorm's diff view to show actual value as expected and vice versa.

PR #702, opened 2016-05-10
.members should show diff

pgarrison:
It would be great to show what's missing; this works for me. Comments are welcome. 
I'm not sure why the previous code had `#{act}` instead of `#{exp}` or why it apparently swapped expected/actual.

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -1589,19 +1589,19 @@ module.exports = function (chai, _) {
     if (flag(this, 'contains')) {
       return this.assert(
           isSubsetOf(subset, obj, cmp)
-        , 'expected #{this} to be a superset of #{act}'
-        , 'expected #{this} to not be a superset of #{act}'
-        , obj
+        , 'expected #{this} to be a superset of #{exp}'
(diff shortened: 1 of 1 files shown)


meeber:
@pgarrison Thanks for the PR! 

The issue with `showDiff` not being `true` here is indeed an issue but has been fixed in the upcoming 4.x.x release per #574.

The issue with the transposed `actual` and `expected` arguments for the members assertion is also a known issue with a PR submitted to fix it per #511. However, the submitter of that PR hasn't responded in a month so I'd be happy to close that one and merge this one instead.

Would you be willing to update this PR to remove the `true` `showDiff` changes, since that conflicts with the existing fix in #574?

Thanks!

pgarrison:
Ah, didn't see the 4.x.x branch--should this be there instead?

meeber:
Nahhh, fixing the swapped `actual` and `expected` bug can go onto master.

Thank you for asking though!

pgarrison:
Alright. It no longer sets showDiff.

meeber:
@pgarrison Thank you, sir. LGTM.

keithamus:
LGTM!

PR #705, opened 2016-05-15
Fix failing tests in IE 10 and 11

meeber:
There are 4 failing tests in IE 11 and 1 failing test in IE 10. 

This PR in combination with #640 (which updates type-detect to v2.0.0) resolves all of them, hopefully making Travis CI happy so that our builds stop showing as "Failing".

Covered by this PR:

- IE 11's `Set` and `Map` don't support constructor arguments or `.set`/`.add` chaining. Therefore, I refactored tests to setup their `Set`s and `Map`s without using either unsupported feature.
- IE 11's `Set` and `Map` don't support the `.keys` method. Therefore, I made a small change to `assertKeys` to populate the `actual` array using `forEach` instead.
- IE 10 doesn't support `__proto__`, which means `addChainableMethod` uses different logic to do its thing. Unfortunately, this different logic isn't compatible with one of the `addMethod` assertions that validates the `.length` chainable method using `instanceof`. Therefore, I switched the assertion to perform the validation using `length.constructor`.

Covered by #640:

- IE 11's `Set` and `Map` aren't properly detected by `type-detect` v1.0.0. Both `assertKeys` and `globalErr` rely on proper type detection, so some tests will fail until type-detect is upgraded.

### DIFF
--- a/test/utilities.js
+++ b/test/utilities.js
@@ -264,12 +264,15 @@ describe('utilities', function () {
     // Checking if it's really an instance of an Assertion
     expect(assertion2).to.be.instanceOf(assertionConstructor);
 
-    // Test chaining `.length` after a method to guarantee it is not a function's `length`
+    // Test chaining `.length` after a method to guarantee it's not a function's
+    // `length`. Note: 'instanceof' cannot be used here because the test will
+    // fail in IE 10 due to how addChainableMethod works without __proto__
+    // support. Therefore, test the constructor property of length instead.
     var anAssertion = expect([1, 2, 3]).to.be.an.instanceof(Array);
-    expect(anAssertion.length).to.be.an.instanceof(assertionConstructor);
+    expect(anAssertion.length.constructor).to.equal(assertionConstructor);
 
     var anotherAssertion = expect([1, 2, 3]).to.have.a.lengthOf(3).and.to.be.ok;
-    expect(anotherAssertion.length).to.be.an.instanceof(assertionConstructor);
+    expect(anotherAssertion.length.constructor).to.equal(assertionConstructor);
   });
 
   it('overwriteMethod', function () {
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -1164,7 +1164,10 @@ module.exports = function (chai, _) {
       , mixedArgsMsg = 'when testing keys against an object or an array you must give a single Array|Object|String argument or multiple String arguments';
 
     if (_.type(obj) === 'map' || _.type(obj) === 'set') {
-      actual = Array.from(obj.keys());
+      actual = [];
+
+      // Map and Set '.keys' aren't supported in IE 11. Therefore, use .forEach.
+      obj.forEach(function (val, key) { actual.push(key) });
 
       if (_.type(keys) !== 'array') {
         keys = Array.prototype.slice.call(arguments);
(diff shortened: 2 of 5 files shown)


test/utilities.js
```diff
@@ -264,12 +264,15 @@ describe('utilities', function () {
...
     expect(assertion2).to.be.instanceOf(assertionConstructor);
 
-    // Test chaining `.length` after a method to guarantee it is not a function's `length`
+    // Test chaining `.length` after a method to guarantee it's not a function's
+    // `length`. Note: 'instanceof' cannot be used here because the test will
+    // fail in IE 10 due to how addChainableMethod works without __proto__
+    // support. Therefore, test the constructor property of length instead.
     var anAssertion = expect([1, 2, 3]).to.be.an.instanceof(Array);
-    expect(anAssertion.length).to.be.an.instanceof(assertionConstructor);
+    expect(anAssertion.length.constructor).to.equal(assertionConstructor);
```
lucasfcosta: Awesome job spotting this one. I think we should have this one merged for now but it would be good for us to fix this behavior regarding `__proto__` and the `instanceof`assertion in the future. What do you think about that?

lucasfcosta:
This one LGTM, awesome job!

Just out of curiosity: how did you manage to track what was failing into IE11?
I tried to read Travis' logs and there was no useful info there.
Unfortunately I can't install IE11 here (OSX) and run Karma tests without an herculean effort 😆

test/utilities.js
meeber: @lucasfcosta I'm actually not sure if the behavior is broken; I think it might've just been the test itself making an assumption that isn't true in legacy browsers.

The reason this assertion was failing in IE 10 is because without `__proto__` support, the `addChainableMethod` utility manually copies the context object's properties onto the `assert` object that's returned by the getter. [Reference](https://github.com/chaijs/chai/blob/6611e1a969a568b8e7996dd407ee8b80fa2f109f/lib/chai/utils/addChainableMethod.js#L95-L104)

This works, it's just that chainable methods aren't an `instanceof assertionConstructor` when using this method of copying properties as opposed to the `assert.__proto__ = Object.create(this);` method that's used when `__proto__` is available. Therefore, I think it's correct behavior that `instanceof assertionConstructor` returns false.

meeber:
@lucasfcosta Thanks! :D

After poking around with different approaches, I ended up using [Microsoft's free virtual machines](https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/windows/).

However, going forward, I think we need to find a way for the error messages in Sauce/Karma to provide stack traces.

Also, do you know why the Sauce/Karma tests aren't run when we submit PRs? I find it strange that those tests are run on PRs submitted by the greenkeeper bot. I've been looking at the Travis and Sauce config files. What am I missing? :(

keithamus:
This LGTM. Truly awesome work @meeber 😄.

> Also, do you know why the Sauce/Karma tests aren't run when we submit PRs? I find it strange that those tests are run on PRs submitted by the greenkeeper bot. I've been looking at the Travis and Sauce config files. What am I missing? :(

Annoyingly, the saucelabs access keys have to be encrypted and Travis doesn't use encrypted variables for pull requests - meaning SauceLabs cannot run. Sadly until saucelabs provides first class integration with travis, similar to what coveralls has done, then we won't get saucelabs builds on PRs.

PR #724, opened 2016-06-09
Fix chai.use as ES6 named export

### LINKED ISSUE #718: "use" doesn't behave as expected as an ES6 named export
The following code throws an error:

```js
import sinonChai from "sinon-chai";
import {expect, use} from "chai";

use(sinonChai);
```

The reason it throws an error is because the necessary `this` binding is lost when calling `use` via the above syntax as opposed to calling it via `chai.use(sinonChai)`.

I'd argue that it's unexpected and undesirable for a singleton module to utilize `this` in such a way.

One possible solution might be to change [use](https://github.com/chaijs/chai/blob/2194a89e0290a3152446995afe877d4248256c91/lib/chai.js#L38-L45) from:

```js
exports.use = function (fn) {
  if (!~used.indexOf(fn)) {
    fn(this, util);
    used.push(fn);
  }

  return this;
};
```

To:

```js
exports.use = function (fn) {
  if (!~used.indexOf(fn)) {
    fn(exports, util);
    used.push(fn);
  }

  return exports;
};
```

meeber:
- Fixes #718 
- Allows `use` to be imported in ES6 as follows:
```js
import {use} from "chai";
```

### DIFF
--- a/lib/chai.js
+++ b/lib/chai.js
@@ -37,11 +37,11 @@ var util = require('./chai/utils');
 
 exports.use = function (fn) {
   if (!~used.indexOf(fn)) {
-    fn(this, util);
+    fn(exports, util);
     used.push(fn);
   }
 
-  return this;
+  return exports;
 };
 
 /*!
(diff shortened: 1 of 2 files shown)


keithamus:
LGTM. Over to @lucasfcosta

lucasfcosta:
LGTM too.
Thanks for this @meeber.

PR #728, opened 2016-06-12
Add ordered flag for members assertions

### LINKED ISSUE #717: Add an assertion for same members where order is important
I've come to discover that it's not straightforward to test that an array has the same members, and that those members are in the expected order. Consider this base example:

```js
var a = {};
var b = {};
var list = [ a, b ];
```

All of the assertions around "members" all explicitly do not consider order, but it would be really convenient to be able to assert that too:

```js
// no idea what to call this assertion
assert.sameExactMembers(list, [ a, b ]);
```

My use-case is that I have an internal list that I can retrieve. In one case I return the list unsorted. (so `sameMembers` works great) In the other case, I am sorting it in a specific way, so I want to assert that my sort worked as expected.

I am using `deepEqual` now, but the member objects can be fairly big/deep and it's very inefficient to traverse these objects, especially since I have the direct object refs in my tests.

```js
// in my case, a & b could be huge, so would prefer not traverse them
assert.deepEqual(list, [ a, b ]);
```

Another option is that I can just use `strictEqual` on each item in the array, but that gets unwieldy quickly imo:

```js
// only 2 in this case... but it's copy-pasta and ugly beyond a few members
assert.strictEqual(list[0], a);
assert.strictEqual(list[1], b);
```

meeber:
- Adds `.ordered` functionality on every interface along with tests
- Adds missing `not` assertions for existing member assertions on `assert` interface
- Makes some existing member assertion tests more consistent across interfaces
- Updates documentation
- Fixes #717

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -93,6 +93,23 @@ module.exports = function (chai, _) {
   });
 
   /**
+   * ### .ordered
+   *
+   * Sets the `ordered` flag, later used by the `members`
+   * assertions.
+   *
+   *     expect(foo).to.have.ordered.members([bar, baz]);
+   *
+   * @name ordered
+   * @namespace BDD
+   * @api public
+   */
+
+  Assertion.addProperty('ordered', function () {
+    flag(this, 'ordered', true);
+  });
+
+  /**
    * ### .any
    *
    * Sets the `any` flag, (opposite of the `all` flag)
@@ -1547,8 +1564,9 @@ module.exports = function (chai, _) {
   Assertion.addMethod('closeTo', closeTo);
   Assertion.addMethod('approximately', closeTo);
 
-  function isSubsetOf(subset, superset, cmp) {
-    return subset.every(function(elem) {
+  function isSubsetOf(subset, superset, cmp, ordered) {
+    return subset.every(function(elem, idx) {
+      if (ordered) return cmp ? cmp(elem, superset[idx]) : elem === superset[idx];
       if (!cmp) return superset.indexOf(elem) !== -1;
 
       return superset.some(function(elem2) {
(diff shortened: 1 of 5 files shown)


lib/chai/core/assertions.js
```diff
@@ -1587,24 +1630,32 @@ module.exports = function (chai, _) {
...
 
     if (flag(this, 'contains')) {
-      return this.assert(
-          isSubsetOf(subset, obj, cmp)
-        , 'expected #{this} to be a superset of #{exp}'
-        , 'expected #{this} to not be a superset of #{exp}'
-        , subset
-        , obj
-      );
+      lengthCheck = true;
```
lucasfcosta: Nice work using `lengthCheck` at this assertion. The code looks more clean 👍

lucasfcosta:
Awesome work, as always. I went through it all and I couldn't even find a single typo.
Nice work reducing the `members` assertion to a single call to `assert` at the end.
LGTM.

lib/chai/core/assertions.js
```diff
@@ -1547,8 +1564,9 @@ module.exports = function (chai, _) {
   Assertion.addMethod('closeTo', closeTo);
   Assertion.addMethod('approximately', closeTo);
 
-  function isSubsetOf(subset, superset, cmp) {
-    return subset.every(function(elem) {
+  function isSubsetOf(subset, superset, cmp, ordered) {
+    return subset.every(function(elem, idx) {
+      if (ordered) return cmp ? cmp(elem, superset[idx]) : elem === superset[idx];
```
lucasfcosta: I'll leave this here just to make other reviewer's job easier.
This line is related to the behaviour we talked about at [this post](https://github.com/chaijs/chai/issues/717#issuecomment-225432216).

meeber:
@lucasfcosta Thank you for reviewing :D

keithamus:
Great work @meeber 😄. LGTM!

PR #729, opened 2016-06-14
docs: add @meeber to core contributor avatars

keithamus:
I just noticed @meeber's lovely profile pic was missing from our readme. This will not stand, and must be fixed! P1 issue!


meeber:
This breaks all of Chai's tests!! But we must merge anyway!! LGTM

keithamus:
Hmm. We need someone else to verify this change. @lucasfcosta what do you think? 😉

lucasfcosta:
LGTM!
:tada: :tada: :tada: :tada:

PR #732, opened 2016-06-20
Added backers and sponsors from OpenCollective

piamancini:
Now sponsors & backers of your collective will show up here together with a button to your collective's public page.

PR #738, opened 2016-06-24
docs: fix link to OpenCollective

keithamus:
I think we were... okay I was... a bit hasty on merging this. The link went no where. This one goes to our Open Collective page. #732

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@
 
 [![Slack Status](https://chai-slack.herokuapp.com/badge.svg)]( https://chai-slack.herokuapp.com/)
 [![Join the chat at https://gitter.im/chaijs/chai](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/chaijs/chai)
-[![OpenCollective](https://opencollective.com/chaijs/backers/badge.svg)](#backers) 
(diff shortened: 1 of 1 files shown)


meeber:
LGTM

lucasfcosta:
Oh, I didn't notice this too.
I should've checked the link.
Well, now everything is working, our OpenCollective page is very nice BTW.

LGTM.

keithamus:
I may have to disable the check for travis considering how flaky our tests are and they prevent merging

lucasfcosta:
@keithamus Makes sense, since we can't run builds on IE for PRs there's no way we can ensure tests will pass both on the PR check and the check after a merge. However, I'd prefer keeping these checks and having to run builds on IE manually on my machine just to make sure everything passes. I think this could now be a new default step of the merge process since we would have to clone the forks and run tests locally anyway before merging.

Thoughts @meeber and @keithamus ?

meeber:
I'd like to keep the checks. They encourage us to fix whatever browser-specific problem popped up from previous PR before moving on.

keithamus:
@meeber We can keep the checks, but right now if the check fails it prevents from merging which means small changes like this get blocked until an admin merges or the tests pass and the pr is rebased. I just wanted to disable the gate that would prevent one of you two merging should the build fail.

PR #739, opened 2016-06-25
Fix duplicate handling in members assertion

### LINKED ISSUE #590: Handling of duplicates when comparing sets
Current behaviour is that 
```
[3,1,2,3].should.have.deep.members([3,2,1]);      
[1,2,3].should.have.deep.members([3,2,1,3]);
[1,2,3].should.deep.include.members([3,2,1,3]);
chai.assert.sameDeepMembers([1,2,3],[3,2,1,3]);
```
all pass.
The documentation uses language of "sets" not "multisets" so it is hard to judge what should be the good behaviour when the arguments contain duplicates. 
One could argue that current implementation of `should.deep.include.members` is consistent with the documentation, as indeed every element of `[3,2,1,3]` belongs to `[1,2,3]`.
Similarly, one could argue similarly for `should.have.deep.members`.
But in practice I found myself in situations where what I really want to test is if the `actual` is a permutation of `expected`. There are many situations in which an order in the array cannot be guaranteed (say, `Object.keys(obj)`).

The issue might seem not so controversial when the arguments are written as literals and one can easily see what is the problem.
But in a more realistic scenarios like:
```
emittedEvents.should.have.deep.members(expectedEvents)
```
one could easily be fooled into thinking that this assertion simply means "expectedEvents is a permutation of expectedEvents" only to be surprised that firing the same event twice does not violate this test. 
Similarly one could have something like:
```
sorted.should.have.deep.members(inputArray)
```
and never spot an error in the implementation which caused removal of duplicates.

Perhaps it would be a good thing to have a separate `should.be.permutation.of` assertion for that case. But care must be taken to redact documentation for the other assertions so they do not suggest wrongly that they could be used for that.

meeber:
- Changes `same.members` to properly handle duplicates by treating each one as a unique member
- Fixes #590 

Prior to this PR, the behavior was for duplicates to be ignored when performing any kind of members assertion. This led to unexpected behavior when asserting for sameness, as exemplified in #590. We consider this to be a bug.

With this PR, duplicates are no longer ignored when asserting for sameness, but are still ignored when asserting for inclusion. Bug fixed!

An argument could be made that duplicates should never be ignored, even for inclusion assertions. But  that's a breaking change instead of a bug fix. Therefore, that debate will be left for another day. It'd be easy to make that change in the future if agreed upon.

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -1571,15 +1571,33 @@ module.exports = function (chai, _) {
   Assertion.addMethod('closeTo', closeTo);
   Assertion.addMethod('approximately', closeTo);
 
-  function isSubsetOf(subset, superset, cmp, ordered) {
+  // Note: Duplicates are ignored if testing for inclusion instead of sameness.
+  function isSubsetOf(subset, superset, cmp, contains, ordered) {
+    if (!contains) {
+      if (subset.length !== superset.length) return false;
+      superset = superset.slice();
+    }
+
     return subset.every(function(elem, idx) {
       if (ordered) return cmp ? cmp(elem, superset[idx]) : elem === superset[idx];
-      if (!cmp) return superset.indexOf(elem) !== -1;
 
-      return superset.some(function(elem2) {
-        return cmp(elem, elem2);
+      if (!cmp) {
+        var matchIdx = superset.indexOf(elem);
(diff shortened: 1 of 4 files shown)


lucasfcosta:
Nice job!
LGTM.

keithamus:
Awesome work @meeber. LGTM 👍

PR #740, opened 2016-06-25
Fix tests for .throws using getConstructor name instead of .name prop

lucasfcosta:
This fixes the failing tests for the `.throws` assertion.
The problem was that I was using `.name` to access the name of an Error instead of calling the `getConstructorName` function, which runs a polyfill for IE.

Sorry for the delay on this one, I had problems with karma and my IE VMs yesterday.


meeber:
LGTM

keithamus:
LGTM 😄

PR #742, opened 2016-06-26
Throws polishment assert interface

lucasfcosta:
Hello guys!
This includes polishments for the `Assert` interface due to the changes on the `throws` assertion.
I have also updated the docStrings for each one of the `throws` related methods in order to make sense with the new behavior, but since I'm not a native english speaker I would appreciate if you guys could review it carefully just to make sure everything is correct.

While doing this I also noticed that we had problems in the way we were handling empty strings for errors, so I included that in a separate commit. Now we do an explicit check to see if the `caughtErr` or the `errMsgMatcher` are `null` or `undefined`, because empty Strings evaluate to `false`.

I was also thinking if we should handle throwing raw empty Strings or other kinds of objects instead of `Error` objects, what do you think? Should we handle those too?

--------------------------
**Edit**: I also improved the DocStrings for the `.throws` assertion explaining which kind of comparison is used for each type of argument.

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -1402,7 +1402,7 @@ module.exports = function (chai, _) {
       }
     }
 
-    if (errMsgMatcher) {
+    if (errMsgMatcher && caughtErr) {
       // Here we check compatible messages
       var placeholder = 'including';
       if (errMsgMatcher instanceof RegExp) {
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -1285,8 +1285,12 @@ module.exports = function (chai, _) {
    * ### .throw(constructor)
    *
    * Asserts that the function target will throw a specific error, or specific type of error
-   * (as determined using `instanceof`), optionally with a RegExp or string inclusion test
+   * (as determined using `instanceof`), optionally with a RegExp or String inclusion test
    * for the error's message.
+   * If an Error instance is provided, it asserts that the error thrown and the instance
+   * provided are the same.
+   * If an Error constructor is provided, it asserts that the error thrown is an instance
+   * of the constructor provided.
    *
    *     var err = new ReferenceError('This is a bad function.');
    *     var fn = function () { throw err; }
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -1402,7 +1402,7 @@ module.exports = function (chai, _) {
       }
     }
 
-    if (errMsgMatcher && caughtErr) {
+    if (caughtErr && errMsgMatcher !== undefined && errMsgMatcher !== null) {
       // Here we check compatible messages
       var placeholder = 'including';
       if (errMsgMatcher instanceof RegExp) {
--- a/lib/chai/interface/assert.js
+++ b/lib/chai/interface/assert.js
@@ -1234,19 +1234,21 @@ module.exports = function (chai, util) {
   /**
    * ### .throws(fn, [errorLike/string/regexp], [string/regexp], [message])
    *
-   * Asserts that `fn` will throw an error that is an instance of `errorLike`
-   * (if `errorLike` is an `Error` constructor) or if it is the same instance as `errorLike`
-   * (if `errorLike` is an `Error` instance). This also asserts that the error thrown
-   * will have a message matching `errMsgMatcher`.
+   * If `errorLike` is an `Error` constructor, asserts that `fn` will throw an error that is an
+   * instance of `errorLike`.
+   * If `errorLike` is an `Error` instance, asserts that the error thrown is the same
+   * instance as `errorLike`.
+   * If `errMsgMatcher` is provided, it also asserts that the error thrown will have a
+   * message matching `errMsgMatcher`.
    *
    *     assert.throws(fn, 'function throws a reference error');
    *     assert.throws(fn, /function throws a reference error/);
    *     assert.throws(fn, ReferenceError);
-   *     assert.throws(fn, new ReferenceError());
+   *     assert.throws(fn, errorInstance);
    *     assert.throws(fn, ReferenceError, 'Error thrown must be a ReferenceError and have this msg');
-   *     assert.throws(fn, new ReferenceError(), 'Error thrown must be the same ReferenceError instance and have this msg');
+   *     assert.throws(fn, errorInstance, 'Error thrown must be the same errorInstance and have this msg');
    *     assert.throws(fn, ReferenceError, /Error thrown must be a ReferenceError and match this/);
-   *     assert.throws(fn, new ReferenceError(), /Error thrown must be the same ReferenceError instance and match this/);
+   *     assert.throws(fn, errorInstance, /Error thrown must be the same errorInstance and match this/);
    *
    * @name throws
    * @alias throw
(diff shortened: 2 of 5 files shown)


lib/chai/core/assertions.js
```diff
@@ -1402,7 +1402,8 @@ module.exports = function (chai, _) {
       }
     }
 
-    if (errMsgMatcher) {
+    if (errMsgMatcher !== undefined && errMsgMatcher !== null &&
+        caughtErr !== undefined && caughtErr !== null) {
```
meeber: Recommend refactor 1405-1406 to:

```js
    if (caughtErr && errMsgMatcher != null) {
```

Reasons:

- `caughtErr` doesn't need empty-string protection; it's either an error or undefined, so a simple truthy test should suffice.
- It's generally considered acceptable to use a loose equality comparison when checking for the `undefined` or `null` combo.

lucasfcosta: What you said about `caughtErr` makes sense, I did it because I thought of handling throwing empty strings at first, but then changed my mind and thought about asking what you think before handling anything other than `Error` objects.

Hmmm, I don't feel so confortable using loose equality comparisons, IMO having some more characters to explicitly check for `null` and `undefined` makes things more explicit, even though the effect is the same in the end. What do you think? If you guys still prefer using loose equality we can do it though 😄

lib/chai/interface/assert.js
```diff
@@ -1232,66 +1232,78 @@ module.exports = function (chai, util) {
...
    *     assert.throws(fn, 'function throws a reference error');
    *     assert.throws(fn, /function throws a reference error/);
    *     assert.throws(fn, ReferenceError);
-   *     assert.throws(fn, ReferenceError, 'function throws a reference error');
-   *     assert.throws(fn, ReferenceError, /function throws a reference error/);
+   *     assert.throws(fn, new ReferenceError());
+   *     assert.throws(fn, ReferenceError, 'Error thrown must be a ReferenceError and have this msg');
+   *     assert.throws(fn, new ReferenceError(), 'Error thrown must be the same ReferenceError instance and have this msg');
+   *     assert.throws(fn, ReferenceError, /Error thrown must be a ReferenceError and match this/);
+   *     assert.throws(fn, new ReferenceError(), /Error thrown must be the same ReferenceError instance and match this/);
```
meeber: I'm torn about the examples in 1245, 1247, and 1249. Although I understand they're merely meant to demonstrate what happens when an instance is provided as the second argument instead of a constructor, I'm not crazy about the fact that these examples would fail 100% of the time no matter what, since there's no way `fn` could possibly throw the same Error instance that's being instantiated in this way. I'd feel fuzzier if these examples demonstrated more realistic usage.

One idea is to change `new ReferenceError()` in the examples to `errInstance`.

lib/chai/core/assertions.js
meeber: The way you have it now is fine. It's a stylistic preference. Only mentioned it cause I was already commenting anyway on the `caughtErr` part. :D

lib/chai/interface/assert.js
```diff
@@ -1232,66 +1232,78 @@ module.exports = function (chai, util) {
...
-   * ### .throws(function, [constructor/string/regexp], [string/regexp], [message])
+   * ### .throws(fn, [errorLike/string/regexp], [string/regexp], [message])
    *
-   * Asserts that `function` will throw an error that is an instance of
-   * `constructor`, or alternately that it will throw an error with message
-   * matching `regexp`.
+   * Asserts that `fn` will throw an error that is an instance of `errorLike`
+   * (if `errorLike` is an `Error` constructor) or if it is the same instance as `errorLike`
+   * (if `errorLike` is an `Error` instance). This also asserts that the error thrown
+   * will have a message matching `errMsgMatcher`.
```
meeber: I like the new phrasing but suggest maybe reordering it a bit. Just an example:

```js
If `errorLike` is an `Error` constructor, assert that `fn` will throw an error that is an instance of `errorLike`.
If `errorLike` is an `Error` instance, assert that `fn` will throw the same instance as `errorLike`.
If `errMsgMatcher` is provided, also assert that the error thrown will have a message matching `errMsgMatcher`.
```

lib/chai/interface/assert.js
```diff
@@ -1232,66 +1232,78 @@ module.exports = function (chai, util) {
...
-   * ### .doesNotThrow(function, [constructor/regexp], [message])
+   * ### .doesNotThrow(fn, [errorLike/string/regexp], [string/regexp], [message])
    *
-   * Asserts that `function` will _not_ throw an error that is an instance of
-   * `constructor`, or alternately that it will not throw an error with message
-   * matching `regexp`.
+   * Asserts that `fn` will _not_ throw an error that is an instance of `errorLike`
+   * (if `errorLike` is an `Error` constructor) or if it is the same instance as `errorLike`
+   * (if `errorLike` is an `Error` instance). This also asserts that the error thrown
+   * will _not_ have a message matching `errMsgMatcher`.
```
meeber: Same thing. Consider reordering. The readability is a little worse on this one than the previous one.

```js
If `errorLike` is an `Error` constructor, assert that `fn` won't throw an error that is an instance of `errorLike`.
If `errorLike` is an `Error` instance, assert that `fn` won't throw the same instance as `errorLike`.
If `errMsgMatcher` is provided, also assert that the error thrown won't have a message matching `errMsgMatcher`.
```

lib/chai/interface/assert.js
```diff
@@ -1232,66 +1232,78 @@ module.exports = function (chai, util) {
...
    *
-   *     assert.doesNotThrow(fn, Error, 'function does not throw');
+   *     assert.doesNotThrow(fn, 'Any Error thrown must not have this message');
+   *     assert.doesNotThrow(fn, /Any Error thrown must not match this/);
+   *     assert.doesNotThrow(fn, Error);
+   *     assert.doesNotThrow(fn, new Error());
+   *     assert.doesNotThrow(fn, Error, 'Error must not have this message');
+   *     assert.doesNotThrow(fn, new Error(), 'Error must not have this message');
+   *     assert.doesNotThrow(fn, Error, /Error must not match this/);
+   *     assert.doesNotThrow(fn, new Error(), /Error must not match this/);
```
meeber: Same feeling about the `new Error()`. Consider `errInstance` instead, since that's how it'd typically be used.

meeber:
@lucasfcosta Great catch on the empty string, I hadn't thought about that! Thanks for updating the assert interface too :D

Just made a couple suggestions with phrasing and examples in the line notes above.

Personally I don't think we should support throwing anything except error types.

lib/chai/interface/assert.js
lucasfcosta: Awesome suggestion, I haven't thought about that, thanks 👍

lucasfcosta:
@meeber Thanks for your input, it was great (as always).
Here are the changes you talked about.

I was also thinking about improving the explanation on the `strict` comparison for instances and `instanceof` comparison for constructors at [the `.throws` assertion docStrings](https://github.com/chaijs/chai/blob/master/lib/chai/core/assertions.js#L1285), what do you think? Would it be useful or would it be too redudant?

meeber:
I agree, I think it'd be good to improve the explanation there, using what you just did with the assert interface docs as a basis.

lucasfcosta:
@meeber Here it is. What do you think? Is this clear and grammatically correct?

meeber:
LGTM!

keithamus:
LGTM 👍

Issue #743, opened 2016-06-30
expect(array).to.not.include(obj) failing with constructors


GabeMedrash:
This may be linked to https://github.com/chaijs/chai/issues/390, but I'm not sure so reporting this separately:

Environment:
  - Chai version: 3.5.0
  - Node version: 6.2.0

Test case:
```js
function SomeConstructor() {}
const instance1 = new SomeConstructor();
const instance2 = new SomeConstructor();

expect(instance1).to.not.equal(instance2) // => passes as expected

const arr = [instance1];
expect(arr).to.include(instance1) // => passes as expected
expect(arr).to.not.include(instance2) // => fails; unexpected
```

As a workaround, I can change the constructor to: 
```js
function SomeConstructor() {
  this.uid = Symbol();
}
```
and `expect(arr).to.not.include(instance2)` will work fine.

lucasfcosta:
Hi @GabeMedrash, thanks for the issue!

I think you are right and this is linked to #390.
I'm now reading [this code here](https://github.com/chaijs/chai/blob/master/lib/chai/core/assertions.js#L233) and it seems like if we've got an `object` inside an `Array`, we look for `val` inside every index of the `Array` [using `deep-eql` from utils](https://github.com/chaijs/chai/blob/master/lib/chai/utils/index.js#L70). The problem is that **deep-eql compares an object's primitive keys one by one instead of doing an strict equality check (`===`)**, this means that if your instances have the same properties Chai will think that they're the same (because that is what `deep-eql` does).

I think that we should fix this by adding the possibility of using the `deep` flag for the `includes` assertion. However, doing this would be a breaking change, since the current default behavior is to use `deep-eql`. This could break a lot of people. If we want this to be a more 'soft' change maybe we could have an `strict` flag instead of `deep` flag.

What do you think @meeber and @keithamus (anyone else is also welcome to share your thoughts on this, I'm just tagging these two guys so they can see it 😄 )?

In any case, implementing that should be reasonably easy, all we've gotta do is add a check before running `_.eql` and then running a strict comparison (`indexOf` or anything like that) instead of `_.eql` when we don't have the `deep` flag turned on (or when we have the `strict` flag turned on, whatever you guys prefer).

meeber:
@GabeMedrash As a workaround I recommend using `.members`:

```js
    expect(arr).to.include.members([instance1]);
    expect(arr).to.not.include.members([instance2]);
```

[...]

Issue #16, opened 2012-02-13
negated property(name, value) assertion


jfirebaugh:
I would expect a negated assertion to pass if and only if it would fail if not negated. But this is not true of the `property(name, value)` assertion:

```
expect('asd').to.have.property('foo', 3); // fails
expect('asd').not.to.have.property('foo', 3); // also fails (I would expect it to pass)
```

I realize the test verify the existing behavior, but is there any chance it could change?

PR #744, opened 2016-07-02
Breaking: Change `.not.property(name, val)` behavior

meeber:
- Previously, `expect(obj).not.property(name, val)` would throw an Error
if `obj` didn't have a property named `name`. This change causes the
assertion to pass instead.
- assert.propertyNotVal renamed to assert.notPropertyVal
- assert.deepPropertyNotVal renamed to assert.notDeepPropertyVal

Notes: 

- This PR is in reference to the issue discussed in https://github.com/chaijs/chai/issues/743#issuecomment-230112615
- It supersedes the decision made in https://github.com/chaijs/chai/issues/16
- It's a prerequisite for completing #743 in such a way that `.include` internally leverages `.property` when performing property assertions while providing a consistent, intuitive experience.

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -835,6 +835,9 @@ module.exports = function (chai, _) {
    *     var obj = { foo: 'bar' };
    *     expect(obj).to.have.property('foo');
    *     expect(obj).to.have.property('foo', 'bar');
+   *     expect(obj).to.not.have.property('baz');
+   *     expect(obj).to.not.have.property('foo', 'baz');
+   *     expect(obj).to.not.have.property('baz', 'bar');
    *
    *     // deep referencing
    *     var deepObj = {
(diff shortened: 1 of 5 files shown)


keithamus:
This LGTM 👍

keithamus:
@lucasfcosta thoughts?

lucasfcosta:
LGTM too.
Since there's no other PR which this one depends upon, I will get this one merged first.

meeber:
@lucasfcosta My thought is that if everything looks good we can merge these in right before starting the 4.x.x-to-master merge.

lucasfcosta:
@meeber Hmmm, I've had that doubt especially because of #751.
I was concerned that if we had that before the `4.0` release we would release a breaking canary version, which would be a really undesirable thing.

keithamus:
@lucasfcosta why would releasing a breaking canary version be a problem?

lucasfcosta:
@keithamus I thought our goal with that would be just to release non-breaking versions, but including latest features and fixes, so anyone could using the `canary` tag in their `package.json` wouldn't get a build broken due to that.

However, I'm not too strict about it, if guys prefer this way I don't see any problem due to the fact that anyone using the `canary` tag should be aware they're using experimental and new code.

keithamus:
@lucasfcosta My assumption is that `canary` would include any features in the next upcoming version of chai. The next upcoming version of chai is `4.0` and as such has breaking changes. It gives folk a chance to fix their tests or brace themselves before `4.0` lands.

meeber:
I'm in favor of doing things in this order:

1. Merge these breaking changes into master
1. Merge 4.x.x into master
1. Release a canary version for testing by our user base
1. Start writing migration guide and/or scripts
1. Merge any bug fixes as needed
1. Merge deep-eql refactor into master
1. Release updated canary version for testing
1. Merge any bug fixes as needed
1. Finish writing migration guide and/or scripts
1. Depending on number of changes, release a final updated canary version for testing
1. Release official v4.0.0

lucasfcosta:
Thank you for sharing you ideas guys, it's great to have such a great team to discuss these kinds of things 😄 
I totally agree with @meeber's post.

PR #750, opened 2016-07-09
Fix new Promise example

shvaikalesh:
`Promise` will throw unless `executor` is a function. Please, see [25.4.3.1, step 2](https://tc39.github.io/ecma262/#sec-promise-executor) of spec.


meeber:
Nice catch. LGTM

keithamus:
LGTM

PR #757, opened 2016-07-26
Breaking: Rename `.deep.property` to `.nested.property`

meeber:
- Previously, the `deep` flag had unexpected behavior when used with the
`property` assertion: Instead of switching from strict to deep equality,
it unlocked the ability to use dot and bracket notation when referencing
property names. This behavior now belongs to the new `nested` flag.
- assert.deepProperty renamed to assert.nestedProperty
- assert.notDeepProperty renamed to assert.notNestedProperty
- assert.deepPropertyVal renamed to assert.nestedPropertyVal
- assert.notDeepPropertyVal renamed to assert.notNestedPropertyVal

Note:
- #744 is a prerequisite for this PR. I recommend reviewing #744 first, and then only reviewing the newest commit in this PR.
- This PR is a prerequisite for completing #743 in such a way that `.include` internally leverages `.property` when performing property assertions while providing a consistent, intuitive experience.

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -835,6 +835,9 @@ module.exports = function (chai, _) {
    *     var obj = { foo: 'bar' };
    *     expect(obj).to.have.property('foo');
    *     expect(obj).to.have.property('foo', 'bar');
+   *     expect(obj).to.not.have.property('baz');
+   *     expect(obj).to.not.have.property('foo', 'baz');
+   *     expect(obj).to.not.have.property('baz', 'bar');
    *
    *     // deep referencing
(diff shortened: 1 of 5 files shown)


keithamus:
LGTM 👍. @lucasfcosta

lucasfcosta:
Awesome job regarding the docs. Very clear instructions. :+1:
Code LGTM too. **But read this before merging**.

Regarding this one I have the same doubt as I've had [here](https://github.com/chaijs/chai/pull/744#issuecomment-239472847).
Shouldn't breaking changes be merged to `4.x.x` first instead of `master`?

PR #758, opened 2016-07-26
Add `.deep.property` for deep equality comparisons

### LINKED ISSUE #745: Unexpected meaning of `deep` flag in `property` assertion
There's an inconsistency with the meaning of the `deep` flag. Consider these examples:

```js
const obj = {a: 1};
expect(obj).to.deep.equal({a: 1});  // Currently passes thanks to deep equality
expect([obj]).to.have.deep.members([{a: 1}]);  // Currently passes thanks to deep equality
expect([obj]).to.deep.include({a: 1});  // Will pass thanks to deep equality once Issue #743 is fixed
expect({foo: obj}).to.deep.include({foo: {a: 1}});  // Will pass thanks to deep equality once Issue #743 is fixed
expect({foo: obj}).to.have.deep.property('foo', {a: 1});  // Fails due to using strict equality
```

The reason the last example fails is because the `deep` flag has a different meaning when applied to the `.property` assertion than it does when applied to any other assertion. Instead of switching from strict to deep equality, the `deep` flag unlocks the ability for the developer to use dot or bracket notation in their `.property` assertion:

```js
const obj = {a: 1};
expect({foo: obj}).to.have.deep.property('foo.a', 1);
```

That's pretty cool, but the inconsistency of the `deep` flag creates a confusing developer experience and brings sadness to my soul. It also makes it impossible for other assertions such as `.include` to internally leverage `.property` by passing along the `deep` flag, as noted in https://github.com/chaijs/chai/issues/743#issuecomment-230186471.

I propose the following breaking change:

- Make the `deep` flag perform a deep equality comparison when used with the `.property` assertion.
- Add either a `nested` or `inner` flag to inherit the old behavior of the `deep` flag, allowing dot or bracket notation in the `.property` assertion.

Examples:

```js
const obj = {a: 1};
expect({foo: obj}).to.have.deep.property('foo', {a: 1});
expect({foo: obj}).to.have.nested.property('foo.a', 1);

const obj2 = {a: {b: 2}};
expect({foo: obj2}).to.have.deep.nested.property('foo.a', {b: 2});
```

meeber:
Notes:
- #744 and #757 are prerequisites for this PR. I recommend reviewing them first, and then only reviewing the newest commit in this PR.
- This PR is a prerequisite for completing #743 in such a way that `.include` internally leverages `.property` when performing property assertions while providing a consistent, intuitive experience.
- Resolves #745

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -835,6 +835,9 @@ module.exports = function (chai, _) {
    *     var obj = { foo: 'bar' };
    *     expect(obj).to.have.property('foo');
    *     expect(obj).to.have.property('foo', 'bar');
+   *     expect(obj).to.not.have.property('baz');
+   *     expect(obj).to.not.have.property('foo', 'baz');
+   *     expect(obj).to.not.have.property('baz', 'bar');
    *
    *     // deep referencing
    *     var deepObj = {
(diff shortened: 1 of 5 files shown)


keithamus:
LGTM 👍. @lucasfcosta?

lucasfcosta:
LGTM too.

meeber:
Let's see if this works:  LGTM

lucasfcosta:
Reposting because LGTM was down before.

PR #760, opened 2016-07-28
Breaking: Fix `.include` to always use strict equality

meeber:
- Previously, `.include` was using strict equality for non-negated property
inclusion, but deep equality for negated property inclusion and array
inclusion. This fix causes `.include` to always use strict equality.

Note:

- #744, #757, and #758 are prerequisites for this PR. I recommend reviewing them first, and then only reviewing the newest commit in this PR.
- This PR is a prerequisite for completing #743.

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -7,6 +7,7 @@
 
 module.exports = function (chai, _) {
   var Assertion = chai.Assertion
+    , AssertionError = chai.AssertionError
     , toString = Object.prototype.toString
     , flag = _.flag;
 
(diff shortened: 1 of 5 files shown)


keithamus:
Yup, LGTM. @lucasfcosta?

lib/chai/core/assertions.js
```diff
@@ -221,6 +222,17 @@ module.exports = function (chai, _) {
    *     expect([1,2,3]).to.include(2);
    *     expect('foobar').to.contain('foo');
    *     expect({ foo: 'bar', hello: 'universe' }).to.include({ foo: 'bar' });
+   * 
+   * By default, strict equality (===) is used.
```
lucasfcosta: Maybe this line isn't totally clear for anyone using the library.
When saying that this uses `strict` equality this may not be exactly clear for everyone. This is because when saying that this assertion uses this kind of comparison I would expect (if I just read the docs) the following assertion to throw an error:

```js
expect({foo: obj1, bar: obj2}).to.include({foo: obj1});
```

Because **I would expect it to compare the object I passed as a parameter to the include method to the one I passed to `expect`, and, since they would not be the same instance, strict equality would be false** and therefore I would expect an `Error` to be raised.

Maybe we could explain this a little better, just to make things as clear as possible, I'm not sure everyone will get it right this way.

What do you guys think?

meeber: @lucasfcosta Agreed. Updated with an explanation.

meeber:
@lucasfcosta @keithamus New version pushed!

lucasfcosta:
Excelent explanation. The inner working of this assertion is now very clear.
LGTM.

Issue #766, opened 2016-08-06
Make proxies optional


meeber:
Some people might not want to have their assertions boxed inside proxies, perhaps due to performance considerations. Since proxies are already automatically disabled in environments that don't support proxies, it should be easy to also make it possible to manually disable them via a Chai config value.

meeber:
This is a good opportunity for anyone wanting to contribute to Chai. New contributors encouraged. Post a comment here before starting so no one duplicates effort. Also please review the [contributor guidelines] (https://github.com/chaijs/chai/blob/master/CONTRIBUTING.md) before opening a PR. The team is happy to answer any questions.

Requirements:

- Add tests proving that proxies are used when `useProxy` is `true`, and aren't used when `useProxy` is `false` (see [test/configuration.js] (https://github.com/chaijs/chai/blob/master/test/configuration.js))
- The new tests should only be run if the current environment supports proxies (see [test/expect.js] (https://github.com/chaijs/chai/blob/fbb4a4e085eca9520146a94ea5fdade7d4508442/test/expect.js#L13-L36) for example)
- Add a new config directive called `useProxy` that defaults to `true` (see [lib/chai/config.js] (https://github.com/chaijs/chai/blob/master/lib/chai/config.js))
- Update logic to only use proxies when `useProxy` is set to `true` (see [lib/chai/utils/proxify.js](https://github.com/chaijs/chai/blob/master/lib/chai/utils/proxify.js))
- If the user's environment doesn't support proxies, then the existing behavior of not using proxies should be maintained, even if the value of `useProxy` is set to `true`

vieiralucas:
Hello @meeber 
I'm interested in solving this issue.

Thank you

keithamus:
Hey @vieiralucas - if you're interested in fixing this then please, go right ahead and make a PR. Put it up as soon as you want us to critique it and we can make sure it's what we want before merging 👍

meeber:
Closed via https://github.com/chaijs/chai/pull/770

Issue #691, opened 2016-04-27
expect(null).to.be.within(0,10) does not fail assertion


helfer:
**Expected behavior:**
`expect(null).to.be.within(0,1)` should return false for any numbers a and b.

**Actual behavior:**
`expect(null).to.be.within(0,1)` returns true. Sam for other falsey values, I assume, and similar problem with true.

**Proposed fix:**
Within should check that the value is a number and reject if it isn't before comparing the range [here](https://github.com/chaijs/chai/blob/df4a261ec80f02889c02efeab16dfc7244bc0285/lib/chai/core/assertions.js#L769).

meeber:
Hi @helfer, thanks for the report!

As a workaround, I'd recommend the following form:
```js
expect(null).to.be.a('number').within(0,1)
```

I would worry about the implications of rejecting non-numbers here. Your example returns `true` because `null` gets coerced into a number (specifically, `+0`) by the `<=` and `>=` operators, per the ECMAScript spec. If anyone out there has written tests that rely on this per-spec coercion, then backward compatibility will be broken. Which isn't necessarily a problem, if this is deemed to be a bug in the `within` logic.

But what about other types of coercion? These valid tests would start failing if only numbers were allowed through:

- expect("3").to.be.within(1,4)
- expect("b").to.be.within("a", "c")

So what we might actually be talking about here is allowing some types of per-spec coercions to occur, but not others. I dunno. It feels like a tough sell, but at the same time I agree it's disorienting for `null` to return `true` in the specific example you provided. However, I also think the already-available workaround is an elegant solution.

Thoughts?

helfer:
@meeber Thanks, the workaround makes sense, I'll use that.

The main reason I was surprised by this behavior is that all the examples in the documentation use numbers, and do not suggest that within coerces other types. I know Javascript is quite loose when it comes to comparators, but I think most people would expect a testing library to be much less tolerant.

To keep backwards compatibility, it's probably best not to change the behavior, but the documentation should be updated to mention that this works with other types and that a('number') should be used if a number is expected.

meeber:
Agreed. If we move forward with the current functionality, then the documentation should indicate so and include examples of coercion. Additionally, tests should be added demonstrating the coercion.

Going to leave this issue open for awhile for further feedback and/or possible pull-request-wanted for docs/tests update.

keithamus:
[...]

PR #768, opened 2016-08-12
Fix indentation of includeStack at config

vieiralucas:
Hello guys,

I was looking at #766, when suddenly, I had an OCD attack when I saw this poorly indented line 😄
I understand that this is clearly me being nitpicking and I will be totally fine if you think that this should not be merged.

Thanks

### DIFF
--- a/lib/chai/config.js
+++ b/lib/chai/config.js
@@ -13,7 +13,7 @@ module.exports = {
    * @api public
    */
 
-   includeStack: false,
+  includeStack: false,
 
   /**
    * ### config.showDiff


keithamus:
LGTM 😆 

@meeber? @lucasfcosta? Could use your expert eyes here.

guax:
👾

keithamus:
OMG I didn't even notice we have another contributor named Lucas! This is definitely a conspiracy.

meeber:
@keithamus We need to remain vigilant at all times. The Lucas Syndicate has already announced their intentions for world domination per [this comment] (https://github.com/chaijs/chai/issues/691#issuecomment-215993108). I think they perceive Chai as a stepping stone toward that end.

PR #770, opened 2016-08-13
Make proxies optional Fix #766

### LINKED ISSUE #766: Make proxies optional

vieiralucas:
Hello

First of all, thank you @meeber for describing what, where and how to do this.
All I had to do was add the code. You are awesome.

Please let me know if I'm missing something here.

Thank you all ❤️


lib/chai/config.js
```diff
@@ -50,6 +50,21 @@ module.exports = {
    * @api public
    */
 
-  truncateThreshold: 40
+  truncateThreshold: 40,
 
+  /**
+   * ### config.useProxy
+   *
+   * User configurable property, define if chai will use a Proxy to throw
```
lucasfcosta: **Nitpicky Warning:** The word `define` here should be written as `defines`, shouldn't it?
(Please correct me if I'm wrong, I'm not a native speaker)

test/configuration.js
```diff
@@ -167,6 +167,33 @@ describe('configuration', function () {
...
+      expect(false).to.be.tue; // typo: tue should be true
+    };
+
+    it('should have default value equal to true', function() {
+      expect(chai.config.useProxy).to.be.true;
+    });
+
+    describe('when true', function() {
+      it('should use proxy unless user\'s environment doesn\'t support', function() {
+        if (typeof Proxy !== 'undefined' || typeof Reflect !== 'undefined') {
```
lucasfcosta: Here we will never test the `else` clause on environments that do support `Proxy`.
What do you guys think about assigning `undefined` to `Proxy` after testing it if the environment supports it, just to see if the code inside the `else` clause will still be valid on every environment? Otherwise, if we test only on environments with `Proxy` support, that condition will never be tested.

lucasfcosta:
BTW I forgot to thank you for this awesome PR.
Since we get to talk to each other almost everyday on college I forgot that.
Just left some minor notes about grammar and another thought I had about test's coverage.

lib/chai/config.js
meeber: Agreed on "defines" over "define".

Also maybe be a bit more explicit on the purpose of this feature. For example adding onto the end:

"...non-existent property is read, which protects users from typos when using property-based assertions."

vieiralucas: Agreed with both you guys. I add the suggestions.
Thanks

test/configuration.js
meeber: The `if` clause should only be tested if both Proxy and Reflect are defined. Therefore, we want `&&` here instead of `||`.

meeber:
@vieiralucas Awesome work sir. Thanks for doing this :D 

It looks like you already fixed one of the two line comments I made while reviewing. Once the other is resolved I think this is set to merge.

Btw, how does one go about joining the Lucas Syndicate? Is being named Lucas a hard requirement?

test/configuration.js
vieiralucas: @lucasfcosta I'm not very happy with an `if` inside the unit test to. This is making the test, test a thing or another depending on a condition, which is very strange. But for chai's context, as we talked it might be a good solution.

@meeber Wow, nice catch. You are absolutely right, thank you.

meeber:
LGTM :D

vieiralucas:
@meeber Thank you, and all the maintainers of chai for making this project such a friendly place to new contributors.

I believe that being named Lucas would be a requirement for joining our Syndicate, but maybe we can add an exception for you.

test/configuration.js
lucasfcosta: Yeah, nice catch.
Also, @meeber and @keithamus what do you think about what we talked regarding always testing the code inside the `else` clause? I'd love to hear your opinion on this matter :smile:

meeber: @lucasfcosta I'm in favor of keeping it the way it is. I prefer achieving complete test coverage by testing in different environments and browsers rather than shimming stuff or whatever the opposite of shimming would be by setting a built-in to `undefined`.

lib/chai/config.js
```diff
@@ -50,6 +50,22 @@ module.exports = {
...
+  truncateThreshold: 40,
 
+  /**
+   * ### config.useProxy
+   *
+   * User configurable property, defines if chai will use a Proxy to throw
+   * an error when a non-existent property is read, which protects users
+   * from typos when using property-based assertions.
+   *
+   * Set it to false if you want to disable this feature.
```
meeber: I just had another thought on this: We should mention that this feature is automatically disabled regardless of this config value in environments that don't support proxies.

vieiralucas: Agreed, should I put this before the example or after?

```javascript
  /**
   * ### config.useProxy
   *
   * User configurable property, defines if chai will use a Proxy to throw
   * an error when a non-existent property is read, which protects users
   * from typos when using property-based assertions.
   * This feature is automatically disabled regardless of this config value
   * in environments that don`t support proxies.
   *
   * Set it to false if you want to disable this feature.
   *
   *     chai.config.useProxy = false;  // disable use of Proxy
   *
   * @param {Boolean}
   * @api public
   */
```

vs

```javascript
  /**
   * ### config.useProxy
   *
   * User configurable property, defines if chai will use a Proxy to throw
   * an error when a non-existent property is read, which protects users
   * from typos when using property-based assertions.
   *
   * Set it to false if you want to disable this feature.
   *
   *     chai.config.useProxy = false;  // disable use of Proxy
   *
   * This feature is automatically disabled regardless of this config value
   * in environments that don`t support proxies.
   *
   * @param {Boolean}
   * @api public
   */
```

meeber: I like the second

vieiralucas: Done

meeber:
LGTM

lucasfcosta:
LGTM to me too.

PR #774, opened 2016-08-23
Configurable proxy blacklist

### LINKED ISSUE #765: Cannot console.log assertions because of proxies
Consider:

```js
const {expect} = require("chai");

it("meep", () => {
  const myAssertion = expect(42).to.equal(42);

  console.log(myAssertion);
});
```

In Chai v3.5, the output is:

> Assertion {
>   __flags: { ssfi: [Function], object: 42, message: undefined } }
>   ✓ meep
> 
>   1 passing (25ms)

But in the current version of Chai off Github, the output is:

>   1) meep
> 
>   0 passing (13ms)
>   1 failing
> 
>   1)  meep:
>      Error: Invalid Chai property: inspect
>       at Error (native)
>       at Object.getProperty [as get] node_modules/chai/lib/chai/utils/proxify.js:30:15
>       at formatValue (util.js:343:19)
>       at inspect (util.js:198:10)
>       at exports.format (util.js:75:24)
>       at Console.log (console.js:43:37)
>       at Context.it (test/index.js:6:11)

Possible solution:

Add `inspect` as a blacklisted property [here](https://github.com/chaijs/chai/blob/fbb4a4e085eca9520146a94ea5fdade7d4508442/lib/chai/utils/proxify.js#L27-L29). `then` already exists as a blacklisted property for promise support. `inspect` is a standard property in Node, as noted in https://github.com/chaijs/chai/issues/727#issuecomment-225373838. 

Rather than hard-coding the blacklist, it might be more appropriate to refactor the blacklist to be a new Chai config value.

lucasfcosta:
This aims to close #765.

As brilliantly suggested by @meeber this adds a new Array containing keys which should be ignored when checking for non-existing properties on an assertion before throwing an error.

This PR contains:
- [x] An awesome and difficult replacement of a `backtick` which should have been an `'`
- [x] A new setting on `chai.config` called `proxyExcludedKeys` which receives an array of Strings representing the properties which should be ignored by the proxy on environments that support it
- [x] Docs explaining how it works, default values and other peculiarities
- [x] Tests checking:
 * Default values
 * If it does not throw an error on non existing `then` and `inspect` properties
 * If it throws an error on non existing `then` and `inspect` properties if they get removed from the `proxyExcludedKeys` array on environments that do support proxies

Let me know if I missed anything.

### DIFF
--- a/lib/chai/config.js
+++ b/lib/chai/config.js
@@ -70,5 +70,25 @@ module.exports = {
    * @api public
    */
 
-  useProxy: true
+  useProxy: true,
+
+  /**
+   * ### config.proxyExcludedKeys
+   *
+   * User configurable property, defines which properties should be ignored
+   * instead of throwing an error if they do not exist on the assertion.
+   * This is only applied if the environment Chai is running in supports proxies and
+   * if the `useProxy` configuration setting is enabled.
+   * By default, `then` and `inspect` will not throw an error if they do not exist on the
+   * assertion object because the `.inspect` property is read by `util.inspect` (for example, when
+   * using `console.log` on the assertion object) and `.then` is necessary for promise type-checking.
+   *
+   *     // By default these keys will not throw an error if they do not exist on the assertion object
+   *     chai.config.proxyExcludedKeys = ['then', 'inspect'];
+   *
+   * @param {Array}
+   * @api public
+   */
+
+  proxyExcludedKeys: ['then', 'inspect']
 };
--- a/lib/chai/config.js
+++ b/lib/chai/config.js
@@ -64,7 +64,7 @@ module.exports = {
    *     chai.config.useProxy = false;  // disable use of Proxy
    *
    * This feature is automatically disabled regardless of this config value
-   * in environments that don`t support proxies.
+   * in environments that don't support proxies.
    *
    * @param {Boolean}
    * @api public
--- a/lib/chai/utils/proxify.js
+++ b/lib/chai/utils/proxify.js
@@ -24,10 +24,12 @@ module.exports = function proxify (obj) {
 
   return new Proxy(obj, {
     get: function getProperty (target, property) {
-      // Don't throw error on Symbol properties such as Symbol.toStringTag, nor
-      // on .then because it's necessary for promise type-checking.
+      // This check is here because we should not throw errors on Symbol properties
+      // such as `Symbol.toStringTag`.
+      // The values for which an error should be thrown can be configured using
+      // the `config.proxyExcludedKeys` setting.
       if (typeof property === 'string' &&
-          property !== 'then' &&
+          config.proxyExcludedKeys.indexOf(property) === -1 &&
           !Reflect.has(target, property))
         throw Error('Invalid Chai property: ' + property);
 
--- a/test/configuration.js
+++ b/test/configuration.js
@@ -196,4 +196,37 @@ describe('configuration', function () {
       });
     });
   });
+
+  describe('proxyExcludedKeys', function() {
+    var readNoExistentProperty = function(prop) {
+      return function() {
+        var assertion = expect(false);
+        expect(assertion).to.not.have.key(prop);
+        assertion[prop];
+      }
+    };
+
+    it('should have default value equal to `[\'then\', \'inspect\']`', function() {
+      expect(chai.config.proxyExcludedKeys).to.be.deep.equal(['then', 'inspect']);
+    });
+
+    it('should not throw when accessing non-existing `then` and `inspect` in an environment with proxy support', function() {
+      // Since these will not throw if the environment does not support proxies we don't need any `if` clause here
+      expect(readNoExistentProperty('then')).to.not.throw();
+      expect(readNoExistentProperty('inspect')).to.not.throw();
+    });
+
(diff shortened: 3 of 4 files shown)


keithamus:
This LGTM (looks _great_ to me). @meeber thoughts?

lib/chai/config.js
```diff
@@ -64,11 +64,31 @@ module.exports = {
...
-  useProxy: true
+  useProxy: true,
+
+  /**
+   * ### config.proxyExcludedKeys
+   *
+   * User configurable property, defines which non-existing properties should be ignored
+   * instead of throwing an error if they do not exist on the assertion.
+   * This is only applied if the environment Chai is running into supports proxies and
+   * if the `useProxy` configuration setting is enabled.
```
meeber: Couple nits:

- Line 78: Can remove  "non-existing" since it's redundant with "if they do not exist on the assertion" later on in the same sentence.
- Line 80: "running in" instead of "running into".

lib/chai/config.js
```diff
@@ -64,11 +64,31 @@ module.exports = {
...
+   *
+   *     // By default these keys will not throw an error if they do not exist on the assertion object
+   *     chai.config.proxyExcludedKeys = ['then', 'inspect'];
+   *
+   * This feature is automatically disabled regardless of this config value
+   * in environments that don't support proxies.
```
meeber: Lines 86-87 aren't needed because it's already mentioned in lines 80-81.

lib/chai/config.js
```diff
@@ -64,11 +64,31 @@ module.exports = {
...
+  /**
+   * ### config.proxyExcludedKeys
+   *
+   * User configurable property, defines which non-existing properties should be ignored
+   * instead of throwing an error if they do not exist on the assertion.
+   * This is only applied if the environment Chai is running into supports proxies and
+   * if the `useProxy` configuration setting is enabled.
+   *
+   *     // By default these keys will not throw an error if they do not exist on the assertion object
+   *     chai.config.proxyExcludedKeys = ['then', 'inspect'];
```
meeber: This is probably the best place to add a small note explaining why these two are ignored by default (instead of in proxify.js).

lib/chai/utils/proxify.js
```diff
@@ -26,8 +26,10 @@ module.exports = function proxify (obj) {
     get: function getProperty (target, property) {
       // Don't throw error on Symbol properties such as Symbol.toStringTag, nor
       // on .then because it's necessary for promise type-checking.
```
meeber: Since the `.then` default exclusion is no longer defined here, I think it'd be best to move the explanation on line 28 to the config file. I think the note you added on lines 29-30 is enough.

test/configuration.js
```diff
@@ -196,4 +196,37 @@ describe('configuration', function () {
...
+      expect(chai.config.proxyExcludedKeys).to.be.deep.equal(['then', 'inspect']);
+    });
+
+    it('should not throw when accessing non-existing `then` and `inspect` on an environment with proxy support', function() {
+      // Since these will not throw if the environment does not support proxies we don't need any `if` clause here
+      expect(readNoExistentProperty('then')).to.not.throw();
+      expect(readNoExistentProperty('inspect')).to.not.throw();
+    });
+
+    it('should throw for properties which are not on the `proxyExcludedKeys` Array on an environment with proxy support', function() {
```
meeber: Nits:

- Line 213+219: "in an environment" instead of "on"
- Line 219: "in the proxyExcludedKeys Array" instead of "on"

meeber:
@lucasfcosta Awesome work, thanks for doing this! I just had a few nits related to comments. I really like how you handled the changes, particularly the tests with the helper function. You've done the world a great service by fixing that errant backtick.

lucasfcosta:
Thanks for the help @meeber, since I'm not a native English speaker I make this kind of mistakes sometimes.
Now the comments really seem less redundant and more informative, thank you 😄 

Let me know if I forgot anything.

lib/chai/config.js
```diff
@@ -64,11 +64,31 @@ module.exports = {
...
+  /**
+   * ### config.proxyExcludedKeys
+   *
+   * User configurable property, defines which properties should be ignored
+   * instead of throwing an error if they do not exist on the assertion.
+   * This is only applied if the environment Chai is running in supports proxies and
+   * if the `useProxy` configuration setting is enabled.
+   * By default, `then` and `inspect` will not throw an error if they do not exist on the
+   * assertion object because we should not throw errors on Symbol properties such as
+   * `Symbol.toStringTag` nor on `.then` because it is necessary for for promise type-checking.
```
meeber: The part about Symbol properties and `Symbol.toStringTag` can remain in proxify.js because that's where the type check is; it's different than the configurable excluded keys. A new note needs to be added here about `.inspect`; it's ignored by default because that property is read by `util.inspect`, for example when using `console.log` on the assertion object.

meeber:
@lucasfcosta I'm jealous of your ability to speak multiple languages!

I added a follow-up note above about the comments.

lucasfcosta:
@meeber thanks for noticing that!
I have made the fixes you suggested and I also read the `inspect` utility and found some other minor typos, so I also fixed them.

meeber:
LGTM!

Issue #777, opened 2016-08-25
False positives with .to.exist


AlexEscalante:
I am using Mocha, Enzyme and Chai, and I am getting false positives on tests like these:

    import React from 'react';
    import {shallow} from 'enzyme';
    import {expect} from 'chai';

    it('renders a barcode', () => {
      const wrapper = shallow(<ReceiptPage {...props} />);
    
      expect(wrapper.find('bogus')).to.exist;
    });

That test should not pass and it does. I should be doing something wrong, but I can't figure out what… Any ideas?

lucasfcosta:
Hi @AlexEscalante, thanks for your issue!
Can you give me some more details about this? I'd like to know what does `wrapper.find('bogus')` returns, is it really `undefined` or is it `null`?
As you can see [here](https://github.com/chaijs/chai/blob/master/lib/chai/core/assertions.js#L483) the code for our `exist` assertion is really simple, I can't really think of why this is happening.

I also noticed that we're using `!=` here. It works, but perhaps writing `<whatever> !== undefined && <whatever> !== null` would be more explicit and it would also attend to most of JS code recommendations. What do you guys think?

lucasfcosta:
Hi @AlexEscalante, thanks for your issue!
Can you give me some more details about this? I'd like to know what does `wrapper.find('bogus')` returns, is it really `undefined` or is it `null`?
As you can see [here](https://github.com/chaijs/chai/blob/master/lib/chai/core/assertions.js#L483) the code for our `exist` assertion is really simple, I can't really think of why this is happening.

I noticed that we're using `!=` here. It works, but perhaps writing `<whatever> !== undefined || <whatever> !== null` would be more explicit and it would also meet most of JS code recommendations.

keithamus:
AFAICT this is because `ShallowWrapper#find` returns a new wrapper object, so that you can chain those kinds of commands.  A `Wrapper` "exists" and so the assertion is failing because `Wrapper !== null && Wrapper !== undefined`.

I would recommend, @AlexEscalante, using something like [chai-enzyme](https://github.com/producthunt/chai-enzyme) - which [overrides `.to.exist`](https://github.com/producthunt/chai-enzyme/#present) to detect for Enzyme Wrapper objects, and so would return correctly. Alternatively, you could reword your assertion to something like `expect(wrapper.find('bogus')).to.have.lengthOf(0);`.

Hopefully this helps. I'll close the issue for now but let us know if you have any more problems with it.

[...]

PR #778, opened 2016-08-27
Improve exist assertion (Triple equals and addition to assert interface)

lucasfcosta:
After reading #777 I noticed we were using `null != flag(this, 'object')` on our `exist` assertion. That is correct indeed (given our main goal with this assertion), but I feel more comfortable using `!==` to check if the value is neither `null` or `undefined`, which makes thinks more explicit.

I also added some tests just to make sure `falsy` values or anything "strange" would work.

And at last, but not least, the `exist` assertion was added to the `assert` interface as `assert.exists` and `assert.notExists`.

------------------------------------

For bonus style points on this commit, here goes a picture of me and @vieiralucas with Jake Archibald:

Jake Archibald Being Awesome with Us

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -481,8 +481,9 @@ module.exports = function (chai, _) {
    */
 
   Assertion.addProperty('exist', function () {
+    var val = flag(this, 'object');
     this.assert(
-        null != flag(this, 'object')
+        val !== null && val !== undefined
       , 'expected #{this} to exist'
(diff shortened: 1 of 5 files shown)


keithamus:
LGTM. @meeber?

meeber:
@lucasfcosta Excellent job! I agree with all the changes and the code looks clean. Not sure how I feel about the appearance of meeber in the tests but I suppose as long as meeber is awesome it's okay :D

LGTM

PR #779, opened 2016-09-01
Adding `.finite`. Fixes #271

### LINKED ISSUE #271: assert.isNumber(NaN) passes
The title says it all. I can see that, [according to the spec](http://www.ecma-international.org/ecma-262/5.1/#sec-4.3.20), this may be considered correct behavior, but it runs pretty contrary to developer expectations.

At very least, it would be great to see a note about this in the docs.

TheSavior:
As discussed in #271, `.isNumber` currently passes for `NaN`. This issue recently bit our team and so we wanted to move forward on the proposal in that issue. 

The implementation is the polyfill from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isFinite.

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -2151,16 +2151,15 @@ module.exports = function (chai, _) {
    *
    * Asserts that the target is a finite number. Unlike `.a('number')`, this will fail for `NaN` and `Infinity`.
    *
-   *     expect(4).to.be.finite();
-   *     expect(NaN).to.not.be.finite();
+   *     expect(4).to.be.finite;
+   *     expect(NaN).to.not.be.finite;
    *
    * @name finite
    * @namespace BDD
    * @api public
    */
 
-  Assertion.addMethod('finite', function(msg) {
-    if (msg) flag(this, 'message', msg);
+  Assertion.addProperty('finite', function(msg) {
     var obj = flag(this, 'object');
 
     this.assert(
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -2149,9 +2149,10 @@ module.exports = function (chai, _) {
   /**
    * ### .finite
    *
-   * Asserts that the target is a finite number. Unlike `.a('number')`, this will fail for `NaN` and `Infinity`
+   * Asserts that the target is a finite number. Unlike `.a('number')`, this will fail for `NaN` and `Infinity`.
    *
-   *     expect(foo).to.be.finite();
+   *     expect(4).to.be.finite();
+   *     expect(NaN).to.not.be.finite();
    *
    * @name finite
    * @namespace BDD
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -2145,4 +2145,27 @@ module.exports = function (chai, _) {
       , 'expected #{this} to not be frozen'
     );
   });
+
+  /**
+   * ### .finite
+   *
+   * Asserts that the target is a finite number. Unlike `.a('number')`, this will fail for `NaN` and `Infinity`
+   *
+   *     expect(foo).to.be.finite();
+   *
+   * @name finite
+   * @namespace BDD
+   * @api public
+   */
+
+  Assertion.addMethod('finite', function(msg) {
+    if (msg) flag(this, 'message', msg);
+    var obj = flag(this, 'object');
+
+    this.assert(
+        typeof obj === "number" && isFinite(obj)
+      , 'expected #{this} to be a finite number'
+      , 'expected #{this} to not be a finite number'
+    );
+  });
 };
--- a/lib/chai/interface/assert.js
+++ b/lib/chai/interface/assert.js
@@ -735,7 +735,6 @@ module.exports = function (chai, util) {
    /**
    * ### .isFinite(value, [message])
    *
-   * Asserts that `value` is a finite number. This will fail for NaN, unlike `.isNumber`
    * Asserts that `value` is a finite number. Unlike `.isNumber`, this will fail for `NaN` and `Infinity`
    *
    *     var cups = 2;
--- a/lib/chai/interface/assert.js
+++ b/lib/chai/interface/assert.js
@@ -750,7 +750,7 @@ module.exports = function (chai, util) {
    */
 
   assert.isFinite = function (val, msg) {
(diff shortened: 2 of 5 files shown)


lib/chai/interface/assert.js
```diff
@@ -732,6 +732,26 @@ module.exports = function (chai, util) {
     new Assertion(val, msg).to.not.be.a('number');
   };
 
+   /**
+   * ### .isFinite(value, [message])
+   *
+   * Asserts that `value` is a finite number. This will fail for NaN, unlike `.isNumber`
+   * Asserts that `value` is a finite number. Unlike `.isNumber`, this will fail for `NaN` and `Infinity`
```
amarcher: The comment on line 738 seems redundant with that on line 739

lucasfcosta: Agreed.
IMO leaving only the last one would be more appropriate.

lucasfcosta:
Hi @TheSavior, thanks for your contribution!
I've taken a look at the issue you referenced and I agree we should have an explicit assertion to check if something is neither `infinity` nor `NaN`.
Since @keithamus already said he thinks having this assertion would be a good idea too, I'll ping @meeber to hear his opinion.

-------------------------------------------

**Regarding the Code:**

Good job! The code looks clean and has tests, but I'd like to ask you to add some [tests for the `should` interface](https://github.com/chaijs/chai/blob/master/test/should.js) too.
Also, as @amarcher has just said, [the comment on this line](https://github.com/chaijs/chai/pull/779#discussion-diff-77220011R738) seems redundant.

Thanks again for your PR! 😄

lib/chai/core/assertions.js
```diff
@@ -2145,4 +2145,27 @@ module.exports = function (chai, _) {
       , 'expected #{this} to not be frozen'
     );
   });
+
+  /**
+   * ### .finite
+   *
+   * Asserts that the target is a finite number. Unlike `.a('number')`, this will fail for `NaN` and `Infinity`
+   *
+   *     expect(foo).to.be.finite();
```
lucasfcosta: Maybe we could change `foo` for any other number and add an example using `.not` and `NaN` or `Infinity`.

lib/chai/core/assertions.js
```diff
@@ -2145,4 +2145,27 @@ module.exports = function (chai, _) {
       , 'expected #{this} to not be frozen'
     );
   });
+
+  /**
+   * ### .finite
+   *
+   * Asserts that the target is a finite number. Unlike `.a('number')`, this will fail for `NaN` and `Infinity`
```
lucasfcosta: Also, I know I'm being nitpicky, but maybe we could have a final dot in this phrase.

lib/chai/interface/assert.js
```diff
@@ -732,6 +732,26 @@ module.exports = function (chai, util) {
...
+   *
+   *     var cups = 2;
+   *     assert.isFinite(cups, 'how many cups');
```
lucasfcosta: Since the main goal of this assertion is to prove something is not `NaN` or `Infinity`, it would be nice to have an example using `.not` and these two values, just to make the docs more clear.

lib/chai/interface/assert.js
TheSavior: Yeah, this was a copy paste error

lib/chai/core/assertions.js
```diff
@@ -2145,4 +2145,28 @@ module.exports = function (chai, _) {
...
+  Assertion.addMethod('finite', function(msg) {
+    if (msg) flag(this, 'message', msg);
+    var obj = flag(this, 'object');
+
+    this.assert(
+        typeof obj === "number" && isFinite(obj)
+      , 'expected #{this} to be a finite number'
+      , 'expected #{this} to not be a finite number'
+    );
+  });
```
meeber: I think this should be a property assertion (like `.frozen`) instead of a method assertion.

test/assert.js
```diff
@@ -459,6 +459,19 @@ describe('assert', function () {
...
+
+    err(function () {
+      assert.isFinite('1');
+    }, "expected \'1\' to be a finite number");
+
+    err(function () {
+      assert.isFinite(NaN);
+    }, "expected NaN to be a finite number");
+  })
+
```
meeber: Preferable that the tests on the `assert` interface are the same as the tests on the `should` and `expect` interfaces.

meeber:
@TheSavior Thank you for your contribution! This seems like a great addition to Chai. I made a couple  minor suggestions above.

@lucasfcosta @keithamus Is the Chai repo configured to allow us to automatically squash commits into a single one on merge?

lib/chai/core/assertions.js
keithamus: Yes I kind of agree; as much as I think we need to get rid of the property interface, we also need to stick to form more. If it takes no args, its a property.

lucasfcosta: Makes sense, I thought we would start using only method assertions

meeber:
LGTM!

keithamus:
Yes yes, LGTM. Let's get it in.

TheSavior:
Awesome. Thanks guys. 

What is the process from here to get this published? Do you guys have a release cadence? Do you try to wait for many things to have larger releases? Or do you just publish new versions as features land?

keithamus:
@TheSavior right now we're on the ramp up to a big release - 4.0 - with a few breaking changes. Once that's done I think it's likely we'll end up releasing as new features land. Ultimately for this PR this means it'll be a few weeks, perhaps more, before it's out. Sorry about that.

TheSavior:
Oh man, sad day. Alright. Thanks for the heads up. That helps align my expectations. 👍

keithamus:
@TheSavior if you're keen to use this feature, you can simply add the parts of the PR to a bootstrap file somewhere in your test code! Then simply remove it when you upgrade to chai 4.0 😄

TheSavior:
Yeah, that's our plan. 👍

lucasfcosta:
@TheSavior you can also start using the current master branch right away by using this repo as the package mirror.

Take a look at [this link](https://docs.npmjs.com/files/package.json#git-urls-as-dependencies) if you'd like to do so! But be aware there may be breaking changes since we will be merging `4.x.x` here. 😄

PR #782, opened 2016-09-04
Add fix suggestions when accessing a nonexistent property in proxy mode

### LINKED ISSUE #771: Use levenshtein distance to suggest alternatives in proxy mode
We should use the levenshtein distance algo to determine correct properties to use as suggestions in the errors for unknown properties - when using proxies.

For example
```js
expect(true).to.be.ture
// Error: unknown property .ture
```

Could be 
```js
expect(true).to.be.ture
// Error: unknown property .ture, did you mean .true?
```

not-an-aardvark:
fixes #771 

As suggested in #771, when a nonexistent property is accessed in proxy mode, this computes the levenshtein distance to all possible properties in order to suggest the best fix to the user.

```js
chai.expect(false).to.be.fals; // Error: Invalid Chai property: fals. Did you mean "false"?
chai.expect('foo').to.be.undefind; // Error: Invalid Chai property: undefind. Did you mean "undefined"?
```

I decided to not suggest a fix if the closest property has a string distance of 4 or more, simply because the suggestion probably will not be very useful at that point.

```js
chai.expect('foo').to.be.fdsakfdsafsagsadgagsdfasf // error thrown, no fix suggested
```

### DIFF
--- a/lib/chai/utils/proxify.js
+++ b/lib/chai/utils/proxify.js
@@ -1,4 +1,5 @@
 var config = require('../config');
+var getProperties = require('./getProperties');
 
 /*!
  * Chai - proxify utility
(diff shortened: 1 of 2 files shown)


keithamus:
Great work here @not-an-aardvark! LGTM. This may make you the coolest non-aardvark I've met!

@meeber @lucasfcosta thoughts?

lucasfcosta:
Hi @not-an-aardvark, thanks for your contribution, I'd like to frame this code and hang it on my wall! It looks really elegant!
Whenever I see an intelligent recursion such as this one my heart beats stronger.

I've gotta confess I needed half an hour to truly understand how that works, but it was totally worth it.

Just in case anyone else in the future also needs some help with this, take a look at [this link](http://www.levenshtein.net/), it's an excellent learning resource.

LGTM!

not-an-aardvark:
> I'd like to frame this code and hang it on my wall! It looks really elegant! 😄 

Thanks!

> Just in case anyone else in the future also needs some help with this, take a look at [this link](http://www.levenshtein.net/), it's an excellent learning resource.

In addition, I would also recommend [this Wikipedia article](https://en.wikipedia.org/wiki/Levenshtein_distance) on Levenshtein distances. The implementation here is based on the article's definition [here](https://en.wikipedia.org/wiki/Levenshtein_distance#Definition).

meeber:
Please don't merge yet, I need to explore a few things with this.

meeber:
This looks pretty neat. But I think we should reduce the suggestion list to Chai keywords only.

If we change `getProperties(target)` to ` Object.getOwnPropertyNames(Object.getPrototypeOf(target))`, then the following suggestions from `Object.prototype` get removed:

```
<   '__defineGetter__',
<   '__defineSetter__',
<   '__flags',
<   '__lookupGetter__',
<   '__lookupSetter__',
<   '__proto__',
<   'hasOwnProperty',
<   'isPrototypeOf',
<   'propertyIsEnumerable',
<   'should',
<   'toLocaleString',
<   'toString',
<   'valueOf',
```

That leaves the following list with three needing to be manually filtered out:

```js
[ 'Arguments',
  'NaN',
  'Throw',
  '__methods',             # Manually filter out; internal use
  '_obj',                  # Manually filter out; internal use
  'a',
  'above',
  'all',
  'an',
  'and',
  'any',
  'approximately',
  'arguments',
  'assert',                # Manually filter out; internal use
  'at',
  'be',
  'been',
  'below',
  'but',
  'by',
  'change',
  'changes',
  'closeTo',
  'constructor',
  'contain',
  'contains',
  'decrease',
  'decreases',
  'deep',
  'does',
  'empty',
  'eq',
  'eql',
  'eqls',
  'equal',
  'equals',
  'exist',
  'extensible',
  'false',
  'finite',
  'frozen',
  'greaterThan',
  'gt',
  'gte',
  'has',
  'have',
  'haveOwnProperty',
  'haveOwnPropertyDescriptor',
  'include',
  'includes',
  'increase',
  'increases',
  'instanceOf',
  'instanceof',
  'is',
  'itself',
  'key',
  'keys',
  'least',
  'length',
  'lengthOf',
  'lessThan',
  'lt',
  'lte',
  'match',
  'matches',
  'members',
  'most',
  'nested',
  'not',
  'null',
  'of',
  'ok',
  'oneOf',
  'ordered',
  'ownProperty',
  'ownPropertyDescriptor',
  'property',
  'respondTo',
  'respondsTo',
  'same',
  'satisfies',
  'satisfy',
  'sealed',
  'string',
  'that',
  'throw',
  'throws',
  'to',
  'true',
  'undefined',
  'which',
  'with',
  'within' ]
```

not-an-aardvark:
> If we change `getProperties(target)` to `Object.getOwnPropertyNames(Object.getPrototypeOf(target))`

That works, but it will cause issues if someone tries to subclass `chai.Assertion`. (I'm not sure whether this is a supported use-case.)

As an alternative, we could specifically exclude keys in `Object.prototype`:

```js
getProperties(target).filter(function (property) {
  return !Object.prototype.hasOwnProperty(property) && ['__methods', '_obj', 'assert'].indexOf(property) === -1;
});
```

meeber:
@not-an-aardvark Alternative looks good.

not-an-aardvark:
Updated to use the alternative version, and added a couple tests for it.

meeber:
@not-an-aardvark LGTM! Thanks for doing this :D

Let's get one more LGTM before merging since there were changes. @keithamus @lucasfcosta?

lucasfcosta:
LGTM too!
Great catch meeber, I haven't thought about it, but now it makes total sense.

PR #692, opened 2016-04-30
[WIP] Verify types on within and related assertions

vieiralucas:
Hello there, this will be my proposal solution for #691 

So why this is a WIP...
I started coding this from the branch master.
When I came here to make the PR, I realized that the 4.x.x branch was 166 commits behind master and that my PR could not be merged.
Well, no problem. I checkout to 4.x.x and made some cherry-picks (actually ctrl-c ctrl-v :smile:)
BUT, I saw that some features that exists on master did not exist here on 4.x.x

The features which are missing on 4.x.x that affect this PR:
- `assert.atLeast`
- `assert.atMost`
- `increase|decrease...by()`
 
What should I do about it?

Things that I'm not sure how to do
- [ ] Documentation updated to indicate type requirements

PR #783, opened 2016-09-05
Merge rebased 4.x.x branch into master

### LINKED ISSUE #772: Merge 4.x.x branch into master
@keithamus @lucasfcosta I'm headed to the opposite side of the world on Thursday and am gonna be stuck in airplanes for a ridiculous amount of time. That means I need a big project to keep me occupied :D

I'd like to take the opportunity to merge the 4.x.x branch into master in preparation for the canary release. Any objections? If no objections, any suggestions? I've never merged two branches together with this substantial of a commit difference (250ish and counting). Would it better to rebase the 4.x.x branch using the master branch? Or perform a traditional merge? Or some other approach? I realize that either way there's gonna be a conflict resolution stage, but I'm a bit fuzzy on the github flow for this kind of operation.

Thanks :D

meeber:
I ran into one minor issue while rebasing the 4.x.x branch:

There was a test in the 4.x.x branch that relied on being able to access an Assertion object's flags after the assertion had failed and been caught in a try/catch block. However, a change in the master branch made it so that assertions return a new Assertion object with flags transferred over, instead of returning the original Assertion object. This means that any subsequent flags are set on the new Assertion object instead of the original. For a successful assertion, this is fine, since the new Assertion object is returned. But in this case, for a failed assertion that throws an error and is caught in a try/catch block, the test no longer has access to any of the failed Assertion object's flags that were added after the first step in the assertion.

I resolved this issue in an extra commit that fixes this broken test by getting rid of the `.to` chain so that the test has access to the relevant Assertion object and its flags.

In addition to the issue above, I made a couple of side notes:

- A few tests broke in Node v6.5 due to a change in how anonymous functions that are assigned to a variable appear in failed assertion error messages. This issue is completely unrelated to 4.x.x. I'll be following up this PR with a separate one that addresses that.
- There's now a large divergence between the behavior of the `.property` and `.ownProperty` assertions due to heavy refactoring of `.property` in master and minor refactoring of `.ownProperty` in the 4.x.x branch. I'll be following up this PR with a separate one that addresses that.

(Closes #772)

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -446,7 +446,7 @@ module.exports = function (chai, _) {
    * ### .NaN
    * Asserts that the target is `NaN`.
    *
-   *     expect('foo').to.be.NaN;
+   *     expect(NaN).to.be.NaN;
    *     expect(4).not.to.be.NaN;
    *
    * @name NaN
@@ -456,7 +456,7 @@ module.exports = function (chai, _) {
 
   Assertion.addProperty('NaN', function () {
     this.assert(
-        isNaN(flag(this, 'object'))
+        _.isNaN(flag(this, 'object'))
         , 'expected #{this} to be NaN'
         , 'expected #{this} not to be NaN'
     );
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -1767,6 +1767,7 @@ module.exports = function (chai, _) {
       , failNegateMsg
       , subset
       , obj
+      , true
     );
   });
 
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -1021,28 +1021,59 @@ module.exports = function (chai, _) {
 
 
   /**
-   * ### .ownProperty(name)
+   * ### .ownProperty(name, [value])
    *
-   * Asserts that the target has an own property `name`.
+   * Asserts that the target has an own property `name` and, optionally, if it has
+   * (or not, if using `.not`) the desired `value`.
    *
    *     expect('test').to.have.ownProperty('length');
+   *     expect('test').to.haveOwnProperty('length');
+   *     expect('test').to.not.have.ownProperty('foo');
+   *     expect('test').to.not.haveOwnProperty('foo');
+   *     expect({ length: 12 }).to.have.ownProperty('length', 12);
+   *     expect({ length: 1337 }).to.not.have.ownProperty('length', 20);
+   *     expect({ length: 12 }).to.haveOwnProperty('length', 12);
+   *     expect({ length: 1337 }).to.not.haveOwnProperty('length', 20);
    *
    * @name ownProperty
    * @alias haveOwnProperty
    * @param {String} name
+   * @param {Mixed} value (optional)
    * @param {String} message _optional_
    * @namespace BDD
    * @api public
    */
 
-  function assertOwnProperty (name, msg) {
+  function assertOwnProperty (name, value, msg) {
     if (msg) flag(this, 'message', msg);
     var obj = flag(this, 'object');
-    this.assert(
-        Object.prototype.hasOwnProperty.call(obj, name)
-      , 'expected #{this} to have own property ' + _.inspect(name)
-      , 'expected #{this} to not have own property ' + _.inspect(name)
-    );
+    var negate = flag(this, 'negate');
+    var objHasProperty = Object.prototype.hasOwnProperty.call(obj, name)
+    var actualValue = obj[name];
+
+    if (negate && value !== undefined) {
+      if (actualValue === undefined) {
+        throw new Error(_.inspect(obj) + ' does not have own property ' + _.inspect(name));
+      }
+    } else {
+      this.assert(
+          objHasProperty
+        , 'expected #{this} to have own property ' + _.inspect(name)
+        , 'expected #{this} to not have own property ' + _.inspect(name)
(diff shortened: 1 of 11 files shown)


keithamus:
This looks like some truly great work @meeber. LGTM, tagging @lucasfcosta for the merge.

Also interesting to see how much of 4.0 has been driven by @lucasfcosta - looks like many commits in that PR. Congrats on the stellar work, both of you!

meeber:
@keithamus Thank you sir, and agreed regarding Professor Lucas's contributions. Most of them were made before I joined the team, so this was my first time reviewing them. Unsurprisingly, they were excellent commits across the board! And everything was well-documented, so I was able to understand and resolve all conflicts during the rebase. Thank you @lucasfcosta!

lucasfcosta:
Yay! I'm happy to hear that! Thank you so much guys!
Actually, you are the ones who are awesome, I could not have made those commits without your help. Your ideas and reviews were essential to guarantee the quality of that work.
I said that yesterday on twitter, but I'd like to repeat it: I learned a lot by reading your code and comments.
I'll use my time this weekend to look at this with attention and make sure everything is okay.

Thanks for the work Mr. @meeber and Mr. @keithamus! It's an honor to be a part of this great team!

lib/chai/core/assertions.js
```diff
@@ -679,14 +684,15 @@ module.exports = function (chai, _) {
   /**
    * ### .least(value)
    *
-   * Asserts that the target is greater than or equal to `value`.
+   * Asserts that the number target is greater than or equal to `value`.
```
lucasfcosta: I know this is a merge, but I'll just leave this here as a note:
Shouldn't it be written `Asserts that the **target number** is greater than ...` instead of `Asserts that the **number target** ...

meeber: Yeah, the reason it was phrased that way is because that's how it was already phrased elsewhere in the code (see: https://github.com/chaijs/chai/pull/692#issuecomment-216054505). I think the thing for us to do is submit a separate PR that rephrases all of them (both old and new) at once.

lucasfcosta: Ahhh, I remember that now!
Thanks for refreshing my memory!
I totally agree with you, that would be a great idea, I will open an issue for that just so we won't forget.

lib/chai/core/assertions.js
```diff
@@ -679,14 +684,15 @@ module.exports = function (chai, _) {
...
    *
    * Can also be used in conjunction with `length` to
    * assert a minimum length. The benefit being a
    * more informative error message than if the length
-   * was supplied directly.
+   * was supplied directly. In this case the target must
+   * have a length property.
    *
    *     expect('foo').to.have.length.of.at.least(2);
    *     expect([ 1, 2, 3 ]).to.have.length.of.at.least(3);
```
lucasfcosta: Another side note here:

Currently we can't do:
`expect([ 1, 2, 3 ]).to.have.a.length.of.at.least(3);`
Because the `a` word returns a function which has a property called `.length` and there is no way we can remove that in some environments, but we could use a `Proxy` in environments that do support it to override what happens when trying to access `length` on the function returned by `addChainableBehavior`, what do you think? Is this possible? I didn't dig deep on this one, but I think it might be a valid strategy.

lib/chai/utils/inspect.js
```diff
@@ -318,6 +352,12 @@ function reduceToSingleString(output, base, braces) {
   return braces[0] + base + ' ' + output.join(', ') + ' ' + braces[1];
 }
 
+function isTypedArray(ar) {
+  // Unfortunately there's no way to check if an object is a TypedArray
+  // We have to check if it's one of these types
```
lucasfcosta: Another minor note:
I remember that when I first did this there was a list of types below this comment, now, since that list was removed it makes no sense anymore. Maybe we should find a way to update this comment line or even remove it. However, I'd like to change it to:

```js
// This regex checks if the object is one of the following types:
// Int8Array
// Int16Array
// Etc, etc, etc...
```

lib/chai/core/assertions.js
meeber: Yup, because `function.length` is configurable in modern environments, it can be intercepted with proxies or manually deleted without the use of proxies. I think one of them is worth doing (in a separate PR) but I'm not sure which approach is best.

lib/chai/utils/inspect.js
meeber: Sounds reasonable. Let's fix in separate PR.

lucasfcosta: Yep! 😄 
The goal of the notes I'm adding here is for me to remember what to do later, they aren't necessarily related to the `4.x.x` merge.
I'm just posting them to keep track of what should be done hehe

lucasfcosta:
I just went through it all and it LGTM!
Excellent work @meeber (as you always do)! 😄 
I left a couple of minor non-merge-related notes of things that should be done after this PR gets merged.

meeber:
@lucasfcosta Thanks for the careful review! I think this is ready to merge. Do you want to do the honors? :D

lucasfcosta:
@meeber Yes! Of course, I haven't seen @keithamus already approved this, thanks for noticing!

HERE WE GO!
Pushing the red button!

meeber:
@lucasfcosta Superb use of Ren and Stimpy gif.

lucasfcosta:
Thanks @meeber, it took years of practice 😄

Issue #124, opened 2012-12-21
Document supported browsers


eastridge:
We're running into some trouble running our test suite for [Thorax](http://walmartlabs.github.com/thorax) with chai. IE9 is fine, but we get this error on IE8:

" Object doesn't support property or method 'create' "

In addition we had to replace `to.be.true` with `to.be['true']` etc.

It would be a nice addition to the website to know which versions of what browsers chai officially supports.

keithamus:
It's come up a few times recently, so I'm going to try to push this issue a little more forward.

Here are my 2c about the whole browser support thing:

I think, given we're a testing framework, we need to support a broad range of versions. But also, given our limited person power, we need to be relatively conservative. I think supporting every engine that is officially supported by their respective vendors is a happy and relatively non-arbitrary point to fix to ([Node](https://github.com/nodejs/LTS), [Firefox](https://en.wikipedia.org/wiki/Firefox_release_history), [Chrome](https://en.wikipedia.org/wiki/Google_Chrome_release_history), [Opera](https://en.wikipedia.org/wiki/History_of_the_Opera_web_browser), [Safari](https://en.wikipedia.org/wiki/Safari_version_history), [Edge](https://en.wikipedia.org/wiki/Microsoft_Edge#Release_history), [IE](https://support.microsoft.com/en-us/help/17454/lifecycle-support-policy-faq-internet-explorer))

This means supporting Node 0.10 until October 1st or thereabouts, and 0.12 until the end of this year. IE wise, this means supporting IE9 until April 2017 (when Vista's extended support ends), IE10 and above for longer. For Firefox we'll support whatever the ESR version is, plus latest. Chrome, Opera and Safari only officially support their latest versions, so we'll do the same. All of these versions should run through CI so we know we're not breaking them.

I think in addition to all of that, we should provide what I'll call "soft" support for other versions. By soft support I mean if someone files an issue, we'll attempt to fix it - but sweeping architectural changes or large fixes will not be considered. This means users are welcome to try to use Chai in unsupported browsers, and if they encounter breakages, feel free to report them - however, also acknowledge that we might close issues saying we don't have manpower to provide the support you're after.

PR #784, opened 2016-09-05
Fix broken tests in Node v6.5 and update supported node versions

meeber:
- #783 is a prerequisite to this PR
- The first commit fixes a number of tests that were broken in Node v6.5 due to a change in how anonymous functions that are assigned to variables appear in failed assertion messages (from "[Function]" to "[Function: _var_name_]". 
- The second commit updates Travis Node versions (adds all versions of node that are officially maintained: 0.10, 0.12, 4, 6; also adds safety net aliases: lts/*, node).
- The third commit updates package.json's engines field to node v0.10 since it's the minimum version that we officially support.

Notes on Travis node versions:
- If there was only ever a total of two versions actively supported at the same time, one LTS and one Current, then the lts/* and node aliases would alone be sufficient. The challenge comes from the planned overlaps that are built into the Node release schedule, causing there to periodically be two LTS or two Current versions actively supported at once. The only way to prevent an actively supported version from being prematurely dropped from automated testing during these periods of overlap is to manually list all actively supported versions. The aliases lts/* and node should still be included as safety nets, ensuring that there's still good (but not perfect) coverage if package maintainers fail to keep on top of manually updating the version list. A downside of providing this safety net is that during periods of non-overlap, the test suite will run twice for the same release.

- Whether or not to use numeric versions like 4 or aliased versions like lts/argon is merely preference. I prefer the former because I think it more accurately reflects how most people discuss and interact with different Node versions.

- Including expiration date comments for each version in .travis.yml is good.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,12 @@ language: node_js
 addons:
   sauce_connect: true
 node_js:
-- 0.1
+ - 0.10 # to be removed 2016-10-01
+ - 0.12 # to be removed 2016-12-31
+ - 4 # to be removed 2018-04-01
+ - 6 # to be removed 2019-04-01
+ - lts/* # safety net; don't remove
+ - node # safety net; don't remove
 
 script:
 - make test-travisci
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "test": "make test"
   },
   "engines": {
-    "node": ">= 0.4.0"
+    "node": ">=0.10"
   },
   "dependencies": {
     "assertion-error": "^1.0.1",
--- a/test/expect.js
+++ b/test/expect.js
@@ -1546,87 +1546,87 @@ describe('expect', function () {
 
     err(function(){
       expect(goodFn).to.throw();
-    }, "expected [Function] to throw an error");
+    }, /^expected \[Function(: goodFn)*\] to throw an error$/);
 
     err(function(){
       expect(goodFn).to.throw(ReferenceError);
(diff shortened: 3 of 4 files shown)


keithamus:
These changes LGTM. We might want to consider dropping Node 0.10 for a clean break with 4.0 of chai. Should probably still support 0.12 for now though

meeber:
@keithamus I'd definitely like to discuss strategy for Node legacy support sometime soon. I haven't solidified my opinion on the issue yet. Thinking about it casually, I'm torn multiple ways.

keithamus:
@meeber I was going to write it here, but instead I decided to move my comment to #124. Let's continue discussion over there, and try to ratify something that users can read too.

meeber:
@keithamus I read your comment in the other thread and agree with you completely. I've updated this PR so that .travis.yml explicitly states all node versions that are officially maintained. I also added a third commit to bump up the minimum node version in package.json's engines field.

lucasfcosta:
LGTM.
I totally agree with @keithamus comment on #124.
Also, nice job adding the expire dates of the engines we're testing on `.travis.yml` file.

meeber:
@lucasfcosta Thanks for reviewing! Adding expire dates in `.travis.yml` was @keithamus's idea :D

I'm not sure we've reached complete consensus on the version list per https://github.com/chaijs/type-detect/pull/69. @keithamus?

keithamus:
Yup I think this is totally right and sets the precedent for all other projects in the chaijs repo. Awesome work, as usual @meeber 😄

PR #789, opened 2016-09-09
Wrap non-chainable methods in proxies

meeber:
Extends proxy protection to non-chainable methods (including overwritten non-chainable methods) with a helpful error message. Example:

```js
expect(true).to.equal.true;  // Invalid Chai property: equal.true. See docs for proper usage of "equal".
```

### DIFF
--- a/lib/chai/utils/proxify.js
+++ b/lib/chai/utils/proxify.js
@@ -11,15 +11,22 @@ var getProperties = require('./getProperties');
  * # proxify(object)
  *
  * Return a proxy of given object that throws an error when a non-existent
- * property is read. (If Proxy or Reflect is undefined, then return object
- * without modification.)
+ * property is read. By default, the root cause is assumed to be a misspelled
+ * property, and thus an attempt is made to offer a reasonable suggestion from
+ * the list of existing properties. However, if a nonChainableMethodName is
+ * provided, then the root cause is instead a failure to invoke a non-chainable
+ * method prior to reading the non-existent property.
+ * 
+ * If proxies are unsupported or disabled via the user's Chai config, then
+ * return object without modification.
  *
  * @param {Object} obj
+ * @param {String} nonChainableMethodName
  * @namespace Utils
  * @name proxify
  */
 
-module.exports = function proxify (obj) {
+module.exports = function proxify (obj, nonChainableMethodName) {
   if (!config.useProxy || typeof Proxy === 'undefined' || typeof Reflect === 'undefined')
     return obj;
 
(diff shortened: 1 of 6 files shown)


lib/chai/utils/proxify.js
```diff
@@ -32,6 +39,12 @@ module.exports = function proxify (obj) {
       if (typeof property === 'string' &&
           config.proxyExcludedKeys.indexOf(property) === -1 &&
           !Reflect.has(target, property)) {
+        // Special message for invalid property access of non-chainable methods.
+        if (nonChainableMethodName) {
+          throw Error('Invalid use of Chai method: ' + nonChainableMethodName +
+            '. Invocation required.');
+        }
```
keithamus: This error message is a bit unfriendly, I feel like it doesn't explicitly point out where the user went wrong. Could we instead have something like `Invalid Chai property equal.pizza. Did you mean equal(pizza)?`

meeber: @keithamus Agreed it can be better. I struggled with the phrasing quite a bit actually. The reason I ended up with the current phrasing is because many of the alternatives I explored only offer sensible advice part of the time.

For example:

```js
expect(x).to.be.equal.to.true;
// Invalid Chai property: equal.to. Did you mean equal(to)?

expect(x).to.be.a.string.with.length(4);
// Invalid Chai property: string.with. Did you mean string(with)?

expect(x).to.throw.an.error;
// Invalid Chai property: throw.an. Did you mean throw(an)?

expect(x).to.be.within.2.and.4;
// Invalid Chai property: within.2. Did you mean within(2)?
```
I feel uncomfortable with Chai reacting to the user's mistake by suggesting an alternative that's also incorrect, especially in the last case since it demonstrates using `within` with only one argument when it actually requires two.

Could do something like: `Invalid Chai property: equal.pizza. See docs for proper usage of the "equal" assertion`.

Or the high-effort solution with customized messages for each method: `Invalid usage of .within. Correct usage: .within(start, finish)`.

keithamus: Okay, I like the idea of "See docs...". I agree that we dont want to present wrong info, but I also what to give users a chance to fix without immediately finding a support channel (like our issue tracker 😵)

meeber:
This one is ready for review. But first, a special shout out to:

- @vieiralucas for making these changes possible via #799.
- @lucasfcosta for not recruiting any additional lucases, thus allowing the Chai team to maintain a balanced lucas-to-non-lucas ratio, thereby ensuring that the world doesn't plunge into chaos.
- @keithamus for being a moose, and a fine one at that.

keithamus:
Okay, this LGTM. Great work @meeber 👍 🎉

vieiralucas:
LGTM

PR #794, opened 2016-09-10
Enabling showDiff for assert's equal and notEqual methods. Closes #790

### LINKED ISSUE #790: Inconsistent between API guide and message from assert.equal
Could you verify why API guide says something different than message from exception ?

Guide:
```
.equal(actual, expected, [message])
```

Example:
```
// test case
assert.equal(result[0].status, 405, 'Response HTTP status equal 405');
// message when assert.equal fire exception
AssertionError: Response HTTP status equal 405: expected 404 to equal 405
```

lucasfcosta:
This aims to close #790.

As explained on [this post](https://github.com/chaijs/chai/issues/790#issuecomment-246078537) the `equal` and `notEqual` methods on the `assert` interface, which called `Assertion.assert` directly, didn't have the `showDiff` flag turned on and that would make their error messages look different because of the way the test runner showed results.

In order to make that behavior consistent with the `equal` related methods on other interfaces, I enabled the `showDiff` flag for these assertions too.

I didn't add any tests since this was a problem related to the test runner and not ours.
Please read the post I linked above for more details.
Thanks everyone 😄

### DIFF
--- a/lib/chai/interface/assert.js
+++ b/lib/chai/interface/assert.js
@@ -129,6 +129,7 @@ module.exports = function (chai, util) {
       , 'expected #{this} to not equal #{act}'
       , exp
       , act
+      , true
     );
   };
 
@@ -156,6 +157,7 @@ module.exports = function (chai, util) {
       , 'expected #{this} to equal #{act}'
       , exp
       , act
+      , true
     );
   };
 


meeber:
LGTM!

lucasfcosta:
Well, actually I was taking a look at #574 and I noticed this would not necessarily be needed, since the new default behavior of `showDiff` on `4.x.x` will be to be turned on if it is not explicitly defined as `false`.

However, I'd like to keep this explicit just to make sure the behavior will still be consistent with the other `equal` methods, since they explicitly set `showDiff` to true.

Let me know if you prefer thrusting on the default behavior instead of adding an explicit `true` value for the `showDiff` flag.

meeber:
I prefer to keep it explicit for reasons of consistency.

keithamus:
Yeah, let's keep it explicit. LGTM

PR #796, opened 2016-09-10
Verify types on increase | decrease

vieiralucas:
Hello there again!

I saw that @meeber did a very good job and now chai 4.x.x is finally merged at master.

So here am I, aiming to solve the rest of #691.
This PR aims to check that the property being tested by increase | decrease is a number.
It also adds tests to `assert.atLeast` and `assert.atMost` since after the merge they automatically started to check types.

I don't know if I'm happy with the documentation changes, so please give your opinions on that.

Thank You

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -1959,7 +1959,7 @@ module.exports = function (chai, _) {
   Assertion.addChainableMethod('changes', assertChanges);
 
   /**
-   * ### .increase(function)
+   * ### .increase(target[, property[, message]])
    *
    * Asserts that a function increases a numeric object property
    *
@@ -1967,11 +1967,18 @@ module.exports = function (chai, _) {
    *     var fn = function() { obj.val = 15 };
    *     expect(fn).to.increase(obj, 'val');
    *
+   * It can also receive a function which returns the property
+   *
+   *     var obj = { val: 10 };
+   *     var fn = function() { obj.val = 5 };
+   *     var getVal = function() { return obj.val };
+   *     expect(fn).to.increase(getVal);
+   *
    * @name increase
    * @alias increases
    * @alias Increase
-   * @param {String} target
-   * @param {String} property name
+   * @param {String|Function} target
+   * @param {String} property name _optional_
    * @param {String} message _optional_
    * @namespace BDD
    * @api public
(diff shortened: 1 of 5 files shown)


meeber:
Round 2!
Fight!

(LGTM!)

lib/chai/core/assertions.js
```diff
@@ -1961,7 +1961,7 @@ module.exports = function (chai, _) {
   /**
    * ### .increase(function)
```
lucasfcosta: Maybe the docs for this could use another name for the argument being passed other than `function`.

lib/chai/core/assertions.js
```diff
@@ -2015,7 +2018,7 @@ module.exports = function (chai, _) {
   /**
    * ### .decrease(function)
```
lucasfcosta: The same applies for this line, what about renaming `function` to `target` on these cases?

lucasfcosta:
Good job @vieiralucas, especially regarding the tests 😄 
This is looking good, but perhaps we could just change the name of the argument on the `increase` and `decrease` assertions. I think `function` does not reflect what can be put there, maybe renaming that argument to `target` would be better.
What do you guys think?

vieiralucas:
Nice one!
I renamed it to target and also added property as the second argument.
Thanks

lucasfcosta:
Hi @vieiralucas, thanks for the corrections, good job noticing the `property` argument too!
I'm sorry to add more comments to this, but I just noticed some things we should fix regarding the docs of these methods, here goes a small list of things to do:
* Since the `property` and `message` arguments are optional they should be represented within brackets, like we do [here](https://github.com/chaijs/chai/blob/master/lib/chai/core/assertions.js#L1143)
* The docs for the types of the arguments also do not say it's possible to pass a function as `target`.
* The docs only talk about a numeric property of an object. They do not make it clear it's also possible to use a `function` as target

Since this PR tackles all those methods maybe we could do all these fixes right here, what do you think?
If you disagree, I'll be happy to accept merging this since the whole code and tests look really good, however, if we could also fix the docs it would be even better.

Also, sorry for not adding comments to lines and for not seeing this before, unfortunately github only allows me to comment lines which have been changed (or are very very near the changes).

vieiralucas:
@lucasfcosta You have very good eyes 😄 

I definitely agree with the things you have said.
Already made the changes.
Please let me know if I missed something.

Thanks again

meeber:
LGTM!

lucasfcosta:
LGTM!
Good job, I'm very happy to merge this 😄

vieiralucas:
\o/

Does that means that #691 can be closed?

lucasfcosta:
Yes it does! Thanks for noticing 👍

PR #797, opened 2016-09-13
Fix #792: Rephrase old docs

### LINKED ISSUE #792: Rephrase old docs and use `target <type>` instead of  `<type> target`
Like Mr. @meeber noticed here and [here](https://github.com/chaijs/chai/pull/783#discussion_r78267204) we've got some doc comments with a phrasing that does not seem good, so a PR fixing all the places where `<type> target` is used instead of `target <type>` would be welcome.
This happens in places like [here, for example](https://github.com/chaijs/chai/blob/4.x.x/lib/chai/core/assertions.js#L507).

nithmanoj:
fixes #792 Rephrase old docs and use `target <type>` instead of `<type> target`

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -622,7 +622,7 @@ module.exports = function (chai, _) {
   /**
    * ### .above(value)
    *
-   * Asserts that the number target is greater than `value`.
+   * Asserts that the target number is greater than `value`.
    *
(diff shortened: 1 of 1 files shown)


keithamus:
Great work, LGTM! 

@lucasfcosta @meeber

lucasfcosta:
Thanks for the PR @nithmanoj :smile: 
Great work!
LGTM too.

PR #799, opened 2016-09-14
Return new assertion with flags copied over instead of this. Fix #791

### LINKED ISSUE #791: Consistency with returning new Assertion instead of this
A change was made a few months ago to return a new Assertion object with flags copied over instead of returning `this`. References: #562 #642 

However, it looks like there are still some spots where `this` is returned. The only one I've actually tested so far is [overwriteProperty](https://github.com/chaijs/chai/blob/e5e6f5cbbd0e3542f928559f2d290663facd82e5/lib/chai/utils/overwriteProperty.js#L61).

For example:

```js
const {expect, use} = require("chai");                                                                                  
                                                                                                                        
use(function (chai, utils) {                                                                                            
  utils.overwriteProperty(chai.Assertion.prototype, "and", function (_super) {                                          
    return function blah () {                                                                                           
      _super.call(this);                                                                                                
    };                                                                                                                  
  });                                                                                                                   
});                                                                                                                     
                                                                                                                        
it("but", () => {                                                                                                      
  expect([1, 2, 3]).to.be.an.but.length.below(1000)    // Passes
});

it("and", () => {                                                                                                      
  expect([1, 2, 3]).to.be.an.and.length.below(1000)    // Fails
});
```

As expected, the second assertion fails with error: "TypeError: expect(...).to.be.an.and.length.below is not a function" because the overwritten `and` is returning the function that was originally returned by the chainable method `an` instead of returning a new Assertion.

Other places where this might be a problem (but I haven't tested):

- [overwriteMethod](https://github.com/chaijs/chai/blob/e5e6f5cbbd0e3542f928559f2d290663facd82e5/lib/chai/utils/overwriteMethod.js#L60)
- [addChainableMethod](https://github.com/chaijs/chai/blob/e5e6f5cbbd0e3542f928559f2d290663facd82e5/lib/chai/utils/addChainableMethod.js#L85)
- [overwriteChainableMethod](https://github.com/chaijs/chai/blob/e5e6f5cbbd0e3542f928559f2d290663facd82e5/lib/chai/utils/overwriteChainableMethod.js#L46-L52)

vieiralucas:
Hello!

This PR addresses #791.

- [x] `overwriteProperty`
- [x] `overwriteMethod`
- [x] `addChainableMethod`
- [x] `overwriteChainableMethod`

### DIFF
--- a/lib/chai/utils/overwriteProperty.js
+++ b/lib/chai/utils/overwriteProperty.js
@@ -4,7 +4,9 @@
  * MIT Licensed
  */
 
+var chai = require('../../chai');
 var flag = require('./flag');
+var transferFlags = require('./transferFlags');
 
 /**
  * ### overwriteProperty (ctx, name, fn)
@@ -58,7 +60,13 @@ module.exports = function (ctx, name, getter) {
         var result = getter(_super).call(this);
         flag(this, 'keep_ssfi', false);
 
-        return result === undefined ? this : result;
+        if (result !== undefined) {
+          return result;
+        }
+
+        var newAssertion = new chai.Assertion();
+        transferFlags(this, newAssertion);
+        return newAssertion;
       }
     , configurable: true
   });
--- a/test/utilities.js
+++ b/test/utilities.js
@@ -849,6 +849,52 @@ describe('utilities', function () {
     })
   });
 
+  it('addChainableMethod should return a new assertion with flags copied over', function () {
+    var assertionConstructor = chai.Assertion;
+
+    chai.use(function(_chai, utils) {
+      assertionConstructor = _chai.Assertion;
+      _chai.Assertion.addChainableMethod('foo', function(str) {
+        utils.flag(this, 'mySpecificFlag', 'value1');
+        utils.flag(this, 'ultraSpecificFlag', 'value2');
+
+        var obj = utils.flag(this, 'object');
+        new _chai.Assertion(obj).to.be.equal(str);
+      });
+
+      _chai.Assertion.addMethod('checkFlags', function() {
+        this.assert(
+            utils.flag(this, 'mySpecificFlag') === 'value1' &&
+            utils.flag(this, 'ultraSpecificFlag') === 'value2'
+          , 'expected assertion to have specific flags'
+          , "this doesn't matter"
+        );
+      });
+    });
+
+    chai.config.proxyExcludedKeys.push('nodeType');
+
+    var assertion1 = expect('bar');
+    var assertion2 = assertion1.foo('bar');
+
+    // Checking if a new assertion was returned
+    expect(assertion1).to.not.be.equal(assertion2);
+
+    // Check if flags were copied
+    assertion2.checkFlags();
+
+    // Checking if it's really an instance of an Assertion
+    expect(assertion2).to.be.instanceOf(assertionConstructor);
+
+    // Test chaining `.length` after a method to guarantee it is not a function's `length`
+    expect('bar').to.be.a.foo('bar').length.above(2);
+
+    // Ensure that foo returns an Assertion (not a function)
+    expect(expect('bar').foo('bar')).to.be.an.instanceOf(assertionConstructor);
+
+    delete chai.Assertion.prototype.foo;
+  });
+
   it('overwriteChainableMethod', function () {
     chai.use(function (_chai, _) {
       _chai.Assertion.overwriteChainableMethod('x',
--- a/test/utilities.js
+++ b/test/utilities.js
@@ -341,22 +341,38 @@ describe('utilities', function () {
   });
 
   describe('overwriteMethod', function () {
+    var assertionConstructor;
+
     before(function() {
       chai.config.includeStack = false;
 
-      chai.use(function(_chai, _) {
+      chai.use(function(_chai, utils) {
+        assertionConstructor = _chai.Assertion;
+
         _chai.Assertion.addMethod('four', function() {
           this.assert(this._obj === 4, 'expected #{this} to be 4', 'expected #{this} to not be 4', 4);
         });
 
         _chai.Assertion.overwriteMethod('four', function(_super) {
           return function() {
+            utils.flag(this, 'mySpecificFlag', 'value1');
+            utils.flag(this, 'ultraSpecificFlag', 'value2');
+
             if (typeof this._obj === 'string') {
               this.assert(this._obj === 'four', 'expected #{this} to be \'four\'', 'expected #{this} to not be \'four\'', 'four');
             } else {
               _super.call(this);
             }
           }
+        });
+
+        _chai.Assertion.addMethod('checkFlags', function() {
+          this.assert(
+              utils.flag(this, 'mySpecificFlag') === 'value1' &&
+              utils.flag(this, 'ultraSpecificFlag') === 'value2'
+            , 'expected assertion to have specific flags'
+            , "this doesn't matter"
+          );
         });
       });
     });
@@ -391,6 +407,26 @@ describe('utilities', function () {
           expect(err.stack).to.not.include('overwriteMethod');
         }
       }
+    });
+
+    it('should return a new assertion with flags copied over', function () {
+      var assertion1 = expect('four');
+      var assertion2 = assertion1.four();
+
+      // Checking if a new assertion was returned
+      expect(assertion1).to.not.be.equal(assertion2);
+
+      // Check if flags were copied
+      assertion2.checkFlags();
+
+      // Checking if it's really an instance of an Assertion
+      expect(assertion2).to.be.instanceOf(assertionConstructor);
+
+      // Test chaining `.length` after a method to guarantee it is not a function's `length`
+      expect('four').to.be.a.four().length.above(2);
+
+      // Ensure that foo returns an Assertion (not a function)
+      expect(expect('four').four()).to.be.an.instanceOf(assertionConstructor);
     });
   });
 
--- a/test/utilities.js
+++ b/test/utilities.js
@@ -209,139 +209,80 @@ describe('utilities', function () {
     });
   });
 
-  it('addMethod', function () {
-    chai.use(function(_chai, utils) {
-      expect(_chai.Assertion).to.not.respondTo('eqqqual');
-      _chai.Assertion.addMethod('eqqqual', function (str) {
-        var object = utils.flag(this, 'object');
-        new _chai.Assertion(object).to.be.eql(str);
-      });
-      expect(_chai.Assertion).to.respondTo('eqqqual');
-    });
-
-    expect('spec').to.eqqqual('spec');
-  });
-
-  it('addMethod returning result', function () {
-    chai.use(function(_chai, utils) {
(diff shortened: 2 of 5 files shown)


lib/chai/utils/overwriteProperty.js
```diff
@@ -4,7 +4,9 @@
  * MIT Licensed
  */
 
+var chai = require('../../chai');
```
keithamus: Does this create a circular dependency? Have we done this before? I feel like this might create issues

lib/chai/utils/overwriteProperty.js
```diff
@@ -58,7 +60,12 @@ module.exports = function (ctx, name, getter) {
         var result = getter(_super).call(this);
         flag(this, 'keep_ssfi', false);
 
-        return result === undefined ? this : result;
+        if (result !== undefined)
+          return result;
```
keithamus: Let's put braces around this.

lib/chai/utils/overwriteProperty.js
lucasfcosta: Currently `addMethod` and `addProperty` have this too and there's been no issue reported about it. However, I'd like to have more time to check this before giving a confident response.

meeber: I'm a bit confused how it's working.

1. `test/bootstrap/index.js` requires `index.js`
1. `index.js` requires `lib/chai.js`
1. `lib/chai.js` requires `lib/chai/utils/index.js`
1. `lib/chai/utils/index.js` requires `lib/chai/utils/addProperty.js`
1. `lib/chai/utils/addProperty.js` requires `lib/chai.js`

Thus circular dependency. -scratches head-

test/utilities.js
```diff
@@ -543,6 +543,53 @@ describe('utilities', function () {
...
       }
     });
+
+    it('should return new assertion with flags copied over', function() {
+      var assertionConstructor;
+
+      chai.use(function (_chai, utils) {
+        assertionConstructor = _chai.Assertion;
+
+        _chai.Assertion.overwriteProperty('and', function (_super) {
```
meeber: I think we need to create a custom property to overwrite instead of using `and` since this change will bleed over into subsequent tests.

lib/chai/utils/overwriteProperty.js
keithamus: I think it works because chai.js only sets `module.exports` and this file only calls stuff from chai.js inside of the methods, so it all kind of works out eventually.

vieiralucas:
Did not managed to reproduce the same problem for `overwriteMethod`.

For `addChainableMethod`:

```javascript
chai.use(function(_chai, utils) {
  _chai.Assertion.addChainableMethod('len', function (ln) {
    var obj = utils.flag(this, 'obj');
    this.assert(obj.length === ln);
  });
});

expect('string').to.len(6).length.below(1000); // Passess

expect('string').to.be.len.and.length.below(1000); // Passes

expect('string').to.be.len.length.below(1000); // Fails with `to.len.length is not a function`
```

The problem is getting length right after the new chainable method is chainned.
This is happening because it is currently returning the length of [this](https://github.com/chaijs/chai/blob/3ef86fdbf62343ef345873c5514cd9e70af36b22/lib/chai/utils/addChainableMethod.js#L80) function.

I was not able to fix this one :(

vieiralucas:
I believe that I was creating more confusion than explaining something (at least to myself).

IMHO, we should return a new Assertion with flags copied over, not only to prevent bugs with `length`, but to keep consistency as well.

test/utilities.js
```diff
@@ -230,6 +230,8 @@ describe('utilities', function () {
     });
 
     expect(expect('foo').result()).to.equal('result');
+
+    delete chai.Assertion.prototype.result;
```
meeber: It's a great idea to add this so the test cleans up after itself, but we need it to run the cleanup operation regardless of success or failure. Therefore, I think we should wrap this test in a `describe` with an `after` function to contain the cleanup activities.

test/utilities.js
```diff
@@ -273,6 +275,9 @@ describe('utilities', function () {
 
     var anotherAssertion = expect([1, 2, 3]).to.have.a.lengthOf(3).and.to.be.ok;
     expect(anotherAssertion.length.constructor).to.equal(assertionConstructor);
+
+    delete chai.Assertion.prototype.returnNewAssertion;
+    delete chai.Assertion.prototype.checkFlags;
```
meeber: Same thing here; cleanup steps should be in an `after` function.

test/utilities.js
```diff
@@ -403,6 +444,8 @@ describe('utilities', function () {
 
     var assert = expect('chai').to.be.tea;
     expect(assert.__flags.tea).to.equal('chai');
+
+    delete chai.Assertion.prototype.tea;
```
meeber: Same thing here; cleanup steps should be in an `after` function.

test/utilities.js
```diff
@@ -446,6 +489,9 @@ describe('utilities', function () {
 
     expect(expect([1, 2, 3]).be).to.be.an.instanceOf(assertionConstructor);
     expect(expect([1, 2, 3]).thing).to.be.an.instanceOf(assertionConstructor);
+
+    delete chai.Assertion.prototype.thing;
+    delete chai.Assertion.prototype.checkFlags;
```
meeber: Same thing here; cleanup steps should be in an `after` function.

test/utilities.js
```diff
@@ -456,11 +502,16 @@ describe('utilities', function () {
     });
 
     expect(expect('foo').result).to.equal('result');
+
+    delete chai.Assertion.prototype.result;
```
meeber: Same thing here; cleanup steps should be in an `after` function.

test/utilities.js
```diff
@@ -477,6 +528,8 @@ describe('utilities', function () {
     expect(matcha.__flags.tea).to.equal('matcha');
     var assert = expect('something').to.be.tea;
     expect(assert.__flags.tea).to.equal('chai');
+
+    delete chai.Assertion.prototype.tea;
```
meeber: Same thing here; cleanup steps should be in an `after` function.

test/utilities.js
```diff
@@ -489,6 +542,8 @@ describe('utilities', function () {
     });
 
     expect(expect('foo').result).to.equal('result');
+
+    delete chai.Assertion.prototype.result;
```
meeber: Same thing here; cleanup steps should be in an `after` function.

test/utilities.js
```diff
@@ -762,46 +866,156 @@ describe('utilities', function () {
       expect(obj).x.to.be.ok;
       expect(obj).to.have.property('__x', 'X!');
     })
+
+    delete chai.Assertion.prototype.x;
```
meeber: Same thing here; cleanup steps should be in an `after` function.

test/utilities.js
```diff
@@ -762,46 +866,156 @@ describe('utilities', function () {
...
+    expect(assertion2).to.be.instanceOf(assertionConstructor);
+
+    // Test chaining `.length` after a method to guarantee it is not a function's `length`
+    expect('bar').to.be.a.foo('bar').length.above(2);
+
+    // Ensure that foo returns an Assertion (not a function)
+    expect(expect('bar').foo('bar')).to.be.an.instanceOf(assertionConstructor);
+
+    delete chai.Assertion.prototype.foo;
+    delete chai.Assertion.prototype.checkFlags;
```
meeber: Same thing here; cleanup steps should be in an `after` function.

test/utilities.js
vieiralucas: Of course, you're absolutely right. Thanks for noticing.
I'll make the changes.

test/utilities.js
vieiralucas: Done

test/utilities.js
vieiralucas: Done

test/utilities.js
vieiralucas: Done

test/utilities.js
vieiralucas: Done

test/utilities.js
vieiralucas: Done

test/utilities.js
vieiralucas: Done

test/utilities.js
vieiralucas: Done

test/utilities.js
vieiralucas: Done

test/utilities.js
vieiralucas: Done

meeber:
@vieiralucas Thanks for the extra effort refactoring the tests to clean up after themselves. Personally, I think the whole test suite could use a refactoring to be better organized and labeled, and only have one assertion per test, but that's a future battle :D

Anyway, this LGTM! @keithamus @lucasfcosta?

lucasfcosta:
@meeber I totally agree with your considerations about the LGTM feature, it would be perfect if we could require more than one review for PR, however, if this is not possible, can we disable this and continue using only LGTM?

Also, this LGTM!

keithamus:
@meeber @lucasfcosta I enabled the github review to see how well it worked. I've disabled it for now, and we can keep using LGTM. When it's configurable for N number of reviewers then we'll look at re-enabling I think 😄

meeber:
Based on my limited interactions with it, I think the builtin review tool has a lot of potential. I particularly like being able to group a bunch of comments together before submitting them during a review.

keithamus:
Yes, I think ultimately it's better than LGTM (for our use case) but also much less configurable - namely the number of required reviewers, which becomes a deal breaker (unless we want to lower the number of required viewers to 1 😆).

lucasfcosta:
@vieiralucas @meeber @keithamus Apparently this makes the build fail, I have no idea how this is possible since the Travis check passed.
I'll try to run tests for this and fix things as soon as I get home.

Any ideas about how this happened?

keithamus:
Ahh sadly they're not - because the saucelabs tests only run on master (because if they were to run on pull requests, we'd risk exposing our saucelabs tokens to malicious PR code).

vieiralucas:
Not very familiarized with saucelabs, but I found this


IE 10.0.0 (Windows 7 0.0.0) utilities overwriteChainableMethod should return a new assertion with flags copied over FAILED

keithamus:
You should try to run the tests in IE10 locally, see why it is failing.

meeber:
I can't look at it closely right now but my quick guess is it's because IE 10.0 takes a different code path [here](https://github.com/chaijs/chai/blob/7683356fff853b0f661ec1fbdd69979c6f31aab2/lib/chai/utils/addChainableMethod.js#L96-L113), causing [this test](https://github.com/chaijs/chai/blob/7683356fff853b0f661ec1fbdd69979c6f31aab2/test/utilities.js#L1029) to fail.

vieiralucas:
Currently downloading IE10 VM.
26 min left...

vieiralucas:
Well...
Now I know how to run tests on IE. 😸 

WHY IS THAT SO HARD MICROSOFT PLS

PR #802, opened 2016-09-15
Fix bug when trying to call inspect on an Assertion

lucasfcosta:
Hello everyone!
As me and @vieiralucas were taking a look at Chai's code yesterday on college, we noticed that there was a bug when an assertion which had other `Assertion` objects as `expected` or `actual` failed.

This was happening because mocha called the `.inspect` method on them and it tried to read the property `nodeType`, which did not exist on the `Assertion` object and ended up throwing an error.

To fix this I simply added a check before accessing that property, calling `hasOwnProperty` and passing the `object` we're inspecting (as `this`) and `'nodeType'`.

### DIFF
--- a/lib/chai/utils/inspect.js
+++ b/lib/chai/utils/inspect.js
@@ -37,6 +37,7 @@ var isDOMElement = function (object) {
   } else {
     return object &&
       typeof object === 'object' &&
+      'nodeType' in object &&
       object.nodeType === 1 &&
       typeof object.nodeName === 'string';
   }
--- a/test/utilities.js
+++ b/test/utilities.js
@@ -819,6 +819,17 @@ describe('utilities', function () {
     });
   });
 
+  it('inspect an assertion', function () {
+    chai.use(function (_chai, _) {
+      var assertion = expect(1);
+      var anInspectFn = function() {
+        return _.inspect(assertion);
+      };
+
+      expect(anInspectFn).to.not.throw();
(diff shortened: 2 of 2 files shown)


vieiralucas:
LGTM

lib/chai/utils/inspect.js
```diff
@@ -37,6 +37,7 @@ var isDOMElement = function (object) {
   } else {
     return object &&
       typeof object === 'object' &&
+      Object.hasOwnProperty.call(object, 'nodeType') &&
```
meeber: I wonder if it'd be safer to do a `typeof` on line 40, just in the off chance that an object is inheriting `nodeType` from its prototype chain.

lucasfcosta: This makes sense, but I will need to check if when using `typeof object.nodeType` it tries to access the property and therefore ends up throwing the proxy error we are trying to avoid.

I totally agree with you, we should support inheriting it from the prototype chain, but I'm not sure `typeof` would be the optimal solution.

Thanks for the review :smile:

meeber: Good point, maybe an `in` check is best then.

lucasfcosta:
Done!
Thanks for noticing the problem regarding having `nodeType` up the `prototype` chain 👍

keithamus:
LGTM then 👍

PR #804, opened 2016-09-19
Fix breaking test on IE10

vieiralucas:
As @meeber said [here](https://github.com/chaijs/chai/pull/799#issuecomment-248042985)
Since IE10 does not support `__proto__`, [this](https://github.com/chaijs/chai/blob/7683356fff853b0f661ec1fbdd69979c6f31aab2/test/utilities.js#L1029) test was failing.


meeber:
LGTM!

lucasfcosta:
This makes sense, so if we're not changing the object prototype but only redefining properties on the Function object, then this check will always fail.

Great catch @meeber!

PR #629, opened 2016-02-25
Fixed JsDoc spelling of Object.freeze() in .isNotExtensible() assertion.

kidmillions:

PR #805, opened 2016-09-26
Fix typo in the docstring of isNotExtensible

remarkablemark:
Fix `Object.freeze` misspelling.

See also https://github.com/chaijs/chai-docs/pull/140

### DIFF
--- a/chai.js
+++ b/chai.js
@@ -3690,7 +3690,7 @@ module.exports = function (chai, util) {
    *
    *     var nonExtensibleObject = Object.preventExtensions({});
    *     var sealedObject = Object.seal({});
-   *     var frozenObject = Object.freese({});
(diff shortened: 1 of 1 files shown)


keithamus:
Thanks @remarkablemark! LGTM

@meeber @lucasfcosta @vieiralucas ➡️

meeber:
@remarkablemark Thanks for the PR. This was actually fixed via #629, but the change doesn't make its way into the `Chai.js` file until it's regenerated as part of a new version release. We plan on releasing 4.0 soon at which point the change will appear.

keithamus:
Ahh, didn't notice that this was on the chai.js file. Good job we have a 2-review system!

remarkablemark:
Thanks @meeber for the clarification and thanks @keithamus for the review!

PR #810, opened 2016-09-28
Refactor `ownProperty`

### LINKED ISSUE #795: Refactor `.ownProperty` to leverage `.property`
There's a divergence between the behavior of the `.property` and `.ownProperty` assertions due to refactoring of `.property` in the master branch and `.ownProperty` in the 4.x.x branch.

I think the thing to do here is:

- Create a new `own` flag.
- Update `.property` to restrict checks to own properties when the  `own` flag is set.
- Change `.ownProperty` and `.haveOwnProperty` into simple assertions that set the `own` flag and then call `.property` to do the rest.

That way, the code paths for `.property` and `.ownProperty` are unified and will never diverge again. It also enables chaining with `.deep`. For example: 

```js
expect(obj).to.have.deep.own.property('kittens', {quantity: 3});
```

The only breaking change to `.ownProperty` would be the same breaking change that happened to `.property` during the refactor in #744.

However, I don't think the `nested` and `own` flags mix conceptually, since the intent of `nested` is to test a property that's multiple levels deep, whereas the purpose of `own` is to test a property that is set directly on an object as opposed to being inherited. Therefore, combining these two flags would cause `.property` to throw an error.

I already started playing around with this on the plane. I'll go ahead and finish it if there's consensus.

@lucasfcosta @keithamus?

meeber:
This PR is split into two commits. The first commit addresses a small issue with the failed assertion error message being inconsistent across property-related assertions. Sometimes the message was "expected blah to have a property..." and other times it was "expected blah to have property". I settled on the second style because otherwise the logic gets more complex having to deal with changing between "a" or "an" depending on if the next word in the message starts with a vowel (such as "own").

The second commit makes the following changes:
- Add `own` flag
- Make `ownProperty` and `haveOwnProperty` aliases of `own.property`
- Add `deep` support to `own.property`
- Add `assert.deepOwnPropertyVal`
- Add `assert.notDeepOwnPropertyVal`

Closes #795.

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -1062,15 +1062,15 @@ module.exports = function (chai, _) {
     if (!negate || arguments.length === 1) {
       this.assert(
           hasProperty
-        , 'expected #{this} to have a ' + descriptor + _.inspect(name)
+        , 'expected #{this} to have ' + descriptor + _.inspect(name)
         , 'expected #{this} to not have ' + descriptor + _.inspect(name));
     }
 
     if (arguments.length > 1) {
       this.assert(
           hasProperty && (isDeep ? _.eql(val, value) : val === value)
-        , 'expected #{this} to have a ' + descriptor + _.inspect(name) + ' of #{exp}, but got #{act}'
-        , 'expected #{this} to not have a ' + descriptor + _.inspect(name) + ' of #{act}'
+        , 'expected #{this} to have ' + descriptor + _.inspect(name) + ' of #{exp}, but got #{act}'
+        , 'expected #{this} to not have ' + descriptor + _.inspect(name) + ' of #{act}'
         , val
         , value
       );
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -104,6 +104,8 @@ module.exports = function (chai, _) {
    *     var deepCss = { '.link': { '[target]': 42 }};
    *     expect(deepCss).to.have.nested.property('\\.link.\\[target\\]', 42);
    *
+   * The `nested` flag cannot be combined with the `own` flag.
+   *
    * @name nested
    * @namespace BDD
    * @api public
@@ -111,6 +113,25 @@ module.exports = function (chai, _) {
 
   Assertion.addProperty('nested', function () {
     flag(this, 'nested', true);
+  });
+
+  /**
+   * ### .own
+   *
+   * Sets the `own` flag, later used by the `property` assertion.
+   *
+   *     expect({a: 1}).to.have.own.property('a');
+   *     expect({a: 1}).to.have.own.property('a', 1);
+   *
+   * The `own` flag cannot be combined with the `nested` flag.
+   *
+   * @name own
+   * @namespace BDD
+   * @api public
+   */
+
+  Assertion.addProperty('own', function () {
+    flag(this, 'own', true);
   });
 
   /**
@@ -954,12 +975,14 @@ module.exports = function (chai, _) {
   /**
    * ### .property(name, [value])
    *
-   * Asserts that the target has a property `name`, optionally asserting that
-   * the value of that property is strictly equal to `value`.
+   * Asserts that the target has a direct or inherited property `name`,
+   * optionally asserting that the value of that property is strictly equal to
+   * `value`.
    *
    *     var obj = { foo: 'bar' };
    *     expect(obj).to.have.property('foo');
    *     expect(obj).to.have.property('foo', 'bar');
+   *     expect(obj).to.have.property('toString');
    *     expect(obj).to.not.have.property('baz');
    *     expect(obj).to.not.have.property('foo', 'baz');
    *     expect(obj).to.not.have.property('baz', 'bar');
@@ -971,6 +994,14 @@ module.exports = function (chai, _) {
    *     expect(obj).to.have.deep.property('foo', { bar: 'baz' });
    *     expect(obj).to.not.have.deep.property('foo', { bar: 'quux' });
    *
+   * If the `own` flag is set, the property must exist directly on the object.
+   * Inherited properties aren't checked.
+   *
+   *     var obj = { foo: 'bar' };
+   *     expect(obj).to.have.own.property('foo');
+   *     expect(obj).to.have.own.property('foo', 'bar');
+   *     expect(obj).to.not.have.own.property('toString');
+   *
    * If the `nested` flag is set, you can use dot- and bracket-notation for
    * nested references into objects and arrays.
    *
@@ -981,11 +1012,6 @@ module.exports = function (chai, _) {
    *     expect(deepObj).to.have.nested.property('green.tea', 'matcha');
    *     expect(deepObj).to.have.nested.property('teas[1]', 'matcha');
    *     expect(deepObj).to.have.nested.property('teas[2].tea', 'konacha');
-   *
-   * The `deep` and `nested` flags can be combined.
-   *
-   *    expect({ foo: { bar: { baz: 'quux' } } })
-   *      .to.have.deep.nested.property('foo.bar', { baz: 'quux' });
    *
    * You can also use an array as the starting point of a `nested.property`
    * assertion, or traverse nested arrays.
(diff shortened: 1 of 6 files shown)


vieiralucas:
So, chai will only throw if `own` is used alongside `nested` flags when a property assertion is made.

Which means that there is "nothing wrong" with this:
```javascript
expect('test').to.have.own.nested.length(4);
```

I'm ok with that.

@meeber, this is awesome work. I can not say anything but LGTM

lib/chai/core/assertions.js
```diff
@@ -1036,23 +1072,28 @@ module.exports = function (chai, _) {
...
       , negate = flag(this, 'negate')
       , obj = flag(this, 'object')
       , pathInfo = isNested ? _.getPathInfo(name, obj) : null
-      , hasProperty = isNested
-        ? pathInfo.exists
-        : _.hasProperty(name, obj)
-      , value = isNested
-        ? pathInfo.value
-        : obj[name];
+      , hasProperty = isOwn ? Object.prototype.hasOwnProperty.call(obj, name)
```
lucasfcosta: I'm not sure I like all these ternary operators.
What do you guys think about using `if` and `else` clauses?
This is totally understandable, but making this behavior more clear might make these conditions look better and more easily understandable.

lib/chai/interface/assert.js
```diff
@@ -1213,173 +1178,260 @@ module.exports = function (chai, util) {
   };
 
   /**
-   * ### .nestedPropertyVal(object, property, value, [message])
+   * ### .ownProperty(object, property, [message])
    *
-   * Asserts that `object` has a property named by `property` with value given
-   * by `value`. `property` can use dot- and bracket-notation for nested
-   * reference. Uses a strict equality check (===).
+   * Asserts that `object` has a direct property named by `property`.
```
lucasfcosta: I think that only saying "direct property" isn't enough to make this explanation clear, maybe we could add `(non-inherited)` after the word `direct` just to make sure we're being clear.

I don't think this is needed in the other methods where we mention inherited properties, since that contrast already explains everything by itself.

lib/chai/interface/assert.js
```diff
@@ -1213,173 +1178,260 @@ module.exports = function (chai, util) {
...
+   */
+
+  assert.ownProperty = function (obj, prop, msg) {
+    new Assertion(obj, msg).to.have.own.property(prop);
+  };
+
+  /**
+   * ### .notOwnProperty(object, property, [message])
+   *
+   * Asserts that `object` does _not_ have a direct property named by `property`.
```
lucasfcosta: Same here as on L1183.

lib/chai/interface/assert.js
```diff
@@ -1213,173 +1178,260 @@ module.exports = function (chai, util) {
...
+   */
+
+  assert.notOwnProperty = function (obj, prop, msg) {
+    new Assertion(obj, msg).to.not.have.own.property(prop);
+  };
+
+  /**
+   * ### .ownPropertyVal(object, property, value, [message])
+   *
+   * Asserts that `object` has a direct property named by `property` and a value
```
lucasfcosta: Same here as on L1183.

lib/chai/interface/assert.js
```diff
@@ -1213,173 +1178,260 @@ module.exports = function (chai, util) {
...
   };
 
   /**
-   * ### .notNestedPropertyVal(object, property, value, [message])
+   * ### .notOwnPropertyVal(object, property, value, [message])
    *
-   * Asserts that `object` does _not_ have a property named by `property` with
-   * value given by `value`. `property` can use dot- and bracket-notation for
-   * nested reference. Uses a strict equality check (===).
+   * Asserts that `object` does _not_ have a direct property named by `property`
```
lucasfcosta: Same here as on L1183.

lib/chai/interface/assert.js
```diff
@@ -1213,173 +1178,260 @@ module.exports = function (chai, util) {
...
   };
 
   /**
-   * ### .deepNestedPropertyVal(object, property, value, [message])
+   * ### .deepOwnPropertyVal(object, property, value, [message])
    *
-   * Asserts that `object` has a property named by `property` with a value given
-   * by `value`. `property` can use dot- and bracket-notation for nested
-   * reference. Uses a deep equality check.
+   * Asserts that `object` has a direct property named by `property` and a value
```
lucasfcosta: Same here as on L1183.

lib/chai/interface/assert.js
```diff
@@ -1213,173 +1178,260 @@ module.exports = function (chai, util) {
...
   };
 
   /**
-   * ### .notDeepNestedPropertyVal(object, property, value, [message])
+   * ### .notDeepOwnPropertyVal(object, property, value, [message])
    *
-   * Asserts that `object` does _not_ have a property named by `property` with
-   * value given by `value`. `property` can use dot- and bracket-notation for
-   * nested reference. Uses a deep equality check.
+   * Asserts that `object` does _not_ have a direct property named by `property`
```
lucasfcosta: Same here as on L1183.

test/expect.js
```diff
@@ -790,34 +791,17 @@ describe('expect', function () {
...
-      .to.have.nested.property('foo\\.bar\\[\\]');
-
-    err(function(){
-      expect({ 'foo.bar': 'baz' })
-        .to.have.nested.property('foo.bar');
-    }, "expected { 'foo.bar': 'baz' } to have a nested property 'foo.bar'");
+    }, "expected { foo: { bar: 'baz' } } to have property 'foo.bar'");
   });
 
   it('property(name, val)', function(){
```
lucasfcosta: IMO it would be good to have tests to see if we're really unable to use both `own` and `nested` together.

test/should.js
```diff
@@ -688,44 +688,33 @@ describe('should', function() {
...
-
-    ({ 'foo.bar[]': 'baz'}).should.have.nested.property('foo\\.bar\\[\\]');
-
-    err(function(){
-      ({ 'foo.bar': 'baz' }).should.have.nested.property('foo.bar');
-    }, "expected { 'foo.bar': 'baz' } to have a nested property 'foo.bar'");
+    }, "expected 'asd' to have property 'foo'");
   });
 
   it('property(name, val)', function(){
```
lucasfcosta: IMO it would be good to have tests to see if we're really unable to use both `own` and `nested` together.

lib/chai/core/assertions.js
meeber: But all these nested ternaries like you :(

I'll change :(

lib/chai/interface/assert.js
meeber: Agreed. I actually had something to the effect of "but not inherited" in a previous iteration, but ended up removing it because the way I had it worded made it sound like the assertion would fail if the object had an inherited property with the given name, whereas it's allowed to have an inherited property with the given name as long as it also has a direct property with the same name. I should be able to reword it to not sound ambiguous though.

test/expect.js
meeber: Agreed!

meeber:
Pushed new version with better comments, more tests, and fewer glorious ternaries.

lib/chai/core/assertions.js
lucasfcosta: Hahahahaha, I see no problem in keeping them there, it's just a matter of personal taste, would you like to keep them?
If you do, please don't be afraid to say it, there's no problem if you prefer that way 😄

meeber: No, I'd rather be sad :(

Plus I checked @keithamus's eslint config and he doesn't like nested ternaries either :(

lucasfcosta:
LGTM!
Awesome work @meeber, I'll leave this here for you to merge if you want to use the ternaries we had before 😄 
If you are happy with the current version of this please feel free to hit the merge button!

meeber:
@lucasfcosta Thanks for reviewing! Good catches :D

I'll leave this open another day for @keithamus and/or @vieiralucas to look at since I made some changes.

I like it fine without the fabulous ternaries, just giving you a hard time :D

Issue #737, opened 2016-06-24
Move pathval into its own module


keithamus:
We're currently not using `pathval` and the code on the [pathval github repo](https://github.com/chaijs/pathval) is quite old. It would be good to make the effort to move the pathval code which resided in chai ([here](https://github.com/chaijs/chai/blob/master/lib/chai/utils/getPathValue.js), [here](https://github.com/chaijs/chai/blob/9dcd0eb6b0a0cbf75f3554d663cbe7a12256da38/lib/chai/utils/getPathInfo.js) and [here](https://github.com/chaijs/chai/blob/9dcd0eb6b0a0cbf75f3554d663cbe7a12256da38/lib/chai/utils/hasProperty.js)) into the pathval module.

Pathval module itself should also be updated to the same standards `type-detect` and `check-error` are - so using the same build, lint, code coverage, and browser testing tools.

lucasfcosta:
Well, if there isn't anyone doing this I'd be happy to help with this issue. Since I've done the `check-error` repo I think I'm pretty familiar with the whole process.

Would you mind if I do so, @keithamus and @meeber?

keithamus:
I'm happy if @meeber's happy 😄

meeber:
Have at it!

lucasfcosta:
Thanks guys, I'm going to have a lot of fun this weekend :smile:

lucasfcosta:
Hello folks, I'd like to update you on the status of this issue and also ask some questions.
Currently I've moved the whole code to the `pathval` repo and the project structure is also done. All that's left is the tests to get 100% coverage, and the rest depends on your answers to the questions below:

1. Should I refactor `pathval`? I think it would be a nice thing to do, but I'm not sure this fits into the scope since it can generate new bugs or unexpected behavior. Should I do it or not?
2. I noticed that here we don't use `pathval`'s`set` method, but it seems like an important method for `pathval` as a standalone method, so I thought it would be nice to keep it, what do you think about it?
3. Can I tweak `.eslintrc` rules a little bit? Some rules are too strict, such as making initializing a variable mandatory (even though I want it to really be `undefined`) and not allowing only `if/else` branches inside an else block. Currently there is no elegant way to make `pathval` work with these rules, so instead of using `eslint-disable-line` I would prefer disabling these rules at all.

Well, I think that's all, thanks for you attention :smile:

keithamus:
[...]

PR #812, opened 2016-09-30
Make 'empty' throw on non-string primitives and functions

shvaikalesh:
Follow-up of #763.

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -541,7 +541,7 @@ module.exports = function (chai, _) {
         throw new TypeError('.empty was passed a function' + name);
       default:
         if (val !== Object(val)) {
-          throw new TypeError('.empty was passed non-string primitive ' + String(val));
+          throw new TypeError('.empty was passed non-string primitive ' + (val && val.toString()));
         }
         itemsCount = Object.keys(val).length;
     }
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -528,10 +528,25 @@ module.exports = function (chai, _) {
    */
 
   Assertion.addProperty('empty', function () {
-    var obj = flag(this, 'object');
-    new Assertion(obj).to.exist;
+    var val = flag(this, 'object')
+      , itemsCount;
+
+    switch (_.type(val)) {
+      case 'array':
+      case 'string':
+        itemsCount = val.length;
+        break;
+      case 'function':
+        throw new TypeError('.empty() was passed a function');
+      default:
+        if (val !== Object(val)) {
+          throw new TypeError('.empty() was passed non-string primitive');
+        }
+        itemsCount = Object.keys(val).length;
+    }
+
     this.assert(
-        Object.keys(Object(obj)).length === 0
+        0 === itemsCount
       , 'expected #{this} to be empty'
       , 'expected #{this} not to be empty'
     );
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -537,10 +537,11 @@ module.exports = function (chai, _) {
         itemsCount = val.length;
         break;
       case 'function':
-        throw new TypeError('.empty() was passed a function');
+        var name = val.name ? ' ' + val.name : '';
+        throw new TypeError('.empty was passed a function' + name);
       default:
         if (val !== Object(val)) {
-          throw new TypeError('.empty() was passed non-string primitive');
+          throw new TypeError('.empty was passed non-string primitive ' + String(val));
         }
         itemsCount = Object.keys(val).length;
     }
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -515,8 +515,8 @@ module.exports = function (chai, _) {
    * ### .empty
    *
    * Asserts that the target's length is `0`. For arrays and strings, it checks
-   * the `length` property. For objects, it gets the count of
-   * enumerable keys.
+   * the `length` property. For non-function objects, it gets the count of own
+   * enumerable string keys.
    *
    *     expect([]).to.be.empty;
    *     expect('').to.be.empty;
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -541,7 +541,7 @@ module.exports = function (chai, _) {
         throw new TypeError('.empty was passed a function' + name);
       default:
         if (val !== Object(val)) {
-          throw new TypeError('.empty was passed non-string primitive ' + (val && val.toString()));
+          throw new TypeError('.empty was passed non-string primitive ' + _.inspect(val));
         }
         itemsCount = Object.keys(val).length;
     }
--- a/test/expect.js
+++ b/test/expect.js
@@ -680,28 +680,53 @@ describe('expect', function () {
 
     err(function(){
       expect(null).to.be.empty;
-    }, "expected null to exist");
+    }, ".empty() was passed non-string primitive");
 
     err(function(){
       expect(undefined).to.be.empty;
-    }, "expected undefined to exist");
+    }, ".empty() was passed non-string primitive");
 
     err(function(){
       expect().to.be.empty;
-    }, "expected undefined to exist");
+    }, ".empty() was passed non-string primitive");
 
     err(function(){
       expect(null).to.not.be.empty;
-    }, "expected null to exist");
+    }, ".empty() was passed non-string primitive");
 
     err(function(){
       expect(undefined).to.not.be.empty;
-    }, "expected undefined to exist");
+    }, ".empty() was passed non-string primitive");
 
     err(function(){
       expect().to.not.be.empty;
-    }, "expected undefined to exist");
+    }, ".empty() was passed non-string primitive");
 
+    err(function(){
+      expect(0).to.be.empty;
+    }, ".empty() was passed non-string primitive");
+
+    err(function(){
+      expect(1).to.be.empty;
+    }, ".empty() was passed non-string primitive");
+
+    err(function(){
+      expect(true).to.be.empty;
+    }, ".empty() was passed non-string primitive");
+
+    err(function(){
+      expect(false).to.be.empty;
+    }, ".empty() was passed non-string primitive");
+
+    if (typeof Symbol !== 'undefined') {
+      err(function(){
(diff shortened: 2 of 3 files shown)


lib/chai/core/assertions.js
```diff
@@ -528,10 +528,25 @@ module.exports = function (chai, _) {
...
+      default:
+        if (val !== Object(val)) {
+          throw new TypeError('.empty() was passed non-string primitive');
+        }
+        itemsCount = Object.keys(val).length;
+    }
+
     this.assert(
-        Object.keys(Object(obj)).length === 0
+        0 === itemsCount
```
shvaikalesh: Is there ESLint config we can add to the project? This will greatly improve contribution/review experience: yoda comparisons and comma-first are not quite popular and it would be awesome to automate codestyle checks.

lib/chai/core/assertions.js
```diff
@@ -528,10 +528,25 @@ module.exports = function (chai, _) {
...
+
+    switch (_.type(val)) {
+      case 'array':
+      case 'string':
+        itemsCount = val.length;
+        break;
+      case 'function':
+        throw new TypeError('.empty() was passed a function');
+      default:
+        if (val !== Object(val)) {
```
shvaikalesh: Other option is to add `case` clause for other primitives' types.

lib/chai/core/assertions.js
```diff
@@ -528,10 +528,25 @@ module.exports = function (chai, _) {
...
-    new Assertion(obj).to.exist;
+    var val = flag(this, 'object')
+      , itemsCount;
+
+    switch (_.type(val)) {
+      case 'array':
+      case 'string':
+        itemsCount = val.length;
+        break;
+      case 'function':
```
shvaikalesh: This should handle generators too (because of recent updates to `type-detect`) and `async` function in the future.

lib/chai/core/assertions.js
```diff
@@ -528,10 +528,25 @@ module.exports = function (chai, _) {
...
 
   Assertion.addProperty('empty', function () {
-    var obj = flag(this, 'object');
-    new Assertion(obj).to.exist;
+    var val = flag(this, 'object')
+      , itemsCount;
+
+    switch (_.type(val)) {
+      case 'array':
+      case 'string':
```
shvaikalesh: Works for strings, wrapped in objects too.

meeber:
@shvaikalesh Thank you for your continued work on this! Implementation looks fine to me. However, I think we should update the jsdoc to clarify that only non-function objects are allowed.

Other thoughts:

- Agreed on ESLint. I'll start a new discussion on this after v4 is released.
- You mentioned previously that you'll be following up this PR with another one that adds support for `Map` and `Set`. At that time, I'm wondering if we should begin throwing a `TypeError` if a `WeakMap` or `WeakSet` is passed. It seems to me that it'd likely be a programmer error to attempt to check if one of those two were empty without realizing that you can't do that.
- It looks like we'll need a 3rd PR to add `empty` to the `assert` interface. Lemme know if you're interested in tackling that as well. Otherwise, I'll open a new issue for it.

shvaikalesh:
@meeber I have updated jsdoc. Throwing on weak collections is nice idea, absolutely agree. Totally interested to add `empty` to `assert` interface.

One more thought: it seems like tests for most assertions are repeated for every interface they are available for. Would be great to have only one test suite for each assertion.

meeber:
@shvaikalesh Yeah adding tests is painful right now due to the multiple interfaces. There was a discussion about this somewhere but I can't find it. Anyway, the long term idea as we move into a modular system as described in #457 is to only have to write tests once for each assertion. But we're not quite there yet.

Regarding this PR: LGTM!

lib/chai/core/assertions.js
```diff
@@ -528,10 +528,25 @@ module.exports = function (chai, _) {
...
+        itemsCount = val.length;
+        break;
+      case 'function':
+        throw new TypeError('.empty() was passed a function');
```
vieiralucas: What about replacing `.empty()` with `.empty` since it is a property assertion?

lib/chai/core/assertions.js
```diff
@@ -528,10 +528,25 @@ module.exports = function (chai, _) {
...
+      default:
+        if (val !== Object(val)) {
+          throw new TypeError('.empty() was passed non-string primitive');
```
vieiralucas: What about replacing `.empty()` with `.empty` since it is a property assertion?

lib/chai/core/assertions.js
lucasfcosta: Great use of `Object()`. 😍

lib/chai/core/assertions.js
lucasfcosta: @shvaikalesh Yes, we thought about that too, we already discussed that [here](https://github.com/chaijs/chai/issues/737#issuecomment-229646757) and we came to the conclusion that by moving all our code to separate repos with their own linting rules would be better, however, if you disagree you're welcome to share your thoughts with us 😄 .

vieiralucas:
I'll just leave my LGTM in case you guys think that my comment is just nitpicking.

Agreed with you and @meeber about `Map`, `Set`, `WekMap` and `WeakSet`.
Agreed about adding `empty` to the `assert` interface
Agreed about the problem with code style and missing eslint thing. 😸 

P.S. This just made me realize that we might be throwing `TypeError` instead of plain `Error` on [closeTo](https://github.com/chaijs/chai/blob/5dd96a0a3d3bd711ddc688bcd395588aa17a8819/lib/chai/core/assertions.js#L1756), [above](https://github.com/chaijs/chai/blob/5dd96a0a3d3bd711ddc688bcd395588aa17a8819/lib/chai/core/assertions.js#L680), [least](https://github.com/chaijs/chai/blob/5dd96a0a3d3bd711ddc688bcd395588aa17a8819/lib/chai/core/assertions.js#L741) and [below](https://github.com/chaijs/chai/blob/5dd96a0a3d3bd711ddc688bcd395588aa17a8819/lib/chai/core/assertions.js#L802).

lib/chai/core/assertions.js
lucasfcosta: As @vieiralucas said, I think `.empty()` should be changed to `.empty` since it's a property assertion.

lib/chai/core/assertions.js
vieiralucas: That blew my mind to!

meeber:
@lucasfcosta @vieiralucas UH OH THE LUCAS SYNDICATE IS IN FULL FORCE TODAY. Just a reminder to whoever ends up merging: Squash and merge! :D

lucasfcosta:
@meeber hahaha yes, thanks for remembering us to do that 😄 
Let's just wait for the `.empty()` change to make the error more self-explanatory (which IMO is a very important thing). Are you guys okay with that or am I being too nitpicky?

shvaikalesh:
I have removed `()` and added function names and primitives values to messages for better debugging experience (is that OK?).

lib/chai/core/assertions.js
```diff
@@ -537,10 +537,11 @@ module.exports = function (chai, _) {
         itemsCount = val.length;
         break;
       case 'function':
-        throw new TypeError('.empty() was passed a function');
+        var name = val.name ? ' ' + val.name : '';
+        throw new TypeError('.empty was passed a function' + name);
       default:
         if (val !== Object(val)) {
-          throw new TypeError('.empty() was passed non-string primitive');
+          throw new TypeError('.empty was passed non-string primitive ' + String(val));
```
shvaikalesh: We need `String` it special-cases symbols. Otherwise, they throw on `ToString`.

vieiralucas: You could use the [inspect](https://github.com/chaijs/chai/blob/master/lib/chai/utils/inspect.js) utility.

lucasfcosta: I think I'm okay without it since the error message is pretty self-explanatory.
Also, try using `_.type()`, it may be more accurate.

lucasfcosta: Ah, I just noticed you want to print the **value** of what was passed, not it's type.
Then @vieiralucas is right.

vieiralucas: Type is better IMO

lucasfcosta: LOL, THE LUCAS SYNDICATE IS WILD TODAY
So, what do we want here, `type` or `value`?

shvaikalesh: Plugged `inspect` in, works like a charm. Thanks, guys!

vieiralucas: OMG I'M SO CONFUSE. inspect it is then?

lucasfcosta: okay, so, if we're going to use `type` we need to rephrase the error message to something like:

`.empty was passed non-string primitive of type ' + _.type(val)`

lucasfcosta: Okay, calm down everyone:

THUMBS UP IF YOU VOTE FOR `TYPE` AND HEART IF YOU VOTE FOR `INSPECT`.
If we three vote there's no way this is going to be a tie.

shvaikalesh: Okay, let's do `type` then.

shvaikalesh: I am for `inspect` because you can definitely tell the type by the value, but not otherwise. May save `console.log` typing for someone.

lucasfcosta: Okay, so we're going for `inspect`.
Also, since `inspect` also adds delimiters which denote the type of the value (like `[]` for arrays and `''` for strings), it is more complete.
Thanks for this @shvaikalesh!

LGTM!

vieiralucas:
LGTM ?

lucasfcosta:
Okay, LGTM, feel free to do the honors and squash and merge if you agree @vieiralucas.

shvaikalesh:
Thanks for the merge, awesome contribution experience ❤️ .

lucasfcosta:
Squashed and merged!
Awesome work @shvaikalesh, I'm looking forward to see you here again 😄

vieiralucas:
Thank you so much @shvaikalesh, looking forward for your next contributions. 😄

meeber:
No more caffeine for The Lucas Syndicate.

PR #813, opened 2016-10-02
Make `utils.getName` more robust & use it in `empty`

shvaikalesh:
Current ponyfill routine does not handle `toString` defined on functions. That is kinda rare, but does not take much effort to support. Beware, proposed "fix" will throw on non-functions (is that desirable?). Also, added basic tests for `getName`.

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -537,8 +537,8 @@ module.exports = function (chai, _) {
         itemsCount = val.length;
         break;
       case 'function':
-        var name = val.name ? ' ' + val.name : '';
-        throw new TypeError('.empty was passed a function' + name);
+        var msg = '.empty was passed a function ' + _.getName(val);
+        throw new TypeError(msg.trim());
       default:
         if (val !== Object(val)) {
           throw new TypeError('.empty was passed non-string primitive ' + _.inspect(val));
--- a/lib/chai/utils/getName.js
+++ b/lib/chai/utils/getName.js
@@ -14,13 +14,14 @@
  * @name getName
  */
 
+var toString = Function.prototype.toString;
 var functionNameMatch = /\s*function(?:\s|\s*\/\*[^(?:*\/)]+\*\/\s*)*([^\s\(\/]+)/;
 
 module.exports = function (func) {
   var name = '';
   if (typeof func.name === 'undefined') {
     // Here we run a polyfill if func.name is not defined
-    var match = String(func).match(functionNameMatch);
+    var match = toString.call(func).match(functionNameMatch);
     if (match) {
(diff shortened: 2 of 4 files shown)


meeber:
Hmm. I think `String` calls `toString` internally. The new tests you added pass without the changes. Do you see any other benefits from doing it this way? Regardless, I think it's worth merging the new tests you added.

shvaikalesh:
It does, but if custom `toString` defined directly on a function it will fail. The tests pass without the changes because both Node and Phantom support decent subset of ES6 name property. However, `bar.toString = function()` test will fail in IE10-11 without the change.

lucasfcosta:
This LGTM.
I think we should also add those changes to `check-error` btw.
Maybe moving that into a separate module would be nice for a next release such as `5.0.0` so we avoid code duplication.

I'll create an issue for that.

meeber:
Okay, I'm convinced. LGTM!

lib/chai/core/assertions.js
```diff
@@ -537,8 +537,8 @@ module.exports = function (chai, _) {
         itemsCount = val.length;
         break;
       case 'function':
-        var name = val.name ? ' ' + val.name : '';
-        throw new TypeError('.empty was passed a function' + name);
+        var msg = '.empty was passed a function ' + _.getName(val);
+        throw new TypeError(msg.trim());
```
keithamus: Ping to @lucasfcosta: I think perhaps we should call `.trim()` any time we return a value from `getName` - just a note for https://github.com/chaijs/check-error/pull/11 and https://github.com/chaijs/chai/pull/813

meeber: The `trim` in this case isn't being used because `getName` might return a string with spaces on the end (as far as I can tell it wouldn't) but rather because if `getName` returned an empty string then the `msg` that's being constructed on line 540 would have a space on the end. In other words, the `trim` would still be needed here even if `getName` did it itself.

shvaikalesh: Yep, that is exactly why I used `trim` here. However, because function `name` property is `configurable` in ES6, it might contain trailing spaces.

meeber: Ehhh, my feeling is that if someone goes out of there way to explicitly add spaces in such a weird way, then `getName` respects it and returns it as-is.

lucasfcosta: @meeber exactly. Great explanation :+1: 
And, as you said in the last comment, if someone explicitly added spaces to a function's name, I think we should keep it that way.

keithamus: Ah yes, that makes sense. Thanks all 😅

PR #814, opened 2016-10-02
Make `empty` assertion work with es6 collections

shvaikalesh:
Follow-up of #763 and #812.

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -540,6 +540,9 @@ module.exports = function (chai, _) {
       case 'set':
         itemsCount = val.size;
         break;
+      case 'weakmap':
+      case 'weakset':
+        throw new TypeError('.empty was called on a weak collection');
       case 'function':
         var name = val.name ? ' ' + val.name : '';
         throw new TypeError('.empty was passed a function' + name);
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -536,6 +536,10 @@ module.exports = function (chai, _) {
       case 'string':
         itemsCount = val.length;
         break;
+      case 'map':
+      case 'set':
+        itemsCount = val.size;
+        break;
       case 'function':
         var name = val.name ? ' ' + val.name : '';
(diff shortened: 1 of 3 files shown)


lib/chai/core/assertions.js
```diff
@@ -536,6 +536,13 @@ module.exports = function (chai, _) {
       case 'string':
         itemsCount = val.length;
         break;
+      case 'map':
+      case 'set':
+        itemsCount = val.size;
+        break;
+      case 'weakmap':
+      case 'weakset':
+        throw new TypeError('.empty was called on a weak collection');
```
vieiralucas: I think that to keep consistency with the TypeError messages added to #812, this should be something like this:
`.empty was passed a weak collection`

vieiralucas:
Hello @shvaikalesh, I'm very happy to see you here again.

This change seems very clean, awesome work here.
I just made a comment on the message you added on weak collections. Let me know what you think, feel free to disagree with me.

lib/chai/core/assertions.js
shvaikalesh: Right, thanks, fixed that. Another error message was just 3 lines away, what's wrong with me...

shvaikalesh:
Not that fast, forgot the tests :(

lucasfcosta:
Well done, LGTM too!
@vieiralucas would you like to do the honors and squash + merge this?

meeber:
Sec let me check something

meeber:
@shvaikalesh Annoyingly, `Map` and `Set` constructor arguments aren't supported in IE11 so will need to set the values manually in the tests. Also recommend adding a little note like [this](https://github.com/chaijs/chai/blob/5b170b2e3bae6e918ef723c168668d829d7bc61b/test/expect.js#L1409-L1415). Otherwise looks great!

shvaikalesh:
@meeber Nice catch, I kinda forgot about partial implementations.

meeber:
LGTM! Thank you for your contributions! :D

vieiralucas:
Squashed and merged!

Thank you again @shvaikalesh

Issue #578, opened 2015-12-24
Error on strict mode


tresdin:
I got this error with the code `expect(value).to.be.null` that run on strict mode.
![chai-error-strict-mode](https://cloud.githubusercontent.com/assets/7781091/11989329/96884c9e-aa2d-11e5-8563-76ed70d3617c.png)

P/s. I'm using `chai@3.4.1` installed via `npm install chai`.

keithamus:
Thanks for the issue @tresdin.

Bow are you running these tests? Through Mocha? Through Karma? Are you using CommonJS style `require`? Or are you dropping in `<script>` tags onto a page? It looks as though you're probably concatenating Chai with something else that declares `"use strict"`.

tresdin:
@keithamus 
> Bow are you running these tests? Through Mocha? Through Karma?

I'm using `Mocha`.

> Are you using CommonJS style `require`?

Yes. I am.

> Or are you dropping in `<script>` tags onto a page

No. This code is run on `nodejs`. This error doesn't occur when I run [`chai`](https://github.com/chaijs/chai/blob/master/chai.js) on browsers

> It looks as though you're probably concatenating Chai with something else that declares "use strict".

I run node globally on strict mode (via `--use_strict` flag)

keithamus:
> I run node globally on strict mode (via `--use_strict` flag)

That's your problem! Remove that flag and things will work fine! Not every module you use will implement strict code (for example, Chai does not!). You'd be better off declaring `"use strict";` at the top of every file that you want to use strict parsing.

I'm going to close this because we rely on non-strict features (which are still compliant to the JS spec, just not the "strict" part of the spec). It's unlikely we'll change this any time soon, and only errors through edge cases (such as enforcing strict mode on all of your modules, or using babel, etc).

tresdin:
@keithamus Bad news, indeed. I think you should seriously consider this issue as a critical bug. `chai` should work on strict mode.

keithamus:
@tresdin I'm curious; why should chai work in strict mode? "Strict Mode" is an opt in part of JavaScript and as such is not a requirement for engines (or libraries) to implement. Some engines can enforce strict mode (e.g. via `--use_strict`) which is arguably non-compliant as not all code has to be strict.

[...]

PR #815, opened 2016-10-02
Remove implicit global `actual`

shvaikalesh:
Would you like a PR with strict mode added?


vieiralucas:
This LGTM.

Regarding strict mode, I think you could read #578.

meeber:
LGTM!

There's a lot of discussion regarding strict mode in #578. Seems inevitable that it'll be added.

shvaikalesh:
Thanks for the link, interesting reading 👍

Issue #45, opened 2012-04-06
Checking `ErrorConstructor.name` does not always work


domenic:
Chai's `throw` <a href="https://github.com/logicalparadox/chai/blob/master/lib/assertion.js#L889-892">checks `ErrorConstructor.name` against the thrown error</a>, and uses it in the assertion output.

This is incompatible with dynamically-constructed errors, e.g. <a href="https://github.com/mcavage/node-restify/blob/master/lib/errors/rest_error.js#L63-76">those from Restify</a>, since the constructors there don't have a `name` property. (See also <a href="https://github.com/mcavage/node-restify/pull/116">proposed solution on Restify's side using `eval`</a>.)

We can fix this in the case of some libraries by testing against `ErrorConstructor.prototype.name`, but note that not all custom errors properly set up the prototype. (In particular, Restify before <a href="https://github.com/mcavage/node-restify/pull/116">my pull request</a> sets it on the instances only.) A more bulletproof, but kind of inelegant, fix would be to construct a new instance of the error simply to inspect its name property:

``` js
var dummyError = new constructor();
var name = dummyError.name;
```

Thoughts? I'd be happy to send a pull request for either approach if desired.

PR #817, opened 2016-10-03
Fix getName for poorlyConstructed errors

lucasfcosta:
This is the same fix as chaijs/check-error#11.

Now `getName` will continue working with `Poorly Constructed Errors` (#45).

This is related to #813.

### DIFF
--- a/lib/chai/utils/getName.js
+++ b/lib/chai/utils/getName.js
@@ -19,13 +19,14 @@ var functionNameMatch = /\s*function(?:\s|\s*\/\*[^(?:*\/)]+\*\/\s*)*([^\s\(\/]+
 
 module.exports = function (func) {
   var name = '';
-  if (typeof func.name === 'undefined') {
-    // Here we run a polyfill if func.name is not defined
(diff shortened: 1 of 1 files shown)


lib/chai/utils/getName.js
```diff
@@ -19,13 +19,14 @@ var functionNameMatch = /\s*function(?:\s|\s*\/\*[^(?:*\/)]+\*\/\s*)*([^\s\(\/]+
 
 module.exports = function (func) {
   var name = '';
-  if (typeof func.name === 'undefined') {
-    // Here we run a polyfill if func.name is not defined
+  if (typeof Function.prototype.name === 'undefined' && typeof func.name === 'undefined') {
```
shvaikalesh: Maybe it is possible to hoist this check to module scope?

keithamus:
Just like chaijs/check-error#11, let's split this into a new module.

lucasfcosta:
I'll close this one since it's going to be split into a new module.
I already opened a PR for that in the repo @keithamus created.

PR #819, opened 2016-10-03
Remove extra `exports` and use version from pkg.json

shvaikalesh:


meeber:
LGTM

vieiralucas:
LGTM!

lib/chai.js
```diff
@@ -4,14 +4,13 @@
...
-var used = []
-  , exports = module.exports = {};
+var used = [];
 
 /*!
  * Chai version
  */
 
-exports.version = '3.5.0';
+exports.version = require('../package').version;
```
keithamus: Just FYI our tooling automates writing this out as the string, so that we don't end up including all of the package.json in the browserify bundle.

meeber: @keithamus Whoops. Revert? Also now that we have more maintainers/reviewers, maybe we should add a minimum time before merging a PR even after it has two approvals.

vieiralucas: OH,

Should we revert this?

shvaikalesh: Yeah, when bundling for browsers result package size is an issue. We can use Rollup to decrease bundle (tree-shaking and stuff) and include only the `version` field of `package.json`.

keithamus: @meeber @vieiralucas don't worry about reverting, our build process will write over this change when we release 4.0

@shvaikalesh We've already got the tooling that writes the version. We can definitely look at Rollup though; do you want to open an issue to discuss it?

keithamus: Also @meeber - I don't think we should add things like minimum time to review. This stuff happens, it's teething issues, but ultimately no big deal. Nothing egregious has happened - when it does we'll deal with it then!

shvaikalesh: #821

PR #820, opened 2016-10-03
Add missing semicolons and remove unnecessary semicolons

vieiralucas:


lucasfcosta:
LGTM.
Also, maybe we could have a single `eslint` run with `auto-fix` flag just to make the whole code more consistent.
What do you guys think?

vieiralucas:
@lucasfcosta I totally agree with that

If everyone agrees I would be very happy to make a PR

eslint ftw

keithamus:
I think we shouldn't devote any time to this right now - this PR is fine and LGTM, but I really feel like we're going to change a lot of this code very soon, so we should focus on putting brain cycles into working that out.

meeber:
Here's what I think:

1. We should wait until after v4 is released before doing any wide-scale linting fixes. We have so many changes being introduced (many of them breaking) that we may need to troubleshoot that I'd prefer not to have the diff substantially altered by linting fixes at this time.
1. After v4 is released, we should prioritize getting linting implemented. I think it's important enough to the reviewer and contributor experience that we shouldn't hold off until stuff is separated out into modules. I don't think the effort will be lost, as the linted fixes will then be carried over to the individual modules.

vieiralucas:
😢

You guys are right haha

lucasfcosta:
@meeber I totally agree with you. It makes total sense, I haven't thought about diffs.
Also, I really think `linting` is important because we need some way to mitigate code-style problems until we've got everything split into separate modules.

Also, a knowledge pearl:
According to Robert Cecil Martin, author of [Clean Code](https://www.google.com.br/url?sa=t&rct=j&q=&esrc=s&source=web&cd=1&cad=rja&uact=8&ved=0ahUKEwiso6bQ4L7PAhUHjZAKHfTDANkQFggcMAA&url=https%3A%2F%2Fwww.amazon.com.br%2FClean-Code-Handbook-Software-Craftsmanship%2Fdp%2F0132350882&usg=AFQjCNEr9KCIm2n799bfO_mBQVy3SJqFEg&sig2=REu52R4QLgkyg5ULMR3wsQ&bvm=bv.134495766,d.Y2I), the most common software problems don't come from high level/intellectual decisions in general. They actually come from code style problems which don't make things clear enough. So organization is really important.

Well, take @meeber's words as mine, excellent arguments.
Also, @keithamus is also right, let's keep our focus on getting `4.x.x` out first.

keithamus:
@meeber with regards to point 2; a lot of the code will undergo refactoring when splitting out into modules - I doubt it will be a straight copy/paste, so that is a good time to standardise on one linting style. If we adopt one now (I'd like to use https://github.com/keithamus/eslint-config-strict) then that it could be a relatively significant distraction for one contributor to maintain lint fixes all the while PRs are happening - compared to chunking out components and having individual contributors fix their individual modules.

keithamus:
Also _ultimately_ I think Linting is incredibly important, which is why I've been pushing for it to be added to all modules. Having said that, I want to be pragmatic about the core codebase - I'd rather take bits out peacemeal and fix them up, just like we've been doing with type-detect, check-error, deep-eql and soon to be loupe.

meeber:
@keithamus A fair approach might be to just drop in the linting as an experiment and see how big of an effort it will really be before deciding one way or the other. But we seem to agree on the most important point for now which is to wait until after 4.x before doing anything :D

lucasfcosta:
Hi friends, I think that we might want to close this one for now and implement linting tasks properly so that we can get more cleaner code. I also would like to implement pre-commit validations such as they do in [Sinon](https://github.com/sinonjs/sinon/blob/master/package.json#L28) to guarantee all the pull requests have the same code standards.

What do you think?

PR #823, opened 2016-10-03
Do not suppress `TypeError`s thrown on integrity checks

shvaikalesh:
Current implementation will report false-positive result if `TypeError` is thrown in some `Proxy` traps (`isExtensible`, `ownKeys`, `getOwnPropertyDescriptor`).

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -2186,10 +2186,10 @@ module.exports = function (chai, _) {
   Assertion.addProperty('extensible', function() {
     var obj = flag(this, 'object');
 
-    // In ES5, if the argument to this method is not an object (a primitive), then it will cause a TypeError.
+    // In ES5, if the argument to this method is a primitive, then it will cause a TypeError.
     // In ES6, a non-object argument will be treated as if it was a non-extensible ordinary object, simply return false.
     // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/isExtensible
-    // The following provides ES6 behavior when a TypeError is thrown under ES5.
+    // The following provides ES6 behavior for ES5 environments.
 
     var isExtensible = obj === Object(obj) && Object.isExtensible(obj);
 
@@ -2222,10 +2222,10 @@ module.exports = function (chai, _) {
   Assertion.addProperty('sealed', function() {
     var obj = flag(this, 'object');
 
-    // In ES5, if the argument to this method is not an object (a primitive), then it will cause a TypeError.
+    // In ES5, if the argument to this method is a primitive, then it will cause a TypeError.
     // In ES6, a non-object argument will be treated as if it was a sealed ordinary object, simply return true.
     // See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/isSealed
-    // The following provides ES6 behavior when a TypeError is thrown under ES5.
+    // The following provides ES6 behavior for ES5 environments.
 
     var isSealed = obj === Object(obj) ? Object.isSealed(obj) : true;
 
@@ -2256,10 +2256,10 @@ module.exports = function (chai, _) {
   Assertion.addProperty('frozen', function() {
     var obj = flag(this, 'object');
 
-    // In ES5, if the argument to this method is not an object (a primitive), then it will cause a TypeError.
+    // In ES5, if the argument to this method is a primitive, then it will cause a TypeError.
     // In ES6, a non-object argument will be treated as if it was a frozen ordinary object, simply return true.
     // See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/isFrozen
-    // The following provides ES6 behavior when a TypeError is thrown under ES5.
+    // The following provides ES6 behavior for ES5 environments.
 
     var isFrozen = obj === Object(obj) ? Object.isFrozen(obj) : true;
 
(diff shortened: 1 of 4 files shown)


meeber:
LGTM. I wasn't aware that `isSealed` and `isFrozen` was caught by the proxy trap for `isExtensible` but now that I see it, it makes sense.

On a side note, I think we should move `obj === Object(obj)` to a new `isObject` utility function, to make it easier for future contributors to use as well as make it easier to change implementation in the future if we ever need to. (But not in this PR)

shvaikalesh:
@meeber Noted on `isObject` helper. I used `isExtensible` for tests because it is shorter than other traps mentioned: `isFrozen` won't call `ownKeys` unless the object is not extensible. I can tests for other methods, though. But I thought it would be too much code for such an edge case.

meeber:
I think it's good the way you have it. I'd just never used the `isExtensible` trap before, and hadn't thought about it being triggered by `isSealed` and `isFrozen`. But it makes sense. And it's indirectly inferred by the ECMAScript spec.

shvaikalesh:
The [spec](https://tc39.github.io/ecma262/#sec-testintegritylevel) is pretty accurate on which internal methods are called and when. I'have diversified tests a bit: I make `proxy` not extendable and throw `TypeError` in `OwnPropertyKeys` trap for `frozen` and `sealed` tests.

lib/chai/core/assertions.js
```diff
@@ -2232,14 +2227,7 @@ module.exports = function (chai, _) {
     // See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/isSealed
     // The following provides ES6 behavior when a TypeError is thrown under ES5.
 
-    var isSealed;
```
meeber: Same thing in 2225-2228 with comments update.

test/assert.js
```diff
@@ -1922,6 +1946,20 @@ describe('assert', function () {
...
+            throw new TypeError();
+          }
+        });
+
+        Object.preventExtensions(proxy);
+
+        err(function() {
+          assert[isSealed](proxy);
+        }, { name: 'TypeError' });
+      }
```
meeber: I think the above test is advanced enough that it calls for a short comment explaining why it is the way it is, particularly with regard to the `preventExtensions` and `ownKeys`. A comment will lower the barrier of entry to new contributors and reduce risk of someone submitting a change to it without realizing how it works internally.

test/assert.js
```diff
@@ -1980,6 +2032,20 @@ describe('assert', function () {
...
+            throw new TypeError();
+          }
+        });
+
+        Object.preventExtensions(proxy);
+
+        err(function() {
+          assert[isFrozen](proxy);
+        }, { name: 'TypeError' });
+      }
```
meeber: Same thing with short comment explaining above test.

test/assert.js
```diff
@@ -2014,6 +2080,20 @@ describe('assert', function () {
...
+            throw new TypeError();
+          }
+        });
+
+        Object.preventExtensions(proxy);
+
+        err(function() {
+          assert[isNotFrozen](proxy);
+        }, { name: 'TypeError' });
+      }
```
meeber: Same thing with short comment explaining above test.

test/expect.js
```diff
@@ -2416,6 +2428,20 @@ describe('expect', function () {
...
+          throw new TypeError();
+        }
+      });
+
+      Object.preventExtensions(proxy);
+
+      err(function() {
+        expect(proxy).to.be.sealed;
+      }, { name: 'TypeError' });
+    }
```
meeber: Same thing with short comment explaining above test.

test/expect.js
```diff
@@ -2463,5 +2489,19 @@ describe('expect', function () {
...
+          throw new TypeError();
+        }
+      });
+
+      Object.preventExtensions(proxy);
+
+      err(function() {
+        expect(proxy).to.be.frozen;
+      }, { name: 'TypeError' });
+    }
```
meeber: Same thing with short comment explaining above test.

test/should.js
```diff
@@ -2241,6 +2253,20 @@ describe('should', function() {
...
+          throw new TypeError();
+        }
+      });
+
+      Object.preventExtensions(proxy);
+
+      err(function() {
+        proxy.should.be.sealed;
+      }, { name: 'TypeError' });
+    }
```
meeber: Same thing with short comment explaining above test.

shvaikalesh:
@meeber Thanks for reviewing this one, your help is greatly appreciated!

meeber:
My pleasure. Awesome work! This is an important PR for improving our ES6 support, especially since it's part of the `isExtensible` proxy trap's behavior to throw a `TypeError` if it returns an untruthful value. Definitely don't want Chai to swallow that error!

Let's get some more eyes on this before merging: @lucasfcosta @vieiralucas @keithamus

lucasfcosta:
Thanks for the PR @shvaikalesh! Awesome work and thanks for including tests, this will be a great improvement.

LGTM. Also, you seem to be a jedi master in the use of `Object()`. I learned a great lesson with you, I'm gonna use it more often. 😄

shvaikalesh:
@lucasfcosta Thanks, but beware, `stuff === Object(stuff)` used to cause wrong results in older  V8 (Node.js 0.8, Chrome 17-18) JITs. It is safe now.

lucasfcosta:
Does any of you have permissions to restart the build? Apparently it has failed for no reason.

meeber:
Restarted

PR #825, opened 2016-10-03
Fix `hasProperty` called on symbols and booleans

shvaikalesh:
Greatly simplifies the code and makes consistent with behavior of other primitive types.

### DIFF
--- a/lib/chai/utils/hasProperty.js
+++ b/lib/chai/utils/hasProperty.js
@@ -4,17 +4,15 @@
  * MIT Licensed
  */
 
-var type = require('type-detect');
-
 /**
- * ### .hasProperty(object, name)
+ * ### .hasProperty(name, object)
  *
- * This allows checking whether an object has
- * named property or numeric array index.
+ * This allows checking whether an object has own
+ * or inherited from prototype chain named property.
  *
  * Basically does the same thing as the `in`
- * operator but works properly with natives
- * and null/undefined values.
+ * operator but works properly with null/undefined values
+ * and other primitives.
  *
  *     var obj = {
  *         arr: ['a', 'b', 'c']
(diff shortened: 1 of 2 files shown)


lucasfcosta:
Hi @shvaikalesh, thanks for the PR and welcome to the team! :smile: 
This is looking good, however I'd just like to talk to other maintainers about all the PRs we've got.
Currently we have several PRs and this made me think that maybe we should start opening issues before sending PRs with changes that have not been discussed before in order for everyone to share their ideas and try having the best implementation possible if we all agree on the change that is going to be made. What do you guys think?

Also, having issues helps us keep track of what should be done and keep discussion there instead of keeping it inside PRs. Not every PR would need an issue, of course, it should just have been discussed before if it's not a fix or anything like that. IMO.

Also, please don't feel bad because just because of this comment, you have been doing a great work, I just want to make sure we find a good way to debate changes  :heart: 

Thanks again for the time and effort dedicated to Chai, I'm sure we all will achieve great things together :smile:

keithamus:
This looks good, but how does it perform compared to the old way?

lib/chai/utils/hasProperty.js
```diff
@@ -35,30 +33,19 @@ var type = require('type-detect');
...
-    'number': Number
-  , 'string': String
-};
-
 module.exports = function hasProperty(name, obj) {
-  var ot = type(obj);
-
-  // Bad Object, obviously no props at all
-  if(ot === 'null' || ot === 'undefined')
+  if (obj === undefined || obj === null) {
```
meeber: We lose a little bit of safety net here from using `obj === undefined` instead of `typeof obj === 'undefined'`, in case the user has done terrible things and redefined `undefined`. However, I'm not sure if this safety net is actually necessary, and I don't have time at the moment to check what we do elsewhere in the code.

shvaikalesh: `undefined` can't be redefined since IE9. According to #124 and #117, IE8 is not supported.

meeber:
LGTM

shvaikalesh:
@keithamus `Object` is almost instant even for primitives, compared to `type-detect` logic.

lucasfcosta:
Hi @shvaikalesh!
Actually this will be replaced by `pathval`'s code, since this functionality has been moved there.
Maybe we could have this change on that repo, would you like to create PR for that there @shvaikalesh?

Also, the arguments' order there should not be replaced, since it's already consistent with the other functions.

keithamus:
@lucasfcosta I don't think the arguments order is changing, I think there's just a fix for wonky docs.

lucasfcosta:
@keithamus ooooooops, our docs are so good I ended up getting confused 😆 
Thanks!

shvaikalesh:
@lucasfcosta got it, will close this one.

lucasfcosta:
@shvaikalesh anyway, feel free to send a PR to `pathval` with that change!
Thanks for your work buddy :smile:

PR #828, opened 2016-10-05
Add .empty to assert interface

shvaikalesh:
Follow-up to #812.

### DIFF
--- a/lib/chai/interface/assert.js
+++ b/lib/chai/interface/assert.js
@@ -2768,6 +2768,58 @@ module.exports = function (chai, util) {
     new Assertion(obj, msg).to.not.be.frozen;
   };
 
+  /**
+   * ### .isEmpty(target)
+   *
+   * Asserts that the target does not contain any values.
+   * For arrays and strings, it checks the `length` property.
+   * For `Map` and `Set` instances, it checks the `size` property.
+   * For non-function objects, it gets the count of own
+   * enumerable string keys.
+   *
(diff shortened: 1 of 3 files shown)


meeber:
LGTM!

lib/chai/interface/assert.js
```diff
@@ -2768,6 +2768,58 @@ module.exports = function (chai, util) {
...
+   * enumerable string keys.
+   *
+   *     assert.isNotEmpty([1, 2]);
+   *     assert.isNotEmpty('34');
+   *     assert.isNotEmpty(new Set([5, 6]));
+   *     assert.isNotEmpty({ key: 7 });
+   *
+   * @name isNotEmpty
+   * @alias notEmpty
+   * @param {Object|Array|String} target
```
lucasfcosta: Isn't it worth it to include `Map` and `Set` here? Just to keep things explicit, what do you guys think?

lucasfcosta:
I just left a minor nitpicky comment about the accepted types of arguments for the method on the assert interface.

If you think it's not needed feel free to merge.

Despite that, LGTM! Awesome work @shvaikalesh! Thank you for this :smile:

shvaikalesh:
@lucasfcosta thanks for the review, fixed that.

lucasfcosta:
LGTM.
I'll merge this since the only change since @meeber's LGTM were comments.
Thanks for you work, good job :smile:

PR #829, opened 2016-10-05
Define variable

aristov:
Global variable `html` leak possible.

### DIFF
--- a/lib/chai/utils/inspect.js
+++ b/lib/chai/utils/inspect.js
@@ -84,7 +84,7 @@ function formatValue(ctx, value, recurseTimes) {
           var container = document.createElementNS(ns, '_');
 
           container.appendChild(value.cloneNode(false));
-          html = container.innerHTML
+          var html = container.innerHTML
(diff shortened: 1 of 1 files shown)


shvaikalesh:
Hey @aristov, thanks for the PR.
Since FF < 12 is not supported by Chai, I think we should just do

```js
return value.outerHTML
```

lucasfcosta:
Hi guys, thanks @aristov for the PR.
Well, I think you're right @shvaikalesh, if we don't support the browser for which we need this feature anymore we should just drop it.
Since I've just seen you made a new PR for that I'll give my LGTM seal of approval for this one and we can discuss this other change on the new PR 😄

aristov:
@shvaikalesh @lucasfcosta 
open source works 👍

keithamus:
Thanks for your contributions @aristov. Sorry we couldn't get this one merged, but it has definitely spurred some improvements, so thanks for making chai better 😄 🎉

PR #830, opened 2016-10-05
Use external pathval module

lucasfcosta:
This includes a refactor for Chai to use the external `pathval` module as we discussed previously on #457 and #737.

I also removed the old tests since they're now running at the separate module and removed an unused utility export (`.getPathValue`), if you guys want me to keep it just tell me, but since we were not using it I just removed that, which is also a breaking change if anyone is using `utils.getPathValue`. If you disagree let me know, I'd love to hear your opinions on this matter.

Also, since we needed `pathval`'s exports twice I added a new declaration at the top of our `utils/index.js` file for that, I thought it would be better to do it this way than thrusting Node would use its cache correctly to find the same module. If you think there's a better way to do this, share your ideas 😄 

At last, but not least, please take a look at [this comment](https://github.com/chaijs/pathval/pull/39#issuecomment-251744081) before merging this PR since the latest release of `pathval` has breaking changes but we just bumped the `minor` version number, so if we decide to release a new `major` version I'll have to change `package.json`.

Thanks guys! ❤️


meeber:
Looks good but agreed on waiting for pathval version fix before merging.

lib/chai/utils/index.js
```diff
@@ -5,6 +5,18 @@
...
+ * Dependencies that are used for multiple exports are required here only once
+ */
+
+var pathval = require('pathval');
+
+/*!
+ * Main exports
+ */
+
+var exports = module.exports = {};
```
shvaikalesh: Am I missing something or this line is extra?

lucasfcosta: You are totally right! I messed up this file on the latest rebase.

vieiralucas:
-318 lines 😎

keithamus:
LGTM! Great work @lucasfcosta

meeber:
@lucasfcosta As mentioned in https://github.com/chaijs/pathval/pull/39#issuecomment-252423864, I think that we should manually update pathval to 1.0.0 and then update this PR to use that.

lucasfcosta:
This one is done!
Also, thanks for helping us release the `1.0.0` version of `pathval` @meeber!

I also added the following note to our migration guide (just in case):

- The `utils` (second argument passed to the `chai.use` callback function) no longer exports the `getPathValue` function. If you want to use that please use the [`pathval`](https://github.com/chaijs/pathval) module, which is what chai uses internally now. *(Related Issues: #457, PRs: #830)*

meeber:
LGTM

PR #835, opened 2016-10-09
Update type-detect to v3.0.0 and fix issues

meeber:
- Use typeof instead of type-detect whenever possible for performance
- Fix casing whenever type-detect is used per v3.0.0 breaking change
- Remove duplicate type-detect calls by storing results in variables

Edit: Depends on the removal of `hasProperty` in #836


lib/chai/core/assertions.js
```diff
@@ -303,11 +303,12 @@ module.exports = function (chai, _) {
 
     if (msg) flag(this, 'message', msg);
     var obj = flag(this, 'object')
+      , objType = _.type(obj).toLowerCase()
       , isDeep = flag(this, 'deep')
       , descriptor = isDeep ? 'deep ' : '';
 
     // This block is for asserting a subset of properties in an object.
-    if (_.type(obj) === 'object') {
+    if (objType === 'object') {
```
vieiralucas: What if instead of doing `.toLowerCase` above.
We changed here to `'Object'` ?

lib/chai/core/assertions.js
```diff
@@ -534,7 +535,7 @@ module.exports = function (chai, _) {
     var val = flag(this, 'object')
       , itemsCount;
 
-    switch (_.type(val)) {
+    switch (_.type(val).toLowerCase()) {
```
vieiralucas: Same thoughts as above.
What if instead of this `.toLowerCase()` we updated the cases, to the capitalized version?
Like you did in the switch case below, at line 1407

lib/chai/core/assertions.js
meeber: I added the `.toLowerCase` anytime that excluding it would mean that we'd have to perform additional comparisons as a result. For example, on line 346 we'd have to perform a second comparison for "String" (to handle the object version) in addition to the current one for "string".

vieiralucas: So, I was missing something 😄 
Understood

lib/chai/utils/hasProperty.js
```diff
@@ -49,7 +49,7 @@ var literals = {
 };
 
 module.exports = function hasProperty(name, obj) {
-  var ot = type(obj);
+  var ot = type(obj).toLowerCase();
```
vieiralucas: It looks like we do not have unit testing for `hasProperty` utility.

I don't know what are the plans for this utility, is this going to a separate module?
Do you think that it's worth to add unit tests to it right now?

vieiralucas: It is already on another module 😸

vieiralucas:
Now that you've enlightened me 💡 . This LGTM

keithamus:
LGTM now 😄

Issue #684, opened 2016-04-21
.length(value) deprecation


lencioni:
I noticed that the documentation says:

> Deprecation notice: Using length as an assertion will be deprecated in version 2.4.0 and removed in 3.0.0. Code using the old style of asserting for length property value using length(value) should be switched to use lengthOf(value) instead.

But I also noticed that chai is on 3.5.0 and `.length(value)` seems to still be a thing. Am I missing something here?

ljharb:
Also, why is `.to.have.length` deprecated at all?

lucasfcosta:
Hi guys, thanks for the issue!

@ljharb 
I hope @keithamus will correct me if I'm wrong, but I think this is related to #642.
When calling `.length` after using an `a` assertion we get back the `.length` property of a function instead of the `.length` property of the Assertion prototype and this causes bug, because people expect it to be the `length` assertion and not a function's `length` property.
If you want more details on it you can take a look at #642, especially at [this post](https://github.com/chaijs/chai/pull/642#issuecomment-198979726).

@lencioni
What's been deprecated is using `length` as an assertion.
At [this issue](https://github.com/chaijs/chai/issues/562#issuecomment-197460995) you can read more about this deprecation. It includes an explanation by @keithamus on this matter. I think it may solve your doubts.

Please let me know if you guys want to know anything else :smile:

keithamus:
Yeah, essentially there are some edge cases around the fluent interface - where chaining functions results in the return value being a function, not the instance object. Ultimately this means _sometimes_ `length` is `Function#length` not our assertion. Rather than detailing exactly where this happens, it's much easier to just deprecate `length` and instead use `lengthOf`.

shvaikalesh:
`4.0` is coming, but `.length` as an assertion is not yet removed. At least, I think, we should remove it from [homepage](http://chaijs.com/).

lucasfcosta:
@shvaikalesh well noticed!
I think we should take advantage of `4.x.x` and remove that from the code as well.
What do you guys think?

meeber:
I have a couple of concerns with this.

My first concern is that `length` as an assertion should only be deprecated if `length` as a flag-setting language chain is also deprecated. The same issue affects both variations so it would be a weird user experience to deprecate one but not the other:

```js
expect("blah").to.have.a.length(4);  // fails
expect("blah").to.have.a.length.within(3, 5);  // fails
```

[...]

PR #841, opened 2016-10-11
`length` as assertion => `lengthOf`

shvaikalesh:
Related to #684.

### DIFF
--- a/lib/chai/interface/assert.js
+++ b/lib/chai/interface/assert.js
@@ -1456,7 +1456,7 @@ module.exports = function (chai, util) {
    */
 
   assert.lengthOf = function (exp, len, msg) {
-    new Assertion(exp, msg).to.have.length(len);
+    new Assertion(exp, msg).to.have.lengthOf(len);
   };
 
   /**
--- a/lib/chai/utils/overwriteChainableMethod.js
+++ b/lib/chai/utils/overwriteChainableMethod.js
@@ -28,8 +28,8 @@ var transferFlags = require('./transferFlags');
  *
  * Then can be used as any other assertion.
  *
- *     expect(myFoo).to.have.length(3);
- *     expect(myFoo).to.have.length.above(3);
+ *     expect(myFoo).to.have.foo(3);
+ *     expect(myFoo).to.have.foo.bar(3);
  *
  * @param {Object} ctx object whose method / property is to be overwritten
  * @param {String} name of method / property to overwrite


meeber:
I have some reservations about deprecating `length` as an assertion but not deprecating `length` as a language chain / flag, given that they're both affected by the same problem. I'd like some time to think about it and organize my thoughts.

keithamus:
Well, the code in this PR LGTM. Let's get this one merged (as it's going towards fixing the issue) and move discussion about whether or not we should deprecate length wholesale back into #684

lucasfcosta:
This LGTM too.
I'll not merge this yet so @meeber can have a chance to do it if he has no urgent concerns about it.

meeber:
Let's table this for post-4.0?

lucasfcosta:
@meeber Agreed. First things first 😄

keithamus:
Shall we squeeze this in before 4.0 final? Canary is released so we can chuck this one in right?

meeber:
As discussed in #684, I'm opposed to deprecating `.length`; I believe the overall cost of doing so is way higher than the benefit. But if the other contributors feel strongly that it should be deprecated anyway, then this PR needs to be updated to also deprecate `.length` as a language chain; the issue being addressed affects `.length` both as an assertion and as a language chain. Note that deprecating `.length` as a language chain will leave some awkward phrasing behind, since we have examples out there like `expect(blah).to.have.length.below(5)` which don't translate very well to `expect(blah).to.have.lengthOf.below(5)`.

vieiralucas:
I'm with @meeber here.

I agree that the cost is way higher than the benefit.

> One consequence of removing length completely is that many Chai examples on the net will no longer work. The confusion experienced by new developers trying to use length based off such examples (or off of intuition) may be greater than they would experience in the event that they tried using have.a.length and didn't understand the error message.

/\ This is very important to me, as @meeber said:
> if Chai were a young project, there would be virtually no cost in doing so, making it the clearly correct choice.

IMHO if we deprecate `length` as assertion, we should deprecate it as a language chain too. And since `lengthOf` doesn't work everywhere, what do you think of using a synonym like `size`, or maybe short it to `len`?

lucasfcosta:
Hey everyone, sorry for the delay, I spent the whole weekend traveling.
I agree with @vieiralucas, maybe we could use a synonym and `size` looks good to me.
As @meeber said, the cost would be way higher than the benefit, so we could just add another synonym and encourage our users to use it instead of using length.

We could also add a deprecation warning on `4.0.0` recommending our users to use `size`, for example, and then remove `length` completely on `5.0.0`.

keithamus:
I'd honestly like to keep `length`, but it's a footgun. We can make it work in ES6 environments, but not on older environs. Cross browser issues are particularly sensitive for us - as an assertion framework - as it can start to look like the user's cross browser errors.

Sadly even IE11 doesn't support setting length, so we're stuck with this problem for a long time.

meeber:
I think we should just leave it as consistently broken across all environments until all environments we support are ES6-compliant, and then we can fix it. It's such a low-impact bug because it only causes a failure in rare circumstances, and it fails loudly instead of silently. But deprecating `length` entirely at this stage in the project is a high-impact solution; everyone will have to rewrite all of their tests using `length`, almost all of which were perfectly valid, and very little is gained as a result. After all, people will still attempt to use `length` even if it's deprecated, just based on intuition or existing code examples on the internet. The situation isn't ideal but the proposed solution is worse than the problem.

PR #843, opened 2016-10-16
Use external get-func-name module

lucasfcosta:
So, this includes a little "refactor" for us to use `get-func-name` instead of our internal implementation of `getName`.
I also removed redundant tests.


meeber:
Thanks for doing this. Let's add it to the "Related Projects" section in the README too.

lucasfcosta:
@meeber, great idea, I will update those too then!

lucasfcosta:
@meeber it's done!
Thanks for noticing this.

meeber:
LGTM!

lib/chai/utils/index.js
```diff
@@ -85,7 +85,7 @@ exports.hasProperty = pathval.hasProperty;
  * Function name
  */
 
-exports.getName = require('./getName');
+exports.getName = require('get-func-name');
```
meeber: I don't think it's worth changing.

vieiralucas:
LGTM

+4 -50
Always good to remove code

lucasfcosta:
Thanks for your reviews guys!
Let's get this merged!

🎉 🎉 🎉

Issue #862, opened 2016-11-07
`canary` version initially published as `latest`


ljharb:
chai v4.0.0-canary.1 was initially published as "latest", which sadly our internal sinopia cached, so it broke some builds.

I see that you've corrected it, and I'm sure it wasn't intentional.

Can you please use https://npmjs.com/safe-publish-latest, and try to use `--tag` in your `npm publish` commands, so that non-latest versions are not published as latest, even initially?

keithamus:
Thanks for the issue @ljharb.

Right now, our main releases are automated through travis. We are manually publishing canary versions for `4.0` as a one-off, but soon (likely 5.0) we'll be moving to [semantic-release](https://github.com/semantic-release/semantic-release), meaning all releases will be fully automated and only ever main (not alpha/beta/canary/whatever) releases.

I suppose we could add this though. @lucasfcosta thoughts?

ljharb:
I highly discourage automated publishing for many reasons, but that's not the point of this issue :-)

When manually publishing versions, if you add `--tag=canary` to `npm publish`, this won't happen.

(I'm not sure how semantic-release handles backports)

vieiralucas:
Hello @ljharb.

First of all, I'm very sorry you had issues with this incident.
I did the publish together with @lucasfcosta and we did use `--tag=canary`, but somehow it got published as latest.
You can read more about this incident [here](https://github.com/chaijs/chai/issues/457#issuecomment-255884038)

keithamus:
Just want to reiterate for crystal clarity - our solution for preventing this in the future will be semantic-release. We're already rolling out semantic-release to sub-projects. All releases will be handled by semantic-releases - by scanning the commit logs or code for what has changed. 

As for backports, I don't think semantic-release has a full solution for backporting - but we haven't backported fixes since I've been here. We effectively (implicitly I suppose) only support the latest version of chai.

ljharb:
Thanks, I think this is answered.

lucasfcosta:
Hi @ljharb, first of all I'd like to apologize for this, I'm sorry this caused you problems :sweat: 

Me and @vieiralucas were extra careful when releasing 4.0.0 and we prepared and tested every single command in a dummy package before running them, but before doing the release itself we ended up adding a new tag on the repo for the canary release and this ended up automatic releasing that version as latest.

As it has already been said, I also agree that manual releases are safer, but given the strict rules we've got on other repos for semantic-release I think this problem can also be solved that way.

[...]

PR #863, opened 2016-11-08
showDiff=true for members, fix #572

### LINKED ISSUE #572: should.have.deep.members diff
I'm new to mocha and chai, so I'm not sure if my issue should be filed against one or the other.
The problem is that I often need to check if two arrays contain same elements up to permutation.
I use `A.should.have.deep.members(B)` for that.
Here are my issues:
1. I'm not sure if this particular combination of `have`, `deep` and `members` is even "official". The documentation at http://chaijs.com/api/bdd/ for `members` does not present this particular combo, but I guess, it's ok?
2. the `mocha -R spec` reporter does not show the diff for this assertions, when they fail. Instead, I get rather not helpful `AssertionError: expected [ Array(3) ] to have the same members as [ Array(2) ]`. 
What would I expect? Something similar to A.sort().should.deep.equal(B.sort()). That is, I'd like to see '-' for each additional member, and '+' for each missing.
3. if this is out of scope of chai project, and should be solved by the mocha test runner, or one of it's reporters, could you please help me, by pointing me in the right direction. Is there a way I could fix this myself?

scottbessler:
As per the discussion in #572, ensuring showDiff=true for members assertions

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -1565,6 +1565,7 @@ module.exports = function (chai, _) {
         , 'expected #{this} to not be a superset of #{act}'
         , obj
         , subset
+        , true
       );
     }
 
@@ -1574,6 +1575,7 @@ module.exports = function (chai, _) {
         , 'expected #{this} to not have the same members as #{act}'
         , obj
         , subset
+        , true
     );
   });
 


meeber:
@scottbessler Do you know if you're still experiencing this issue with Chai's 4.0.0-canary.1 release? Because I suspect #574 already fixed it (by enabling `showDiff` by default unless explicitly disabled by passing in `false`).

scottbessler:
It looks fixed in 4 but don't usually use pre release deps :)

meeber:
Gonna go ahead and close this since it's fixed in the upcoming release. Shouldn't be too long; we don't have any other major changes planned prior to release. Thanks for the PR!

scottbessler:
Doesn't 4.0 have breaking changes? Seems unfortunate to just close when all
this should take is a merge and publish of 3.5.1 Oh well I guess I'll just
point at my repo for now.

lucasfcosta:
Hi, @scottbessler thanks for your PR!
Well, since this has been solved on `4.0.0` only, I think it may be a good idea to start porting fixes to our previous versions too, however, this only happened as it is now because we took too long to release `4.0.0`. Ideally we would have released some intermediary versions with fixes and only then we would release 4.0.0.

I totally understand your point and it does make sense, but let's hear what other team members have to say about it.

keithamus:
We were only discussing this yesterday! My points here https://github.com/chaijs/chai/issues/862#issuecomment-258993968.

Not ideal that we're making such a jump from 3.5 to 4, but in the future it wont be necessary as every PR will be a release. I have no opinion about making effort to backport to 3.x versions.

meeber:
As with most things, the primary cost of backporting fixes to 3.x is time.

The commit that fixed this issue in 4.x (by making `showDiff` default to `true` instead of `false`) is what I'd consider a low-impact breaking change, but a breaking change nevertheless. Therefore, to backport this fix to 3.5, a different fix (such as the one introduced by this PR) would need to be applied instead, thus resulting in divergent branches. Diverging branches doesn't have a noticeable impact if we're just backporting a handful of small fixes, but if the desire is to continue backporting fixes in the future, then the impact of divergent branches is cumulative; it takes increasing amounts of time to apply fixes to increasingly divergent branches.

Consider also that if we wish to start backporting fixes, then we'll need to reconfigure our CI to apply the appropriate node/browser version testing based on Chai version.

There are certainly some low-hanging fixes (such as this one) between 3.5 and 4.0 that could be backported without a huge time investment, but anything beyond that quickly ramps up.

Anyway, I don't see anything wrong with someone volunteering their time to take up that burden as long as it doesn't create the expectation that it must be carried by everyone going forward.

PR #866, opened 2016-11-15
docs: rewrite `.throw` jsdoc

### LINKED ISSUE #864: "expect to throw" not working as expected
Chai version: 3.5.0

New to using Chai so may not be using it correctly, but my "expect .to.throw" tests are not working correctly (contrast with working try/catch version):

```javascript
/*jshint node:true, mocha:true, expr:true*/
'use strict';

var expect = require('chai').expect;

function Foo () {
    this._instance = false;
}

Foo.prototype.init = function () {
    if (this._instance) {
        throw new Error('Already initialised');
    }

    this._instance = true;
};

Foo.prototype.bar = function () {
    if (!this._instance) {
        throw new Error('Not intialised');
    }
};

describe('test - normal', function () {
    var foo = new Foo();

    it('throws exception when not initialised', function () {
        expect(foo.bar).to.throw('Not intialised');
    });

    it('initialise', function () {
        expect(foo.init).to.be.ok;
        expect(foo.bar).to.be.ok;
    });
    
    it('throws exception when re-initialised', function () {
        expect(foo.init).to.throw('Already initialised');
    });
});

describe('test - try/catch', function () {
    var foo = new Foo();

    it('throws exception when not initialised', function () {
        try {
            foo.bar();

        } catch (e) {
            expect(e.message).to.equal('Not intialised');
        }
    });

    it('initialise', function () {
        expect(foo.init).to.be.ok;
        expect(foo.bar).to.be.ok;
    });
    
    it('throws exception when re-initialised', function () {
        try {
            foo.init();

        } catch (e) {
            expect(e.message).to.equal('Already intialised');
        }
    });
});
```

returns:

```
  test - normal
    1) throws exception when not initialised
    ✓ initialise
    2) throws exception when re-initialised

  test - try/catch
    ✓ throws exception when not initialised
    ✓ initialise
    ✓ throws exception when re-initialised


  4 passing (11ms)
  2 failing

  1) test - normal throws exception when not initialised:
     AssertionError: expected [Function] to throw error including 'Not intialised' but got 'Cannot read property \'_instance\' of undefined'
      at Context.<anonymous> (test/testSpec.js:28:33)

  2) test - normal throws exception when re-initialised:
     AssertionError: expected [Function] to throw error including 'Already initialised' but got 'Cannot read property \'_instance\' of undefined'
      at Context.<anonymous> (test/testSpec.js:37:34)
```

meeber:
Closes #864

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -1501,31 +1501,113 @@ module.exports = function (chai, _) {
   Assertion.addMethod('key', assertKeys);
 
   /**
-   * ### .throw(constructor)
+   * ### .throw([errorLike], [errMsgMatcher])
    *
-   * Asserts that the target function will throw a specific error, or specific type of error
-   * (as determined using `instanceof`), optionally with a RegExp or String inclusion test
-   * for the error's message.
-   * If an Error instance is provided, it asserts that the error thrown and the instance
-   * provided are the same.
-   * If an Error constructor is provided, it asserts that the error thrown is an instance
-   * of the constructor provided.
+   * Invokes the target function and asserts that an error is thrown. The
+   * `throw` assertion accepts optional arguments that can be used to perform
+   * additional assertions on the thrown error, as explained below.
    *
-   *     var err = new ReferenceError('This is a bad function.');
-   *     var fn = function () { throw err; }
-   *     expect(fn).to.throw();
-   *     expect(fn).to.throw(ReferenceError);
-   *     expect(fn).to.throw(Error);
-   *     expect(fn).to.throw(/bad function/);
-   *     expect(fn).to.not.throw('good function');
-   *     expect(fn).to.throw(ReferenceError, /bad function/);
-   *     expect(fn).to.throw(err);
+   * If no arguments are provided, `throw` invokes the function and asserts that
+   * an error is thrown.
    *
-   * Furthermore, `throw` changes the context of the assertion (just like `property`)
-   * to the thrown error. This permits for further chainable assertions on the thrown error.
+   *     var badFn = function () { throw new TypeError("Illegal salmon!"); };
+   *     expect(badFn).to.throw();
+   *     var goodFn = function () {};
+   *     expect(goodFn).to.not.throw();
    *
-   *     expect(fn).to.throw(ReferenceError)
-   *       .and.have.property('message').equal('This is a bad function.');
+   * If the first argument is an error constructor, `throw` invokes the function
+   * and asserts that an error is thrown that's an instance of that constructor.
+   *
+   *     var badFn = function () { throw new TypeError("Illegal salmon!"); };
+   *     expect(badFn).to.throw(TypeError);
+   *     expect(badFn).to.not.throw(ReferenceError);
+   *
+   * If the first argument is an error instance, `throw` invokes the function
+   * and asserts that an error is thrown that's referentially equal (===) to
+   * that instance.
+   *
+   *     var err = new TypeError("Illegal salmon!");
+   *     var badFn = function () { throw err; };
+   *     expect(badFn).to.throw(err);
+   *     expect(badFn).to.not.throw(new TypeError("Illegal salmon!"));
+   *
+   * If the first argument is a string, `throw` invokes the function and asserts
+   * that an error is thrown with a message that contains that string.
+   *
+   *     var badFn = function () { throw new TypeError("Illegal salmon!"); };
+   *     expect(badFn).to.throw("salmon");
+   *     expect(badFn).to.not.throw("kangaroo");
+   *
+   * If the first argument is a regular expression, `throw` invokes the function
+   * and asserts that an error is thrown with a message that matches that
+   * regular expression.
+   *
+   *     var badFn = function () { throw new TypeError("Illegal salmon!"); };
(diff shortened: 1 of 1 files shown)


lib/chai/core/assertions.js
```diff
@@ -1501,31 +1501,108 @@ module.exports = function (chai, _) {
...
+   *     var badFn = function () { throw new TypeError("Illegal salmon!"); };
+   *     expect(badFn).to.throw(TypeError);
+   *     expect(badFn).to.not.throw(ReferenceError);
+   *
+   * If the first argument is an error instance, `throw` invokes the function
+   * and asserts that an error is thrown that's referentially equal to that
+   * instance.
+   *
+   *     var err = new TypeError("Illegal salmon!");
+   *     var badFn = function () { throw err };
```
shvaikalesh: `throw err` <-- semicolon here

shvaikalesh:
Fantastic job, @meeber. Very well written.

vieiralucas:
Thank you @meeber. This is really awesome work!

Just for the sake of discussion, this made me realize that this assertion actually checks if anything is thrown not only errors.
```javascript
function stringThrower() {
   throw 'this is a string';
}

expect(stringThrower).to.throw();
```
IMHO this is an ok behavior and the docs can remain as it is.

What do you think?

lib/chai/core/assertions.js
```diff
@@ -1501,31 +1501,108 @@ module.exports = function (chai, _) {
...
-   *
-   *     expect(fn).to.throw(ReferenceError)
-   *       .and.have.property('message').equal('This is a bad function.');
+   * ### .throw([errorLike], [errMsgMatcher])
+   *
+   * Invokes the target function and asserts that an error is thrown, optionally
+   * asserting that the thrown error is an instance of the given error
+   * constructor, or that the thrown error is referentially equal to the given
+   * error instance, and/or that the thrown error's message contains a given
+   * string or matches a given regular expression.
```
keithamus: This is quite a chunky sentence, I'd like to simplify this into a series of shorter statements, maybe even bullets? Maybe something like...

> Invokes the target function, expecting it to throw an error. Passing in an Error constructor as the first argument will assert that the thrown error should match. Passing in an Error instance (`new Error('...')`) will assert that both errors should be referentially equal (`===`). Passing a string will assert that the thrown error's message matches the given string. Passing a regular expression will assert that the thrown error's message matches the given regular expression.

lib/chai/core/assertions.js
```diff
@@ -1501,31 +1501,108 @@ module.exports = function (chai, _) {
...
+   *
+   * If the first argument is an error constructor, `throw` invokes the function
+   * and asserts that an error is thrown that's an instance of that constructor.
+   *
+   *     var badFn = function () { throw new TypeError("Illegal salmon!"); };
+   *     expect(badFn).to.throw(TypeError);
+   *     expect(badFn).to.not.throw(ReferenceError);
+   *
+   * If the first argument is an error instance, `throw` invokes the function
+   * and asserts that an error is thrown that's referentially equal to that
```
keithamus: We should say here: 
> referentially equal (`===`)

lib/chai/core/assertions.js
```diff
@@ -1501,31 +1501,108 @@ module.exports = function (chai, _) {
...
+   * function that relies on `this`). Doing so is problematic because the `this`
+   * context will be lost when the function is invoked by the `throws`
+   * assertion; there's no way for it to know what `this` is supposed to be.
+   * There are two ways around this problem. One solution is to wrap the method
+   * or function call inside of another function. Another solution is to use
+   * `bind`. For example:
+   *
+   *     expect(function () { cat.meow(); }).to.throw();  // Function expression
+   *     expect(() => cat.meow()).to.throw();             // ES6 arrow function
+   *     expect(cat.meow.bind(cat)).to.throw();           // Bind
```
keithamus: This is all _great_. Very clear. Kudos.

keithamus:
Couple of requested changes @meeber - mostly around simplifying the wording, or making sure when we use terms like `referentially equal` we explain what that is. I want to be very very careful about our documentation - that it remains very approachable and we don't need users to learn new concepts or stumble over unknown words without reasonable redundancy in the explanation. Phrases like "referentially equal", while of course are common parlance for experienced devs, can quickly fall apart when explained to less experienced devs (even though they're aware of `===`). Hopefully y'all agree 😄

lib/chai/core/assertions.js
lucasfcosta: +1

lucasfcosta:
Awesome job! I totally agree with the changes @keithamus suggested, specially the first one. Shorter sentences could satisfy people's need to use this without having to read many details.

Thanks @meeber! 😄

meeber:
@keithamus @lucasfcosta Agreed on the first paragraph and the `===`. Fixed both.

@vieiralucas Great point. Actually, I think it's worth mentioning how `throw` handles non-Errors, while also noting that it's a bad practice. Therefore, I've added a new paragraph to the end. Thoughts?

vieiralucas:
Awesome! LGTM

shvaikalesh:
LGTM, very much agreed on specifying handling non-errors.

lucasfcosta:
LGTM too!
Awesome job.

keithamus:
Let's merge this!

PR #880, opened 2016-12-04
(WIP) Implement .{include,contain}.{all,any}.strings

postatum:
Hello dear chai team :)

This is a WIP PR (as suggested by @vieiralucas) that should implement https://github.com/chaijs/chai/issues/858.

This is an initial version just to get your feedback on style, corner cases, tests, etc. Maybe something else needs to be changed to make new assertion work properly.
One thing I know is missing - docstrings for a new method and I also need to check with chai coding style guidelines.
Will be happy to hear your thoughts on this.

Have a good day!

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -1516,7 +1516,9 @@ module.exports = function (chai, _) {
       , mixedArgsMsg = 'when testing strings against strings you must give a single Array|String argument or multiple String arguments';
 
     if (stringsType === 'Array') {
-      if (arguments.length > 1) throw new Error(mixedArgsMsg);
+      if (arguments.length > 1) {
+        throw new Error(mixedArgsMsg);
+      }
     } else {
       strings = Array.prototype.slice.call(arguments)
     }
@@ -1525,7 +1527,9 @@ module.exports = function (chai, _) {
       return this.string(strings[0])
     }
 
-    if (!strings.length) throw new Error('strings required');
+    if (!strings.length) {
+      throw new Error('strings required');
+    }
 
     var len = strings.length
       , any = flag(this, 'any')
@@ -1537,18 +1541,19 @@ module.exports = function (chai, _) {
     }
 
     function containsStr (expectedStr) {
-      return ~obj.indexOf(expectedStr)
+      return obj.indexOf(expectedStr) !== -1;
     }
 
+    var reducer
     // Has any
     if (any) {
-      var reducer = negate ? 'every' : 'some'
-      ok = expected[reducer](containsStr)
+      reducer = negate ? 'every' : 'some';
+      ok = expected[reducer](containsStr);
     }
 
     // Has all
     if (all) {
-      var reducer = negate ? 'some' : 'every'
+      reducer = negate ? 'some' : 'every';
       ok = expected[reducer](containsStr)
     }
 
@@ -1572,7 +1577,7 @@ module.exports = function (chai, _) {
     str = (len > 1 ? 'strings ' : 'string ') + str;
 
     // Have / include
-    str = 'contain ' + str
+    str = 'contain ' + str;
 
     // Assertion
     this.assert(
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -1505,106 +1505,84 @@ module.exports = function (chai, _) {
   Assertion.addMethod('keys', assertKeys);
   Assertion.addMethod('key', assertKeys);
 
-  function assertStrings (keys) {
+  function assertStrings (strings) {
+    _.expectTypes(this, ['string']);
     var obj = flag(this, 'object')
-      , keysType = _.type(keys)
-      , isDeep = flag(this, 'deep')
+      , negate = flag(this, 'negate')
+      , stringsType = _.type(strings)
       , str
       , deepStr = ''
       , ok = true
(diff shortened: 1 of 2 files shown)


keithamus:
One thing I'd like us to consider, is [chai-things](https://github.com/chaijs/chai-things) which adds `any` and `all`. Will this PR impact the behaviour of that plugin?

meeber:
@postatum Thanks so much for working on this! While I do have some specific comments on this PR that I want to share, it made me realize a potentially broader problem we have in Chai. I opened a new issue (#881) that I'd like to discuss some before we go any further into the particulars of this PR.

lib/chai/core/assertions.js
```diff
@@ -1500,6 +1505,85 @@ module.exports = function (chai, _) {
...
+      }
+    } else {
+      str = _.inspect(strings[0]);
+    }
+
+    // Form
+    str = (len > 1 ? 'strings ' : 'string ') + str;
+
+    // Have / include
+    str = 'contain ' + str
```
vieiralucas: Missing semicolon

lib/chai/core/assertions.js
```diff
@@ -1500,6 +1505,85 @@ module.exports = function (chai, _) {
...
+      if (arguments.length > 1) throw new Error(mixedArgsMsg);
+    } else {
+      strings = Array.prototype.slice.call(arguments)
+    }
+
+    if (strings.length === 1) {
+      return this.string(strings[0])
+    }
+
+    if (!strings.length) throw new Error('strings required');
```
vieiralucas: Add curly braces around `if`

lib/chai/core/assertions.js
```diff
@@ -1500,6 +1505,85 @@ module.exports = function (chai, _) {
...
+    var obj = flag(this, 'object')
+      , negate = flag(this, 'negate')
+      , stringsType = _.type(strings)
+      , str
+      , deepStr = ''
+      , ok = true
+      , mixedArgsMsg = 'when testing strings against strings you must give a single Array|String argument or multiple String arguments';
+
+    if (stringsType === 'Array') {
+      if (arguments.length > 1) throw new Error(mixedArgsMsg);
```
vieiralucas: Add curly braces around `if`

lib/chai/core/assertions.js
```diff
@@ -1500,6 +1505,85 @@ module.exports = function (chai, _) {
...
+      , any = flag(this, 'any')
+      , all = flag(this, 'all')
+      , expected = strings;
+
+    if (!any && !all) {
+      all = true;
+    }
+
+    function containsStr (expectedStr) {
+      return ~obj.indexOf(expectedStr)
```
vieiralucas: Not a big fan of bitwise operators.
I prefer the following: 
```javascript
 return obj.indexOf(expectedStr) !== -1
```
I understand that we are using this kind of "trick" everywhere in the codebase, so I can live with that.

Also you're missing a semicolon here.

lib/chai/core/assertions.js
```diff
@@ -1500,6 +1505,85 @@ module.exports = function (chai, _) {
...
+    }
+
+    // Has any
+    if (any) {
+      var reducer = negate ? 'every' : 'some'
```
vieiralucas: Missing semicolon

lib/chai/core/assertions.js
```diff
@@ -1500,6 +1505,85 @@ module.exports = function (chai, _) {
...
+      ok = expected[reducer](containsStr)
```
vieiralucas: Missing semicolon

lib/chai/core/assertions.js
```diff
@@ -1500,6 +1505,85 @@ module.exports = function (chai, _) {
...
+    }
+
+    // Has all
+    if (all) {
+      var reducer = negate ? 'some' : 'every'
```
vieiralucas: Missing semicolon

lib/chai/core/assertions.js
```diff
@@ -1500,6 +1505,85 @@ module.exports = function (chai, _) {
...
+      ok = expected[reducer](containsStr)
```
vieiralucas: Missing semicolon

lib/chai/core/assertions.js
vieiralucas: `reducer` is already declared here, because the `if (any)` above.
I suggest you to move the declaration of this `reducer` variable to before both ifs

postatum:
Hi again everyone :)

I've just fixed all the notes by @vieiralucas.
Sorry it took me so long to fix them - didn't have any time.

meeber:
@postatum thanks for your continued work on this. I suspect we're going to need to change the behavior a bit based on concerns in #881. I have a pretty good sense now on what changes need to be made to make some of these flag behaviors consistent across Chai, but I'm probably not going to be able to get around to articulating them until post-4.0 release. Will circle back to this as soon as I have cycles!

keithamus:
Hi @postatum! Thanks for making this PR. We're currently cleaning house on PRs and issues for housekeeping reasons. We're going to close this PR because - while we agree with the principles of the PR, the code will end up looking slightly different, so we're going to close this as merging it would send a false signal that this might be released soon, when it won't.

Please don't take that as a sign that your contribution wasn't worthwhile! You can see we've already added cards in our roadmap to tackle these: https://github.com/chaijs/chai/projects/2. We hugely appreciate the time you've put into this and please, don't let this dissuade you from making more PRs! I really hope we see more contributions from you 😃

Issue #878, opened 2016-12-02
Question regarding stack trace.


jaridmargolin:
Currently chai implements some magic to hide internal stack traces on `AssertionError`'s. Cool stuffs btw! I am however curious why the stack includes the actual assert implementation. Does it provide additional value to the user? There may be a very obvious reason, so please bare with any possible stupidity on my part :)

Example:

```
AssertionError: expected false to be true
      at Function.assert.isTrue (node_modules/chai/lib/chai/interface/assert.js:332:31)
      at Context.<anonymous> (test/integration/index.js:13:12)
```

99.99% of the time I am looking for:
```
at Context.<anonymous> (test/integration/index.js:13:12)
```

In fact, in the few years I have been using chai, I not once followed a stacktrace to view the internal assert implementation.

---

Just as a little background, I'm working on a small util to pretty up stack traces for a testing tool I am developing:

![https://cldup.com/fovVPUi28W.png](https://cldup.com/fovVPUi28W.png)

keithamus:
Chai's core is pretty solid - we code pretty defensively, but we don't have much influence over plugins, which can cause genuine errors within the assertion. Sometimes it is actually useful to read the code of the assertion that failed. I've certainly - more than a handful of times - read the code of the assertion that's throwing to check what is going wrong. 

Your utility looks super useful - and I've been thinking a lot about doing the exact same thing somehow in chai or perhaps up the stack in the test runner (like mocha). Please let us know when you've finished developing this! I'd love to take a look ❤️

jaridmargolin:
Hmmm ok.

So if I want to display the code frame at the point in which the assert was called  (which I believe is more useful as I write tests), I am going to have to sniff for the chai `AssertError` and shift off the first entry in the call stack.

I see it as the only solution, but it arrises the following issues:
1. If I put it in the core stacktrace styling lib, the library becomes coupled to chai.
2. I could pull the chai specific functionality out into a separate module, but then I don't see it actually ever being implemented.

Ideally I was hoping it could just work out of the box. Ex: `mocha --require flvr`

Am I overlooking any possible solutions?

---

Quick example of what the output looks like after quickly hacking in the `shift`:

![https://cldup.com/vcen70ISmf.png](https://cldup.com/vcen70ISmf.png)

meeber:
@jaridmargolin I'm gonna take a look at this stuff this weekend. Note that there's a related issue in chai@4.0.0-canary.1 with the proxy-related frames showing up in the stack trace.

[...]

Issue #464, opened 2015-06-05
TypeError: 'caller', 'callee', and 'arguments' properties may not be accessed on strict mode functions or the arguments objects for calls to them


moroshko:
chai 3.0.0 doesn't seem to be compatible with babel.

See: http://stackoverflow.com/q/30663989/247243

moroshko:
I updated babel to 5.5.3 and it fixed the issue.

keithamus:
Hey @moroshko thanks for the issue. I was just typing a reply and saw you closed it. Thanks!

frederickfogerty:
Hey, I'm experiencing this error now on babel versions 5.6.x. I just upgraded from 5.6.4 to 5.6.14 and it doesn't fix the issue.

A reproducible case is 
```js
[1].should.have.length(1)
expect([1]).to.have.length(1)
```

I'll have a look at the source but I'm not sure I'll be able to fix this at first glance.

frederickfogerty:
I keep thinking I'm doing something wrong here, as even things like `expect([1].length === 1).to.be.true` is failing

frederickfogerty:
Even things like `assert.strictEqual([1].length, 1)` are failing :cry:

keithamus:
Sounds like that's probably not an issue with Chai. What errors are coming back? Can you compile the babel output and take a look at the source (or push it here).

frederickfogerty:
I'll try make a small gist with the error

frederickfogerty:
So just compiling with babel, it works, but when I bundle it up with webpack, it fails. There is no point posting the whole webpack bundle (8352 lines), but I might making chai et al. external and see if I can get a smaller webpack file

frederickfogerty:
Monkey patching the resulting bundle to look like this: https://gist.github.com/frederickfogerty/114d436f5184a1339bea works, but when I don't external chai, the resulting bundle gives me the error `TypeError: 'caller', 'callee', and 'arguments' properties may not be accessed on strict mode functions or the arguments objects for calls to them`

frederickfogerty:
It looks like the same issue as #384

keithamus:
Yeah, the `"use strict"` problem is because chai _doesn't_ use strict mode, but your compiled output is forcing it into strict mode. The only real solution here is just... don't do that. I'd need to look at your whole build to point out where it's going wrong, but at any rate - this won't be a chai issue.

keithamus:
[...]

PR #884, opened 2016-12-11
Remove proxy frames from stack traces and improve docs/tests

meeber:
This PR is intentionally split into a number of commits, each one of which has a description within the commit message.

Highlights of this PR:

- Remove Chai's internal proxy frames from stack traces of failed property assertions (note: they were already correctly removed from the stack traces of failed method and chainable method assertions)
- Improve tests and inline documentation related to Chai's stack trace wizardry
- Remove dead code related to Chai's stack trace wizardry

Note that this doesn't address #878 regarding extra frames in stack traces when using the `assert` interface.


lib/chai/utils/addChainableMethod.js
```diff
@@ -58,7 +58,7 @@ var call  = Function.prototype.call,
  * @api public
  */
 
-module.exports = function (ctx, name, method, chainingBehavior) {
+module.exports = function addChainableMethod(ctx, name, method, chainingBehavior) {
```
lucasfcosta: I'm a big fan of naming every function even if they're being assigned to a variable.
Bonus style points for this excellent practice.

lib/chai/utils/proxify.js
```diff
@@ -64,7 +64,7 @@ module.exports = function proxify (obj, nonChainableMethodName) {
         }
       }
 
-      return target[property];
+      return Reflect.get(target, property);
```
shvaikalesh: FYI, `Reflect.get` is not different from regular computed access (unless you are using `receiver` parameter).
It also calls `[[Get]]` and `ToPropertyKey`. The only thing they are different is that `Reflect` methods throw on non-object `target` (however `target` here will always be an object). Same goes for `Reflect.has` above: `in` does the same thing. We can simplify checks and code by removing `Reflect` usages.

lucasfcosta: I think that by using `Reflect` we get more semantic code, for me it just feels like a good practice and a more functional way to invoke the language's internal operations.

However, since it won't have any impact and it's just a matter of personal preference I'd be happy with either options.

lib/chai/utils/addChainableMethod.js
```diff
@@ -75,13 +75,18 @@ module.exports = function (ctx, name, method, chainingBehavior) {
...
-          if (old_ssfi)
-            flag(this, 'ssfi', assert);
+        var chainableMethodWrapper = function () {
+          // Use this chainable method wrapper as the starting point for
+          // removing implementation frames from the stack trace of a failed
+          // assertion. Note that this is the correct starting point even if
+          // this assertion has been overwritten since overwriting a chainable
+          // method merely replaces the saved methods in `ctx.__methods` instead
+          // of completely replacing the overwritten assertion.
+          flag(this, 'ssfi', chainableMethodWrapper);
```
meeber: According to [this comment](https://github.com/chaijs/chai/issues/464#issuecomment-116600541), it's "start stack function indicator". Also mentioned on http://chaijs.com/guide/plugins/. I don't feel strongly one way or another about renaming it.

lib/chai/utils/proxify.js
meeber: I don't have much of a preference either way. Anyone have a strong preference here?

lib/chai/utils/overwriteMethod.js
```diff
@@ -52,12 +52,24 @@ module.exports = function (ctx, name, method) {
...
+    // up in the stack trace since it's invoked before this method wrapper.
+    if (!flag(this, 'keep_ssfi')) {
+      flag(this, 'ssfi', overwritingMethodWrapper);
+    }
 
+    // The `keep_ssfi` flag is set so that if this assertion ends up calling
+    // the overwritten assertion, then the overwritten assertion doesn't attempt
+    // to use itself as the starting point for removing implementation frames
+    // from the stack trace of a failed assertion.
     flag(this, 'keep_ssfi', true);
```
lucasfcosta: Let me see if I've got this right, I'm not sure if I fully understand what this does.

So, we use the SSFI flag to store the function which should indicate where the real stack trace will start in order to remove internal implementation details, right?

Whenever an assertion gets overwritten we must turn the `keep_ssfi` flag to true before calling the old assertion (`_super`) in order to avoid that assertion being used as the start of the stack trace. Right?

lib/chai/utils/proxify.js
shvaikalesh: I have an idea how to improve proxies performance by much, utilizing third parameter to `Reflect.get`, so I think we should leave it like this.

PS: much pleasure to review such careful commits 👍

lib/chai/utils/addChainableMethod.js
lucasfcosta: Hmm, since it's pretty specific I think there's nothing we can do about it.
Let's leave it that way.

Maybe a comment indicating what it means somewhere would also be nice, especially if we added a link to that comment, which was really useful for me when understanding how the whole stack trace manipulation works.

lucasfcosta:
This LGTM!

I wasn't really aware of how this whole Stack Trace manipulation worked, so I did a bit of research and I found it to be very interesting.

Just in case anyone wants to read more about it, take a look at the [`AssertionError` constructor](https://github.com/chaijs/assertion-error/blob/master/index.js#L55) and the [`Error.captureStackTrace` docs](https://nodejs.org/api/errors.html#errors_error_capturestacktrace_targetobject_constructoropt).

I made a few comments on the source just to make sure I got it right and added a little consideration to improve the readability of the code regarding the `ssfi` flag we're using, just to make it more clear to new contributors.

Awesome job @meeber! 😄

lib/chai/utils/overwriteMethod.js
meeber: Right, so the function stored in the `ssfi` flag serves as the second parameter for `Error.captureStackTrace`, meaning that the function itself, and all functions called after it (which are just the internals of Chai and plugins), will be removed from the stack trace if an error is thrown. In the case of property assertions, this function is actually the proxy getter, which is why this PR was needed. But in the case of method assertions, the proxy getter function completely returns before the method assertion's function is invoked, so the method assertion's function needs to be stored in the `ssfi` flag instead of the proxy getter.

But what about all functions that came before the function stored in the `ssfi` flag? Usually, the first function that comes before it is the function that the user passed as the second argument to a Mocha `it` invocation. This function contains the user's assertion that failed, so it's good that it's included in the stack trace. But there's a bunch of other functions still in the stack before this one: Mocha's internals. It turns out that these don't appear in the stack traces either because Mocha does its own manual filtering of the stack trace to get rid of their internals. It's important to remember this when troubleshooting an issue with Mocha.

(Note that because the `assert` interface acts as a wrapper around Chai assertions, there's currently an extra function call between the user's assertion and the one stored in the `ssfi` flag, so it shows in the stack trace. That's what #878 is about. It's fixable but will take some work on the `assert` interface.)

As for your final question, yes, the `keep_ssfi` flag is needed so that the overwriting function remains stored in the `ssfi` flag even after it proceeds to call the original function that it overwrote. The first function after the user's assertion is the one that needs to be in the `ssfi` flag in order for this to all work correctly.

meeber:
Pushed another commit:

- Rename third parameter of Assertion constructor from `stack` to
  `ssfi` for consistency's sake.
- Add documentation to Assertion constructor explaining what the `object`,
  `message`, and `ssfi` flags are for.

lucasfcosta:
@meeber awesome job! Chai's code is becoming even more easy and pleasurable to read than it was before with all these useful and well written comments.

I'm in love with this codebase ❤️

lucasfcosta:
Hi friends, sorry for pinging everyone right after holidays, but I've seen we have many open Pull Requests so I thought it would be a good idea for us to start reviewing and approving in order to avoid accumulating too much work.

cc @keithamus @shvaikalesh @vieiralucas

shvaikalesh:
Awesome changes. Makes code so much clearer. LGTM.

keithamus:
@meeber think you could resolve the conflicts for this, so we can merge. All LGTM 😄

meeber:
@keithamus Rebased and resolved conflicts!

PR #889, opened 2016-12-18
Center repo name on README

astorije:
I haven't taken any liberties regarding the name, but I have noticed it is sometimes "chai", "Chai", and I'm pretty sure I have seen "ChaiJS" somewhere. Might want to pick an official representation :)

Before | After
--- | ---
screen shot 2016-12-18 at 00 01 58 | screen shot 2016-12-18 at 00 03 37

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,9 @@
 <h1 align=center>
   <a href="http://chaijs.com" title="Chai Documentation">
-    <img alt="ChaiJS" src="http://chaijs.com/img/chai-logo.png"/> chai
+    <img alt="ChaiJS" src="http://chaijs.com/img/chai-logo.png">
   </a>
+  <br>
+  chai
 </h1>
 
 <p align=center>


keithamus:
LGTM 😀

shvaikalesh:
LGTM, thanks @astorije

lucasfcosta:
LGTM too! Thanks 😄

astorije:
I think I've seen LGTM down more often than not :D

Joke aside, have you thought about using a combination of GitHub reviews + protected branches with required checks? I think that would replace the need for LGTM fairly well.

lucasfcosta:
@astorije I'd love to use that feature, but I think the biggest problem about it is not being able to configure how many approvals something should have before merging.

I think the review feature is awesome, but if a single person approves a change it can already be merged and that's what makes me not want to use it.

But if the rest of the team feels like we should start using it instead of LGTM I'd also be happy with this change.

keithamus:
I think LGTM.co is down for good this time. So I think our hand has been forced. We could take the source for LGTM and host it ourselves but already saying that has exhausted me so I think switching to GitHub's internal review approval system is our only realistic option.

PR #897, opened 2017-01-02
Un-deprecate `length` and add guard

### LINKED ISSUE #684: .length(value) deprecation

meeber:
This PR is broken into two commits. The first commit makes `length` and `lengthOf` true aliases, with documentation updated to remove the deprecation notice and favor `lengthOf` over `length`. The second commit adds a guard against chaining `length` directly off of an uninvoked method assertion.

- Closes #684
- Closes #841

### DIFF
--- a/lib/chai/utils/addLengthGuard.js
+++ b/lib/chai/utils/addLengthGuard.js
@@ -1,0 +1,62 @@
+var config = require('../config');
+
+var fnLengthDesc = Object.getOwnPropertyDescriptor(function () {}, 'length');
+
+/*!
+ * Chai - addLengthGuard utility
+ * Copyright(c) 2012-2014 Jake Luer <jake@alogicalparadox.com>
+ * MIT Licensed
+ */
+
+/**
+ * # addLengthGuard(fn, assertionName, isChainable)
+ *
+ * Define `length` as a getter on the given uninvoked method assertion. The
+ * getter acts as a guard against chaining `length` directly off of an uninvoked
+ * method assertion, which is a problem because it references `function`'s
+ * built-in `length` property instead of Chai's `length` assertion. When the
+ * getter catches the user making this mistake, it throws an error with a
+ * helpful message.
+ *
+ * There are two ways in which this mistake can be made. The first way is by
+ * chaining the `length` assertion directly off of an uninvoked chainable
+ * method. In this case, Chai suggests that the user use `lengthOf` instead. The
+ * second way is by chaining the `length` assertion directly off of an uninvoked
+ * non-chainable method. Non-chainable methods must be invoked prior to
+ * chaining. In this case, Chai suggests that the user consult the docs for the
+ * given assertion.
+ *
+ * If the `length` property of functions is unconfigurable, then return `fn`
+ * without modification.
+ *
+ * Note that in ES6, the function's `length` property is configurable, so once
+ * support for legacy environments is dropped, Chai's `length` property can
+ * replace the built-in function's `length` property, and this length guard will
+ * no longer be necessary. In the mean time, maintaining consistency across all
+ * environments is the priority.
+ *
+ * @param {Function} fn
+ * @param {String} assertionName
+ * @param {Boolean} isChainable
+ * @namespace Utils
+ * @name addLengthGuard
+ */
+
+module.exports = function addLengthGuard (fn, assertionName, isChainable) {
+  if (!fnLengthDesc.configurable) return fn;
+
+  Object.defineProperty(fn, 'length', {
+    get: function () {
+      if (isChainable) {
+        throw Error('Invalid Chai property: ' + assertionName + '.length. Due' +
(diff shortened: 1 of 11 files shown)


shvaikalesh:
Great work, @meeber.

However, I think trapping `length` makes things more complicated than they could be. What about checking `configurable` attr of a function's `"length"` and set up a getter?

meeber:
@shvaikalesh Seems like a good suggestion, even if only for the benefit of "Works down to Node.js 4". Can you elaborate on the third point about optimization?

lucasfcosta:
Awesome work! Thanks @meeber!
As @shvaikalesh suggested, since we're targeting newer environments by fixing this problem with `Proxies` I think that checking for `configurable` `length` properties would be a great idea.

However, due to the fact that older browsers don't have neither `configurable` `length` nor `Proxy` support I think this is a problem we can't really handle without an explicit warning.

meeber:
@lucasfcosta Just to make sure we're on the same page regarding your second point: With this PR, the proxy-based `length` protection would throw a helpful error message in environments that support proxies, but in other environments, it would still throw the somewhat vague error message that it currently does in master. Likewise, if I refactor this PR to instead use @shvaikalesh's suggestion, then the getter-based `length` protection would throw a helpful error message in environments that support configurable `.length` on functions, but in other environments, it would still throw the somewhat vague error message that it currently does in master. So either way, the error messages in the legacy environments aren't improved upon, but at least @shvaikalesh's idea allows more environments to benefit from the helpful error messages.

shvaikalesh:
@meeber 

Right now, we wrap every `Assertion` instance/chainable in proxy. That is kinda expensive. But what more expensive is that every `[[Get]]` getting trapped and invariant-checked.

Now:

Correct method:
`Proxy {expensive} --> Assertion instance --> Assertion.prototype [hit!] -!- Object.prototype -!- null`

Misspelled method:
`Proxy {expensive}[error!] -!- Assertion instance -!- Assertion.prototype [hit!] -!- Object.prototype -!- null`

I suggest:

Correct method:
`Assertion instance --> Assertion.prototype [hit!] -!- Proxy {expensive} -!- Object.prototype -!- null`

meeber:
@shvaikalesh Gotcha. Seems like another strong reason to limit proxy protection to undefined properties.

meeber:
Pushed a new version of this that uses getters instead of proxies.

lib/chai/utils/addLengthGuard.js
```diff
@@ -0,0 +1,65 @@
+var config = require('../config');
+
+var fnDesc = Object.getOwnPropertyDescriptor(function () {}, 'length');
+var isFnLengthConfigurable = typeof fnDesc === 'object'
```
shvaikalesh: Descriptor is guaranteed to be an object here.

lib/chai/utils/addLengthGuard.js
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * @param {Function} fn
+ * @param {String} assertionName
+ * @param {Boolean} isChainable
+ * @namespace Utils
+ * @name addLengthGuard
+ */
+
+module.exports = function addLengthGuard (fn, assertionName, isChainable) {
+  if (isFnLengthConfigurable !== true) return fn;
```
shvaikalesh: Maybe we can do `if (!fnDesc.configurable)` here?

lib/chai/utils/addLengthGuard.js
```diff
@@ -0,0 +1,65 @@
...
+
+  Object.defineProperty(fn, 'length', {
+    get: function () {
+      if (isChainable === true) {
```
shvaikalesh: Maybe `if (isChainable)` suffice?

meeber:
Updated!

test/utilities.js
```diff
@@ -1089,4 +1089,33 @@ describe('utilities', function () {
       }).to.not.throw();
     });
   });
+
+  describe('addLengthGuard', function () {
+    var fnDesc = Object.getOwnPropertyDescriptor(function () {}, 'length');
+    if (typeof fnDesc !== 'object' || fnDesc.configurable !== true) return;
```
shvaikalesh: I've missed this. I believe,  `if (!fnDesc.configurable)` would be enough.

meeber:
Gottem.

shvaikalesh:
Thanks for addressing the issues, @meeber! LGTM.

PS. Maybe we should change PR title to reflect the changes?

keithamus:
This LGTM too. I'll force a merge, LGTM.co is no longer working so we need to disable and find a workable alternative.

Issue #893, opened 2016-12-28
Sanity check for assertion instanceOf


fgarcia:
The current implementation for the `instanceOf` assert does not check if the given constructor is defined

https://github.com/chaijs/chai/blob/104a6b7db9e02d24cd296fbeddc3f55907e8fda0/lib/chai/core/assertions.js#L989

There is a common error I hit several times where an imported variable stops working. I use that variable to assert the instance of an object, like:

```javascript
import { Pathname } from 'core'
....
expect(x).to.be.instanceOf(Pathname)
```

When the import fails (silently) with an undefined value, the error thrown by the assertion shows a normal Javascript error

`TypeError: Cannot read property 'name' of undefined`

It would be great getting something more meaningful like

`ChaiError: instanceOf() needs a constructor. undefined value given`

**NOTE:**

I've realized there is no such thing as `ChaiError` error subtype. I am not sure how far the current code base goes to improve bad use cases

vieiralucas:
Hello @fgarcia thanks for your issue.

IMHO this is a chai problem, I agree with you that this should provide a more meaningful error message.
Since we move the `getName` utility to it's own module ([chaijs/get-func-name](https://github.com/chaijs/get-func-name)), currently on `4.0.0-canary.1` this error would be throw from [here](https://github.com/chaijs/get-func-name/blob/b4a20e94373a86822e317ad510cd259592fce043/index.js#L33).

So, IMHO:
 -  `get-func-name` should throw its own exception if `undefined` is given. Maybe we should open an issue there
 - `instanceOf` should do something similar to [above](https://github.com/chaijs/chai/blob/104a6b7db9e02d24cd296fbeddc3f55907e8fda0/lib/chai/core/assertions.js#L707)

It should be very easy to do this, we just need to add a check [here](https://github.com/chaijs/chai/blob/104a6b7db9e02d24cd296fbeddc3f55907e8fda0/lib/chai/core/assertions.js#L990)

Am I missing something here? Do someone disagree?

lucasfcosta:
Hi @fgarcia, thanks for your issue! 😄 

Well, I totally agree with @vieiralucas that we should fix this.

However, I think this is something that should be done within Chai's core.

IMO the whole point of having a separate module for `get-func-name` is making it modular enough to be used by anyone wanting to, including us. By throwing an error there we would be coupling what we want for `Chai` with the module's own purpose.

[...]

PR #899, opened 2017-01-11
Throw error when instanceof is passed something that is not a function as constructor

lucasfcosta:
This aims to solve #893 and chaijs/get-func-name#20 is related to this issue.

Whenever passed something that is not an instance of a function, the `instanceof` assertion will now throw an error informing the user that he should pass a constructor to this assertion, but instead he has passed `<type>`

I also added tests for this in all three interfaces.

As I've said on the other PR an `instanceof` check is enough to determine whether or not something is a function, even though the `@@hasInstance` symbol exists.

Please let me know if I forgot anything.

------------------

EDIT: I have also started using the `typeof` check due to [this explanation](https://github.com/chaijs/get-func-name/pull/20#discussion_r95651819) by @shvaikalesh.

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -988,9 +988,29 @@ module.exports = function (chai, _) {
 
   function assertInstanceOf (constructor, msg) {
     if (msg) flag(this, 'message', msg);
+
+    var target = flag(this, 'object')
+    var validInstanceOfTarget = constructor === Object(constructor) && (
+        typeof constructor === 'function' ||
+        (typeof Symbol !== 'undefined' &&
+         typeof Symbol.hasInstance !== 'undefined' &&
+         Symbol.hasInstance in constructor)
+    );
+
+    if (!validInstanceOfTarget) {
+      var constructorType = constructor === null ? 'null' : typeof constructor;
+      throw new Error('The instanceof assertion needs a constructor but ' + constructorType + ' was given.');
+    }
+
+    var isInstanceOf = target instanceof constructor
+
     var name = _.getName(constructor);
+    if (name === null) {
+        name = 'an unnamed constructor';
+    }
+
     this.assert(
-        flag(this, 'object') instanceof constructor
+        isInstanceOf
       , 'expected #{this} to be an instance of ' + name
       , 'expected #{this} to not be an instance of ' + name
     );
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "assertion-error": "^1.0.1",
     "check-error": "^1.0.1",
     "deep-eql": "^2.0.1",
-    "get-func-name": "^1.0.0",
+    "get-func-name": "^2.0.0",
     "pathval": "^1.0.0",
     "type-detect": "^4.0.0"
   },
--- a/test/assert.js
+++ b/test/assert.js
@@ -141,6 +141,60 @@ describe('assert', function () {
     function Foo(){}
     assert.instanceOf(new Foo(), Foo);
 
+    err(function(){
+      assert.instanceOf(new Foo(), 1);
+    }, "The instanceof assertion needs a constructor but number was given.");
+
+    err(function(){
+      assert.instanceOf(new Foo(), 'batman');
+    }, "The instanceof assertion needs a constructor but string was given.");
+
+    err(function(){
+      assert.instanceOf(new Foo(), {});
+    }, "The instanceof assertion needs a constructor but object was given.");
+
+    err(function(){
+      assert.instanceOf(new Foo(), true);
+    }, "The instanceof assertion needs a constructor but boolean was given.");
+
+    err(function(){
+      assert.instanceOf(new Foo(), null);
+    }, "The instanceof assertion needs a constructor but null was given.");
+
+    err(function(){
+      assert.instanceOf(new Foo(), undefined);
+    }, "The instanceof assertion needs a constructor but undefined was given.");
+
+    var expectedError;
+    try {
+      t instanceof Thing;
+    } catch (err) {
+      errMsg = '[object Object] instanceof function Thing(){} failed: ' + err.message + '.';
+    }
+
+    err(function(){
+      function Thing(){};
+      var t = new Thing();
+      Thing.prototype = 1337;
+      assert.instanceOf(t, Thing);
+    }, expectedError);
+
+    if (typeof Symbol !== 'undefined' && typeof Symbol.hasInstance !== 'undefined') {
+        err(function(){
+          assert.instanceOf(new Foo(), Symbol());
+        }, "The instanceof assertion needs a constructor but symbol was given.");
+
+        err(function() {
+            var FakeConstructor = {};
+            var fakeInstanceB = 4;
+            FakeConstructor[Symbol.hasInstance] = function (val) {
+                return val === 3;
+            };
+
+            assert.instanceOf(fakeInstanceB, FakeConstructor);
+        }, 'expected 4 to be an instance of an unnamed constructor')
+    }
+
     err(function () {
       assert.instanceOf(5, Foo);
     }, "expected 5 to be an instance of Foo");
@@ -155,6 +209,47 @@ describe('assert', function () {
   it('notInstanceOf', function () {
     function Foo(){}
     assert.notInstanceOf(new Foo(), String);
+
+    err(function(){
+      assert.notInstanceOf(new Foo(), 1);
+    }, "The instanceof assertion needs a constructor but number was given.");
+
+    err(function(){
+      assert.notInstanceOf(new Foo(), 'batman');
+    }, "The instanceof assertion needs a constructor but string was given.");
+
+    err(function(){
+      assert.notInstanceOf(new Foo(), {});
+    }, "The instanceof assertion needs a constructor but object was given.");
+
+    err(function(){
+      assert.notInstanceOf(new Foo(), true);
+    }, "The instanceof assertion needs a constructor but boolean was given.");
+
+    err(function(){
+      assert.notInstanceOf(new Foo(), null);
+    }, "The instanceof assertion needs a constructor but null was given.");
+
+    err(function(){
+      assert.notInstanceOf(new Foo(), undefined);
+    }, "The instanceof assertion needs a constructor but undefined was given.");
+
+
+    if (typeof Symbol !== 'undefined' && typeof Symbol.hasInstance !== 'undefined') {
+        err(function(){
+          assert.notInstanceOf(new Foo(), Symbol());
+        }, "The instanceof assertion needs a constructor but symbol was given.");
+
+        err(function() {
+            var FakeConstructor = {};
+            var fakeInstanceB = 4;
+            FakeConstructor[Symbol.hasInstance] = function (val) {
+                return val === 4;
+            };
+
+            assert.notInstanceOf(fakeInstanceB, FakeConstructor);
+        }, 'expected 4 to not be an instance of an unnamed constructor');
+    }
 
     err(function () {
       assert.notInstanceOf(new Foo(), Foo);
--- a/test/expect.js
+++ b/test/expect.js
@@ -72,61 +72,61 @@ describe('expect', function () {
           expect(42).ok.pizza;
         }, 'Invalid Chai property: pizza');
       });
-  
+
       it('throws when invalid property follows overwritten property assertion', function () {
         err(function () {
           expect(42).tmpProperty.pizza;
         }, 'Invalid Chai property: pizza');
       });
-  
+
       it('throws when invalid property follows uncalled method assertion', function () {
         err(function () {
           expect(42).equal.pizza;
         }, 'Invalid Chai property: equal.pizza. See docs for proper usage of "equal".');
       });
-  
+
       it('throws when invalid property follows called method assertion', function () {
         err(function () {
           expect(42).equal(42).pizza;
         }, 'Invalid Chai property: pizza');
       });
-  
+
       it('throws when invalid property follows uncalled overwritten method assertion', function () {
         err(function () {
           expect(42).tmpMethod.pizza;
         }, 'Invalid Chai property: tmpMethod.pizza. See docs for proper usage of "tmpMethod".');
       });
-  
+
       it('throws when invalid property follows called overwritten method assertion', function () {
         err(function () {
           expect(42).tmpMethod().pizza;
         }, 'Invalid Chai property: pizza');
       });
-  
+
       it('throws when invalid property follows uncalled chainable method assertion', function () {
         err(function () {
           expect(42).a.pizza;
         }, 'Invalid Chai property: pizza');
       });
-  
+
       it('throws when invalid property follows called chainable method assertion', function () {
         err(function () {
           expect(42).a('number').pizza;
         }, 'Invalid Chai property: pizza');
       });
-  
+
       it('throws when invalid property follows uncalled overwritten chainable method assertion', function () {
        err(function () {
           expect(42).tmpChainableMethod.pizza;
         }, 'Invalid Chai property: pizza');
       });
-  
+
       it('throws when invalid property follows called overwritten chainable method assertion', function () {
         err(function () {
           expect(42).tmpChainableMethod().pizza;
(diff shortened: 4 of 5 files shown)


keithamus:
Despite this throwing an error, I assume it's not a breaking change? Just makes the error message a user would receive more descriptive?

lucasfcosta:
@keithamus It was already throwing an error before, which was:

`Right-hand side of 'instanceof' is not an object`

I think it's important to tell the user which type was passed in order for him to be able to fix what went wrong. Also, whenever passing `undefined` we had problems with `getName` due to trying to access the non-existing `name` property of undefined.

lib/chai/core/assertions.js
```diff
@@ -988,6 +988,13 @@ module.exports = function (chai, _) {
 
   function assertInstanceOf (constructor, msg) {
     if (msg) flag(this, 'message', msg);
+    var isFunction = constructor instanceof Function;
+
+    if (!isFunction) {
```
shvaikalesh: I think we can do

```js
var type = typeof constructor;
if (typeof !== 'function') {
    throw new Error
}
```

Maybe instead of `var type = constructor === null ? 'null' : typeof constructor;` we can use `type-detect` here?

lucasfcosta: I think that it would be an overkill. In this case we just need to know whether something is a function or not in order to see if it is a constructor, there's no need for anything more complex than that.

Please let me know if I forgot any edge case, but I think we should go native if we can.

shvaikalesh: Oops, speaking of edge cases, `typeof` is not sufficient here.

`instanceof` works not only with functions, but with any objects that have `@@hasInstance`.

lib/chai/core/assertions.js
```diff
@@ -988,6 +988,12 @@ module.exports = function (chai, _) {
 
   function assertInstanceOf (constructor, msg) {
     if (msg) flag(this, 'message', msg);
+    var constructorType = constructor === null ? 'null' : typeof constructor;
```
shvaikalesh: The most hardcore option is

```js

  function assertInstanceOf (constructor, msg) {
    if (constructor !== Object(constructor)) throw new TypeError();
    var object = flag(this, 'object');
    var hasInstance = constructor[Symbol.hasInstance];
    var isInstance;
    if (hasInstance == null) {
        if (typeof constructor !== 'function') throw new TypeError();
        isInstance = Function[Symbol.hasInstance].call(constructor, object);
    } else {
        isInstance = object instanceof constructor;
    }

    var name = _.getName(constructor);
    this.assert(isInstance
      , 'expected #{this} to be an instance of ' + name
      , 'expected #{this} to not be an instance of ' + name
    );
  };
```

shvaikalesh: I've reimplemented part of `instanceof` algo to avoid getting `@@hasInstance`  twice.

shvaikalesh: More concise option

```js
  function assertInstanceOf (constructor, msg) {
    var validInstanceOfTarget = constructor === Object(constructor) && (
        typeof constructor == 'function' ||
        typeof constructor[Symbol.hasInstance] == 'function'
    );
    if (!validInstanceOfTarget) throw new TypeError();

    var name = _.getName(constructor);
    this.assert(
        flag(this, 'object') instanceof constructor
      , 'expected #{this} to be an instance of ' + name
      , 'expected #{this} to not be an instance of ' + name
    );
  };
```

But this one calls `constructor.[[Get]](@@hasInstance)` twice. `[[Get]]` may be observable.

meeber:
Agreed with @shvaikalesh. Whenever `constructor` has a `Symbol.hasInstance` property defined, this assertion should respect whatever `instanceof` returns, even if `constructor` isn't a function. However, the failed assertion message needs to be able to handle the fact that the return value of `getName` would be `null` in such a situation.

Example:

```js
var FakeConstructor = {};  // Not a function, but it has `Symbol.hasInstance` defined:
FakeConstructor[Symbol.hasInstance] = function (val) {
  return val === 3;
};

var fakeInstanceA = 3;

// Should pass even though FakeConstructor isn't a function
expect(fakeInstanceA).to.be.an.instanceof(FakeConstructor);

var fakeInstanceB = 4;

// Should fail because 4 !== 3, but what does failed assertion error message use as "name"?
expect(fakeInstanceB).to.be.an.instanceof(FakeConstructor);
```

shvaikalesh:
Yet another thing to consider: if `constructor` have no `.prototype` (method or `=>`), `instanceof` will fail with `TypeError`. We can't simply check for `.prototype`, because non-generator bound functions won't have one, but would work with `instanceof` because of internal slots.

keithamus:
Good points from @shvaikalesh - we should handle all these cases. I suppose we could just catch any `TypeError`s thrown by `instanceof` and make them friendlier?

lucasfcosta:
@shvaikalesh I'm not sure I get it. Can't we just use `Object.getPrototypeOf` to see if something has a `prototype`? I've just tested this with both a `method` and an arrow function and it worked.

Can you please describe with more detail?

Also, thanks everyone for your feedback!

shvaikalesh:
@lucasfcosta We can check for `.prototype` on function, however lack of it does not necessarily mean `instanceof` will wail: when function is bound with `Function#bind`, it is created without `.prototype`, but still works with `instanceof` because of `[[BoundTargetFunction]]`.


With all this, I agree with @keithamus that the most simple, reliable and future-compatible way is to wrap `instanceof` in `try`/`catch` and improve error message.

lucasfcosta:
Thanks for the detailed explanation! Any suggestions on what the error message should be?

I was reading the specs and according to the algorithm I can only think of:

> The provided argument does not have a prototype.

But I think that's not explicit enough and I'm not sure it is a good idea to talk about internal implementation details on error messages.

Please let me know if you can think of anything better.

Also, if anyone wants to read about how `instanceof` works, [here is the link](https://tc39.github.io/ecma262/#sec-instanceofoperator).

shvaikalesh:
`instanceof` may fail, if
+ `constructor` is not an Object
+ `@@hasInstance` getter threw
+ `constructor` has `!= null` and not callable `@@hasInstance`
+ `@@hasInstance` threw
+ `constructor` is not callable
+ `.prototype` is not an Object
+ `[[GetPrototypeOf]]` threw anywhere in prototype chain (proxies, ya know)

Many reasons here; we would have to reimplement `instanceof` to cover all of them. I believe `${instance} instanceof ${constructor} failed: ${originalErrorMessage}` is the way to go.

lucasfcosta:
Well, we already cover some of these cases such in our `validInstanceOfTarget` check. That's why I thought of a more concise error message.

However, your suggestions look awesome. Thanks!

lucasfcosta:
Hey friends, here it is. I also added a test to check if the error messages are being passed to our own error when `instanceof` throws.

Let me know if you have any further suggestions.

lib/chai/core/assertions.js
```diff
@@ -988,9 +988,28 @@ module.exports = function (chai, _) {
...
+    }
+
+    var isInstanceOf;
+    try {
+        isInstanceOf = target instanceof constructor
+    } catch (err) {
+        throw new Error(target + ' instanceof ' + constructor + ' failed: ' + err.message + '.')
+    }
+
     var name = _.getName(constructor);
```
meeber: Slight issue with error message wording in rare circumstances.

Example:

```js
var FakeConstructor = {};
FakeConstructor[Symbol.hasInstance] = function (val) {
  return val === 3;
};

var fakeInstanceB = 4;

expect(fakeInstanceB).to.be.an.instanceof(FakeConstructor);
```

Current output:

> AssertionError: expected 4 to be an instance of undefined

Output after https://github.com/chaijs/get-func-name/pull/20 is merged:

> AssertionError: expected 4 to be an instance of null

I'm thinking that if the result of `getName` is `null`, then the best thing we can do is just print out the `constructor` object; `null` will be meaningless/misleading to the end user. In this case, that'd be:

> AssertionError: expected 4 to be an instance of [object Object]

Not particularly helpful, but at least it's not misleading. An alternative would be to say "of an unnamed constructor".

> AssertionError: expected 4 to be an instance of an unnamed constructor

lucasfcosta: Well noticed!
This change is ready. I'll push it to this branch as soon as we release `get-func-name` with that new change so we can guarantee our travis builds will succeed.

lucasfcosta:
This has just been updated to include the changes you requested. Now everything should be working as expected and this also includes the new version of get-func-name, which has just been released.

lib/chai/core/assertions.js
```diff
@@ -988,9 +988,32 @@ module.exports = function (chai, _) {
 
   function assertInstanceOf (constructor, msg) {
     if (msg) flag(this, 'message', msg);
+
+    var target = flag(this, 'object')
+    var validInstanceOfTarget = constructor === Object(constructor) && (
+        typeof constructor === 'function' ||
+        (typeof Symbol !== 'undefined' && typeof constructor[Symbol.hasInstance] === 'function')
```
meeber: Are there any edge cases (environments or polyfills) in which `Symbol` is defined but `Symbol.hasInstance` isn't? Safe thing to do here is check that it isn't undefined either.

lucasfcosta: Good catch!

meeber:
Looks like Node v4 has failing tests.

lucasfcosta:
@meeber done!
This happened because Node v4 didn't implement `Symbol.hasInstance` but did implement `Symbol`. I added this new check before running tests on environment that do not support it.

lib/chai/core/assertions.js
```diff
@@ -988,9 +988,33 @@ module.exports = function (chai, _) {
...
+    if (!validInstanceOfTarget) {
+      var constructorType = constructor === null ? 'null' : typeof constructor;
+      throw new Error('The instanceof assertion needs a constructor but ' + constructorType + ' was given.');
+    }
+
+    var isInstanceOf;
+    try {
+        isInstanceOf = target instanceof constructor
+    } catch (err) {
+        throw new Error(target + ' instanceof ' + constructor + ' failed: ' + err.message + '.')
```
shvaikalesh: Stack trace of original error message is lost here.

lib/chai/core/assertions.js
```diff
@@ -988,9 +988,33 @@ module.exports = function (chai, _) {
...
+    var target = flag(this, 'object')
+    var validInstanceOfTarget = constructor === Object(constructor) && (
+        typeof constructor === 'function' ||
+        (typeof Symbol !== 'undefined' && typeof Symbol.hasInstance !== 'undefined' &&
+         typeof constructor[Symbol.hasInstance] === 'function')
+    );
+
+    if (!validInstanceOfTarget) {
+      var constructorType = constructor === null ? 'null' : typeof constructor;
+      throw new Error('The instanceof assertion needs a constructor but ' + constructorType + ' was given.');
```
shvaikalesh: `TypeError` fits better

lib/chai/core/assertions.js
```diff
@@ -988,9 +988,33 @@ module.exports = function (chai, _) {
 
   function assertInstanceOf (constructor, msg) {
     if (msg) flag(this, 'message', msg);
+
+    var target = flag(this, 'object')
+    var validInstanceOfTarget = constructor === Object(constructor) && (
+        typeof constructor === 'function' ||
+        (typeof Symbol !== 'undefined' && typeof Symbol.hasInstance !== 'undefined' &&
+         typeof constructor[Symbol.hasInstance] === 'function')
```
shvaikalesh: `[[Get]]` is called on `@@hasInstance` twice: here and in `instanceof`. What if it is observable?

lib/chai/core/assertions.js
meeber: After this was merged, I was gonna rebase #922 and turn this error into an `AssertionError` that properly retains `ssfi` (and then a follow up PR will properly retain the user's custom message).

lib/chai/core/assertions.js
meeber: I realize we're going a bit full-circle here, but after further consideration I think that when the user's code itself is potentially responsible for the error being thrown, we should just pass along that error (original error object, stack trace, and message) instead of attempting to make it friendlier.

Conversely, when it's Chai that is purposefully throwing the error, then it needs to be an `AssertionError` that properly passes along the `ssfi` flag as well as respecting the user's custom failed assertion message. #922 handles the `ssfi` part, and a small follow-up PR will handle the custom message part.

In this particular case, the user could potentially have some custom crap defined via `@@hasInstance` which is responsible for an error being thrown. Therefore, I think we should pass along that error object as-is, which will be done automatically if we just remove the whole `try`/`catch`.

lib/chai/core/assertions.js
meeber: I guess the question is: Is there an implicit understanding that Chai, in the process of performing an `instanceof` assertion, will only trigger `@@hasInstance` once? Or is it allowed to trigger it as much as it wants?

The benefit of not worrying about calling it multiple times is that the code is less complex. The cost is that to test an object with an observable `@@hasInstance`, you'll have to manually use JavaScript's `instanceof` and then assert on the result of that, rather than using Chai's builtin `instanceof` assertion.

My gut feeling is that having an observable `@@hasInstance` is such an extreme edge case that it's not worth Chai jumping through any hoops to accommodate, especially given the easy workaround of invoking `instanceof` manually and asserting on the result. But it doesn't add too much complexity to accommodate it either, so I won't protest if we want to do it.

Anyone have strong feelings one way or another?

lucasfcosta: I think that if it's easier to accomodate that we really should do it.

It's better to perform this work once and in our side than passing the burden of having to work around this to our users every time they need it.

lib/chai/core/assertions.js
lucasfcosta: As @meeber noticed, I also think we're going a bit full-circle here. The whole purpose of this PR was to give our users more friendly error messages, but given what was said I'm not sure this is the best choice anymore.

So, what do you think? Should we cancel the `Error` related change and keep the other improvements or should we keep it?

I'd like to hear more opinions about that.

lib/chai/core/assertions.js
meeber: Sounds good to me. See https://github.com/chaijs/chai/pull/899#discussion_r95682203 for ideas on only triggering `@@hasInstance` once.

lib/chai/core/assertions.js
meeber: To be clear, I like all the changes except the part where we catch an error that could possibly have been thrown by the user's code (if they had a custom `@@hasInstance` that was responsible for the error). If we remove that change but keep the rest, I think we still come out ahead, and the changes we keep still satisfy the original problem from #893.

lib/chai/core/assertions.js
lucasfcosta: @meeber I can trigger the getter for @@hasInstance zero times if I use the `in` operator.

I also removed the `try/catch` statements on this commit.

What do you think about this?

Is there anything I should improve?

lib/chai/core/assertions.js
```diff
@@ -988,9 +988,29 @@ module.exports = function (chai, _) {
...
+         typeof Symbol.hasInstance !== 'undefined' &&
+         Symbol.hasInstance in constructor)
+    );
+
+    if (!validInstanceOfTarget) {
+      var constructorType = constructor === null ? 'null' : typeof constructor;
+      throw new Error('The instanceof assertion needs a constructor but ' + constructorType + ' was given.');
+    }
+
+    var isInstanceOf = target instanceof constructor
```
meeber: Semi-colon :D

lib/chai/core/assertions.js
meeber: Using the `in` operator seems like a good strategy; that way, `@@hasInstance` is only triggered once when the actual `instanceof` call is made.

lib/chai/core/assertions.js
lucasfcosta: Thanks!
I miss linting tools 😅

keithamus:
Agreed, we need to get some of these outstanding PRs pushed in before 4.0 final, otherwise they'll only be held up for many more months before 5.0

PR #900, opened 2017-01-11
Refactor `utils.addChainableMethod` helper

shvaikalesh:
### Use `Object.setPrototypeOf` instead of `__proto__`

* `Object.setPrototypeOf` is more strict than `__proto__`: it throws on failure instead of being silent.
* `__proto__` is in Annex B of the spec; using it is kinda discouraged.

> This annex describes various legacy features and other characteristics of web browser based ECMAScript implementations. All of the language features and behaviours specified in this annex have one or more undesirable characteristics and in the absence of legacy usage would be removed from this specification. 

* I would avoid relying on early implementations of `__proto__` (before it was defined on `Object.prototype`) .
* `Object.setPrototypeOf` reads better :smile:.
* Every browser we support, has either both `__proto__` and `Object.setPrototypeOf` or neither of them.

When we drop IE < 11, we would simplify `addChainableMethod` even more.

### Remove hardcoded non-configurable function keys

They used to cause troubles in ancient IEs, but work fine in IE.

### DIFF
--- a/lib/chai/utils/addChainableMethod.js
+++ b/lib/chai/utils/addChainableMethod.js
@@ -18,13 +18,13 @@ var transferFlags = require('./transferFlags');
  * Module variables
  */
 
-// Check whether `__proto__` is supported
-var hasProtoSupport = '__proto__' in Object;
 
 // Without `__proto__` support, this module will need to add properties to a function.
 // However, some Function.prototype methods cannot be overwritten,
 // and there seems no easy cross-platform way to detect them (@see chaijs/chai/issues/69).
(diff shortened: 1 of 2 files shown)


meeber:
LGTM

shvaikalesh:
I've updated the PR with perf improvement:

We get list of non-configurable props on functions only once and then only `indexOf` in hot code.
Internet Explorers would love that.

meeber:
Seems like a good improvement. Still LGTM.

vieiralucas:
LGTM

PR #901, opened 2017-01-11
chore: update supported Node versions

meeber:
- Dropped support for Node v0.12
- Made Node v4 the minimum required version
- Added Node v7 to travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,9 @@ language: node_js
 addons:
   sauce_connect: true
 node_js:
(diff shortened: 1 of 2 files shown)


shvaikalesh:
LGTM

vieiralucas:
LGTM

keithamus:
So btw, I've disabled LGTM as it's been sunset, and switched to the github review system. It means we only need one reviewer before a merge and there's no way to configure that, but it actually works whereas LGTM no longer does.

Issue #919, opened 2017-01-27
`.not.have.keys` incorrectly failing


meeber:
Because of [this line](https://github.com/chaijs/chai/blob/a64eb623449f058143930bd3a9d943985afa13b6/lib/chai/core/assertions.js#L1464) (I presume), the following tests both fail:

```js
expect({a: 1, b: 2, c: 3}).to.have.keys('a', 'b');  // Correctly fails
expect({a: 1, b: 2, c: 3}).to.not.have.keys('a', 'b'); // Should pass but it fails too
```

The second test is incorrectly failing; it should pass right away on the basis that the target and given values don't have the same number of keys, which is a requirement when the `contains` flag isn't set.

leggsimon:
Personally I would say that semantically both of those seem like they should fail. One of the best things about Chai is it's human readability. To which both of these seem wrong.
-------
Given the examples above
```js
expect({a: 1, b: 2, c: 3}).to.have.keys('a', 'b'); // I would've expected to pass
```
I would have expected to pass. To have the behaviour that I guess is the correct behaviour for that assertion (failing) I would expect something like
```js
expect({a: 1, b: 2, c: 3}).to.have.exact.keys('a', 'b'); 
// or
expect({a: 1, b: 2, c: 3}).to.only.have.keys('a', 'b'); 
```
Semantically I would expect both of these to fail when reading them.

----------
Likewise, the second one, semantically, I think **should** fail,
`{a: 1, b: 2, c: 3}` _has_ got keys `'a'` & `'b'`.

meeber:
@leggsimon Yeah, we have a bit of a problem right now with some of our list-like assertions. The default behavior is inconsistent among them, and there's also inconsistency in how they interact with the `contains`, `any`, and `all` flags. A detailed description of `.keys` current behavior is documented [here](https://github.com/meeber/chai/blob/ed4fd84c95e7a33841ce985db812e7aa448b89f4/lib/chai/core/assertions.js#L2026-L2132), and #881 goes into more details about some of the inconsistencies among different assertions.

Right now, the problem described in this issue is considered a bug because it doesn't match the currently documented behavior of `.keys`. It needs to be fixed in the short term.

In the long term, a major change needs to be made to bring all of these related assertions into a state of consistency. I agree with you that given the two options below, the first seems more semantic:

1. Default behavior is that the target keys can be a superset of the given keys. Adding the `.exact` or `.only` flag makes it so that target keys and given keys must be identical sets.

[...]

PR #920, opened 2017-01-27
Overhaul bdd documentation

### LINKED ISSUE #902: Documentation error
```
.property(name, [value])

@param { String } name
@param { Mixed } value (optional)
@param { String } message _optional_
```

... so can I attach a message, or can't I? I seem to be having trouble getting a message to appear so that's probably a no?

### LINKED ISSUE #914: Documentation error
http://chaijs.com/api/bdd/ 

`.ownPropertyDescriptor(name[, descriptor[, message]])`

has an assertion

`expect('test').ownPropertyDescriptor('length').to.have.keys('value');`

which fails like this:
```
AssertionError: expected { Object (value, writable, ...) } to have key 'value'
+ expected - actual

 [
-  "configurable"
-  "enumerable"
   "value"
-  "writable"
 ]
```

> Note, either any or all should be used in the assertion. If neither are used, the assertion is defaulted to all.
>When both all and have are used, the target object must both contain all of the passed-in keys AND the number of keys in the target object must match the number of keys passed in (in other words, a target object must have all and only all of the passed-in keys).

to pass the assertion should be changed to
`expect('test').ownPropertyDescriptor('length').to.contain.keys('value');`
or
`expect('test').ownPropertyDescriptor('length').to.have.any.keys('value');`
or
`expect('test').ownPropertyDescriptor('length').to.include.keys('value');`

meeber:
Note to self: Rebase once #947 is merged and update docs for `by` since it now accepts a message argument.

There are two commits in this PR.

The first commit is a small change that renames the `target` argument to `subject` in a few assertions because "target" is an overloaded term which made it difficult to document.

The second commit is a huge overhaul of the bdd documentation intended to bring clarity, consistency, and guidance to every assertion. All of the examples that are intended to pass do so with the exception of two due to #919.

Closes #914
Closes #902

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -2153,14 +2153,14 @@ module.exports = function (chai, _) {
    * @name change
    * @alias changes
    * @alias Change
-   * @param {String} target
+   * @param {String} subject
    * @param {String} property name _optional_
    * @param {String} message _optional_
    * @namespace BDD
    * @api public
    */
 
-  function assertChanges (target, prop, msg) {
+  function assertChanges (subject, prop, msg) {
     if (msg) flag(this, 'message', msg);
     var fn = flag(this, 'object')
       , flagMsg = flag(this, 'message')
@@ -2169,16 +2169,16 @@ module.exports = function (chai, _) {
 
     var initial;
     if (!prop) {
-      new Assertion(target, flagMsg, ssfi, true).is.a('function');
-      initial = target();
+      new Assertion(subject, flagMsg, ssfi, true).is.a('function');
+      initial = subject();
     } else {
-      new Assertion(target, flagMsg, ssfi, true).to.have.property(prop);
-      initial = target[prop];
+      new Assertion(subject, flagMsg, ssfi, true).to.have.property(prop);
+      initial = subject[prop];
     }
 
     fn();
 
-    var final = prop === undefined || prop === null ? target() : target[prop];
+    var final = prop === undefined || prop === null ? subject() : subject[prop];
     var msgObj = prop === undefined || prop === null ? initial : '.' + prop;
 
     // This gets flagged because of the .by(delta) assertion
@@ -2199,7 +2199,7 @@ module.exports = function (chai, _) {
   Assertion.addMethod('changes', assertChanges);
 
   /**
-   * ### .increase(target[, property[, message]])
+   * ### .increase(subject[, property[, message]])
    *
    * Asserts that a function increases a numeric object property.
    *
@@ -2217,14 +2217,14 @@ module.exports = function (chai, _) {
    * @name increase
    * @alias increases
    * @alias Increase
-   * @param {String|Function} target
+   * @param {String|Function} subject
    * @param {String} property name _optional_
    * @param {String} message _optional_
    * @namespace BDD
    * @api public
    */
 
-  function assertIncreases (target, prop, msg) {
+  function assertIncreases (subject, prop, msg) {
     if (msg) flag(this, 'message', msg);
     var fn = flag(this, 'object')
       , flagMsg = flag(this, 'message')
@@ -2233,11 +2233,11 @@ module.exports = function (chai, _) {
 
     var initial;
     if (!prop) {
-      new Assertion(target, flagMsg, ssfi, true).is.a('function');
-      initial = target();
+      new Assertion(subject, flagMsg, ssfi, true).is.a('function');
+      initial = subject();
     } else {
-      new Assertion(target, flagMsg, ssfi, true).to.have.property(prop);
-      initial = target[prop];
+      new Assertion(subject, flagMsg, ssfi, true).to.have.property(prop);
+      initial = subject[prop];
     }
 
     // Make sure that the target is a number
@@ -2245,7 +2245,7 @@ module.exports = function (chai, _) {
 
     fn();
 
-    var final = prop === undefined || prop === null ? target() : target[prop];
+    var final = prop === undefined || prop === null ? subject() : subject[prop];
     var msgObj = prop === undefined || prop === null ? initial : '.' + prop;
 
     flag(this, 'deltaMsgObj', msgObj);
@@ -2265,7 +2265,7 @@ module.exports = function (chai, _) {
   Assertion.addMethod('increases', assertIncreases);
 
   /**
-   * ### .decrease(target[, property[, message]])
+   * ### .decrease(subject[, property[, message]])
    *
    * Asserts that a function decreases a numeric object property.
    *
@@ -2283,14 +2283,14 @@ module.exports = function (chai, _) {
    * @name decrease
    * @alias decreases
    * @alias Decrease
-   * @param {String|Function} target
+   * @param {String|Function} subject
    * @param {String} property name _optional_
    * @param {String} message _optional_
    * @namespace BDD
    * @api public
    */
 
-  function assertDecreases (target, prop, msg) {
+  function assertDecreases (subject, prop, msg) {
     if (msg) flag(this, 'message', msg);
     var fn = flag(this, 'object')
       , flagMsg = flag(this, 'message')
@@ -2299,11 +2299,11 @@ module.exports = function (chai, _) {
 
     var initial;
     if (!prop) {
-      new Assertion(target, flagMsg, ssfi, true).is.a('function');
-      initial = target();
+      new Assertion(subject, flagMsg, ssfi, true).is.a('function');
+      initial = subject();
     } else {
-      new Assertion(target, flagMsg, ssfi, true).to.have.property(prop);
-      initial = target[prop];
+      new Assertion(subject, flagMsg, ssfi, true).to.have.property(prop);
+      initial = subject[prop];
     }
 
     // Make sure that the target is a number
@@ -2311,7 +2311,7 @@ module.exports = function (chai, _) {
 
     fn();
 
-    var final = prop === undefined || prop === null ? target() : target[prop];
+    var final = prop === undefined || prop === null ? subject() : subject[prop];
     var msgObj = prop === undefined || prop === null ? initial : '.' + prop;
 
     flag(this, 'deltaMsgObj', msgObj);
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -13,10 +13,8 @@ module.exports = function (chai, _) {
   /**
    * ### Language Chains
    *
-   * The following are provided as chainable getters to
-   * improve the readability of your assertions. They
-   * do not provide testing capabilities unless they
-   * have been overwritten by a plugin.
+   * The following are provided as chainable getters to improve the readability
+   * of your assertions.
    *
    * **Chains**
    *
@@ -51,12 +49,20 @@ module.exports = function (chai, _) {
   /**
    * ### .not
    *
-   * Negates any of assertions following in the chain.
+   * Negates all assertions that follow in the chain.
    *
-   *     expect(foo).to.not.equal('bar');
-   *     expect(goodFn).to.not.throw(Error);
-   *     expect({ foo: 'baz' }).to.have.property('foo')
-   *       .and.not.equal('bar');
+   *     expect(function () {}).to.not.throw();
+   *     expect({a: 1}).to.not.have.property('b');
+   *     expect([1, 2]).to.be.an('array').that.does.not.include(3);
+   *
+   * Just because you can negate any assertion with `.not` doesn't mean you
+   * should. With great power comes great responsibility. It's often best to
+   * assert that the one expected output was produced, rather than asserting
+   * that one of countless unexpected outputs wasn't produced. See individual
+   * assertions for specific guidance.
+   *
+   *     expect(2).to.equal(2); // Recommended
+   *     expect(2).to.not.equal(1); // Not recommended
    *
    * @name not
    * @namespace BDD
@@ -70,15 +76,34 @@ module.exports = function (chai, _) {
   /**
    * ### .deep
    *
-   * Sets the `deep` flag, later used by the `equal`, `include`, `members`, `keys` and
-   * `property` assertions.
+   * Causes all `.equal`, `.include`, `.members`, `.keys`, and `.property`
+   * assertions that follow in the chain to use deep equality instead of strict
+   * (`===`) equality. See the `deep-eql` project page for info on the deep
+   * equality algorithm: https://github.com/chaijs/deep-eql.
    *
-   *     const obj = {a: 1};
-   *     expect(obj).to.deep.equal({a: 1});
-   *     expect([obj]).to.deep.include({a:1});
-   *     expect({foo: obj}).to.deep.include({foo: {a:1}});
-   *     expect([obj]).to.have.deep.members([{a: 1}]);
-   *     expect({foo: obj}).to.have.deep.property('foo', {a: 1});
+   *     // Target object deeply (but not strictly) equals `{a: 1}`
+   *     expect({a: 1}).to.deep.equal({a: 1});
+   *     expect({a: 1}).to.not.equal({a: 1});
+   *
+   *     // Target array deeply (but not strictly) includes `{a: 1}`
+   *     expect([{a: 1}]).to.deep.include({a: 1});
+   *     expect([{a: 1}]).to.not.include({a: 1});
+   *
+   *     // Target object deeply (but not strictly) includes `x: {a: 1}`
+   *     expect({x: {a: 1}}).to.deep.include({x: {a: 1}});
+   *     expect({x: {a: 1}}).to.not.include({x: {a: 1}});
+   *
+   *     // Target array deeply (but not strictly) has member `{a: 1}`
+   *     expect([{a: 1}]).to.have.deep.members([{a: 1}]);
+   *     expect([{a: 1}]).to.not.have.members([{a: 1}]);
+   *
+   *     // Target set deeply (but not strictly) has key `{a: 1}`
+   *     expect(new Set([{a: 1}])).to.have.deep.keys([{a: 1}]);
+   *     expect(new Set([{a: 1}])).to.not.have.keys([{a: 1}]);
+   *
+   *     // Target object deeply (but not strictly) has property `x: {a: 1}`
+   *     expect({x: {a: 1}}).to.have.deep.property('x', {a: 1});
+   *     expect({x: {a: 1}}).to.not.have.property('x', {a: 1});
    *
    * @name deep
    * @namespace BDD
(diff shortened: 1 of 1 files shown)


lib/chai/core/assertions.js
```diff
@@ -235,55 +346,144 @@ module.exports = function (chai, _) {
...
+   *
+   * Because the behavior of `.include` differs based on the target's type, it's
+   * important to always pair it with another assertion that verifies the
+   * target's type. Otherwise, `.include` will produce inaccurate results if a
+   * bug is later introduced in your code that changes the target's type. See
+   * the `.a` assertion for details on how to test a target's type.
+   *
+   *     expect([1, 2, 3]).to.be.an('array').that.includes(2);
+   *
+   * If the `deep` flag is set when the target is an array or object, deep
```
keithamus: 💭 I wonder if it is worth adding headers to these, to break up the flow a bit and also aid in discerning the individual features. Right now it feels like they blend into one and other, which is not ideal.

lib/chai/core/assertions.js
```diff
@@ -235,55 +346,144 @@ module.exports = function (chai, _) {
...
+   * The `own` and `nested` flags cannot be combined.
+   *
+   * `.include` can be negated by adding `.not` earlier in the assertion chain.
+   *
+   *     expect('foobar').to.not.include('taco');
+   *     expect([1, 2, 3]).to.not.include(4);
+   * 
+   * However, `.include` should rarely be negated when the target is an object.
+   * The problem is that it creates uncertain expectations by asserting that the
+   * target object doesn't have all of `val`'s properties but may or may not
```
keithamus: 💭 It's actually worse than that, it will assert that the target doesn't have those properties _with those values_. 

```js
expect({ c: 3 }).to.not.include({ c: 4 }) // passes
```

lib/chai/core/assertions.js
```diff
@@ -357,13 +557,37 @@ module.exports = function (chai, _) {
...
+   *     expect(false).to.be.false; // Recommended
+   *     expect(false).to.not.be.ok; // Not recommended
+   *
+   *     expect(null).to.be.null; // Recommended
+   *     expect(null).to.not.be.ok; // Not recommended
+   *
+   *     expect(undefined).to.be.undefined; // Recommended
+   *     expect(undefined).to.not.be.ok; // Not recommended
+   *
+   * A custom error message can be supplied as the second argument to `expect`.
```
keithamus: 👍 Makes me wonder if we should just straight up deprecate `ok`?

lib/chai/core/assertions.js
```diff
@@ -466,10 +740,26 @@ module.exports = function (chai, _) {
...
-   * Asserts that the target is `NaN`.
+   * Asserts that the target is exactly `NaN`.
    *
    *     expect(NaN).to.be.NaN;
-   *     expect(4).not.to.be.NaN;
+   *
+   * In modern environments, this assertion is performed using the built-in ES6
+   * `Number.isNaN` method. In legacy environments, the target is compared to
+   * itself using strict (`===`) equality. The result is the same either way:
+   * the target is only considered to be `NaN` if it's exactly `NaN`.
```
keithamus: 🛑 I feel like this is not relevant to the user here. If there is a genuine edge case that users should be wary of, we should consider it a bug and fix the edge case.

lib/chai/core/assertions.js
```diff
@@ -487,15 +777,31 @@ module.exports = function (chai, _) {
   /**
    * ### .exist
    *
-   * Asserts that the target is neither `null` nor `undefined`.
+   * Asserts that the target is not strictly (`===`) equal to either `null` or
+   * `undefined`. However, `.exist` should rarely be used. Whenever possible,
+   * it's best to assert that the target is equal to its expected value, rather
+   * than asserting that the target doesn't exist. See the `.equal` assertion
+   * for a better alternative.
```
keithamus: 💭 Again this pretty much feels like we should deprecate it.

lib/chai/core/assertions.js
```diff
@@ -511,21 +817,54 @@ module.exports = function (chai, _) {
...
+   *
    *     expect({}).to.be.empty;
    *
+   * Because the behavior of `.empty` differs based on what type of target is
+   * provided, it's important to always pair it with another assertion that
+   * verifies the target's type. Otherwise, `.empty` will produce inaccurate
+   * results if a bug is later introduced in your code that changes the target's
+   * type. See the `.a` assertion for details on how to test a target's type.
+   *
+   *     expect([]).to.be.an('array').that.is.empty;
```
keithamus: 👍

lib/chai/core/assertions.js
```diff
@@ -567,12 +906,29 @@ module.exports = function (chai, _) {
...
+   * target's type.
+   *
+   *     expect('foo').to.be.a('string'); // Recommended
+   *     expect('foo').to.not.be.arguments; // Not recommended
+   *
+   * A custom error message can be supplied as the second argument to `expect`.
+   *
+   *     expect({}, 'nooo why fail??').to.be.arguments;
+   *
+   * The alias `.Arguments` can be used interchangeably with `.arguments`.
```
keithamus: 👍

keithamus:
Quite a lot to get through here. I got up to `arguments`, will review the rest later. I've added some comments, but because this is so big I'll do some now and some later. 

### Key:
💭 - Just me expression thoughts, I won't block the review for these, but it might be nice to discuss these and maybe work to improve them, if you have time
🛑 - In my opinion, this should definitely change. This is why I have _requested changes_, fixing this will move my vote to _approved_
👍 - This is awesome work! Good job!

meeber:
I'll hold off on commenting or making any changes until everyone (who wants to) has had a chance to perform a complete review.

One question though: Is it possible to add titled subsections to a particular assertion's docs? If so, how? My guess would be adding four pound signs like `#### <subtitle>`, but I don't know if that screws up how the website generation works in terms of headers and links in the sidebar?

lib/chai/core/assertions.js
```diff
@@ -51,12 +50,20 @@ module.exports = function (chai, _) {
...
-   *       .and.not.equal('bar');
+   *     expect(function () {}).to.not.throw();
+   *     expect({a: 1}).to.not.have.property('b');
+   *     expect([1, 2]).to.be.an('array').that.does.not.include(3);
+   *
+   * Just because you can negate any assertion with `.not` doesn't mean you
+   * should. With great power comes great responsibility. Whenever possible,
+   * it's best to assert that the one expected output was produced, rather than
+   * asserting that one of countless unexpected outputs wasn't produced. See
+   * individual assertions for specific guidance.
```
lucasfcosta: +1 great work.

lib/chai/core/assertions.js
```diff
@@ -136,10 +171,17 @@ module.exports = function (chai, _) {
...
-   * Sets the `ordered` flag, later used by the `members`
-   * assertions.
+   * Sets the `ordered` flag, causing all `.members` assertions that follow in
+   * the chain to require that members be in the same order.
+   *
+   *     expect([1, 2]).to.have.ordered.members([1, 2])
+   *       .but.not.have.ordered.members([2, 1]);
+   *
+   * If both the `contains` and `ordered` flags are set, the ordering begins
+   * with the first element in the target.
```
lucasfcosta: Shouldn't it be the opposite? The ordering should follow the same order as the array given to the `include` assertion, right? I'm not sure it was just me that got it the wrong way but maybe this could be more clear.

lib/chai/core/assertions.js
```diff
@@ -153,10 +195,38 @@ module.exports = function (chai, _) {
...
    *
-   *     expect(foo).to.have.any.keys('bar', 'baz');
+   * 2. `.any.keys`: Asserts that the target has at least one of the given keys
+   * but may or may not have more of them.
+   *
+   * 3. `.not.keys`: Asserts that the target doesn't have all of the given keys
+   * but may or may not have some of them.
+   *
+   * 4. `.not.any.keys`: Asserts that the target doesn't have any of the given
+   * keys.
```
lucasfcosta: +1

lib/chai/core/assertions.js
```diff
@@ -172,10 +242,14 @@ module.exports = function (chai, _) {
   /**
    * ### .all
    *
-   * Sets the `all` flag (opposite of the `any` flag)
-   * later used by the `keys` assertion.
+   * Unsets the `any` flag, causing all `.keys` assertions that follow in the
+   * chain to revert to the default behavior of requiring that the target have
+   * all of the given keys, rather than only needing one or more of them.
```
lucasfcosta: I think we could have an example here too. This is kind of an "advanced" feature and I think many newcomers will be confused if they see this.

lib/chai/core/assertions.js
```diff
@@ -235,55 +346,144 @@ module.exports = function (chai, _) {
...
+   *
+   *     expect('foobar').to.include('foo');
+   *
+   * When invoked as a method, and when the target is an array, `.include`
+   * asserts that the given `val` is a member of the target array. By default,
+   * strict (`===`) equality is used to compare members.
+   *
+   *     expect([1, 2, 3]).to.include(2);
+   *
+   * When invoked as a method, and when the target is an object, `.include`
```
lucasfcosta: I think that maybe we could have a more general sentence telling people the main goal of this assertion which is to check if a value includes another one and then detail both `method` and `property` behaviors.

lib/chai/core/assertions.js
```diff
@@ -637,17 +1017,44 @@ module.exports = function (chai, _) {
...
+   *     expect({a: 1}, 'nooo why fail??').to.eql({b: 2});
    *
-   * Asserts that the target is deeply equal to `value`.
+   * The alias `.eqls` can be used interchangeably with `.eql`.
    *
-   *     expect({ foo: 'bar' }).to.eql({ foo: 'bar' });
-   *     expect([ 1, 2, 3 ]).to.eql([ 1, 2, 3 ]);
+   * The `deep.equal` assertion is almost identical to `.eql` but with one
+   * difference: `deep.equal` sets the `deep` flag which could impact other
+   * assertions that follow in the chain.
```
lucasfcosta: Hmmm, that's interesting. Good job!

lib/chai/core/assertions.js
```diff
@@ -968,19 +1492,32 @@ module.exports = function (chai, _) {
   });
 
   /**
-   * ### .instanceof(constructor)
+   * ### .instanceof(constructor[, msg])
```
lucasfcosta: Maybe we could add some text detailing that this fails whenever not passed a valid constructor instance and which criteria are used to determine whether something is valid or not.

lib/chai/core/assertions.js
```diff
@@ -1213,42 +1865,60 @@ module.exports = function (chai, _) {
...
+   *     expect([1, 2, 3]).to.have.lengthOf.above(2);
+   *     expect([1, 2, 3]).to.have.lengthOf.below(4);
+   *     expect([1, 2, 3]).to.have.lengthOf.at.least(3);
+   *     expect([1, 2, 3]).to.have.lengthOf.at.most(3);
+   *     expect([1, 2, 3]).to.have.lengthOf.within(2,4);
+   *
+   * Due to a compatibility issue, the alias `.length` can't be chained directly
+   * off of an uninvoked method such as `.a`. Therefore, `.length` can't be used
+   * interchangeably with `.lengthOf` in every situation. It's recommended to
+   * always use `.lengthOf` instead of `.length`.
```
lucasfcosta: 👍

lucasfcosta:
Awesome job! I just left a few new minor comments.

I think this will be awesome for our contributors to read since it gives them so much information about how to use Chai effectively instead of just telling them how things work.

Great job! Also, @keithamus comments were great too, I agree with them all.

meeber:
Note to self: Update `.change`, `.increase`, and `.decrease` docs to reflect change from chainable method to normal method.

lib/chai/core/assertions.js
```diff
@@ -1845,17 +2725,42 @@ module.exports = function (chai, _) {
   Assertion.addMethod('satisfies', satisfy);
 
   /**
-   * ### .closeTo(expected, delta)
+   * ### .closeTo(expected, delta[, msg])
+   *
+   * Asserts that the target is a number that's within a given +/- `delta` range
+   * of the given number `expected`. However, `.closeTo` should rarely be used.
+   * Whenever possible, it's best to assert that the target is equal to its
```
keithamus: 💭 another "rarely used" which is potentially hyperbolic.

lib/chai/core/assertions.js
```diff
@@ -1845,17 +2725,42 @@ module.exports = function (chai, _) {
...
+   *     // Recommended
+   *     expect(1.5).to.equal(1.5);
    *
+   *     // Not recommended
    *     expect(1.5).to.be.closeTo(1, 0.5);
+   *     expect(1.5).to.be.closeTo(2, 0.5);
+   *     expect(1.5).to.be.closeTo(1, 1);
+   *
+   * `.closeTo` can be negated by adding `.not` earlier in the assertion chain,
+   * but again, `.closeTo` should rarely be used.
```
keithamus: 💭 more redundant "rarely used"

lib/chai/core/assertions.js
```diff
@@ -2030,43 +2968,129 @@ module.exports = function (chai, _) {
...
 
   /**
-   * ### .change(function)
+   * ### .change(subject[, prop[, msg]])
+   *
+   * When invoked as a method with one argument, `.change` asserts that the
+   * given function `subject` returns a different value when it's invoked before
+   * the target function compared to when it's invoked afterward. Strict (`===`)
+   * equality is used to compare `subject`'s before and after return values.
+   * `.change` should typically only be used when it has been negated by adding
```
keithamus: 🛑 should be something like "`.change` - when used with one argument - should typically only be used with `.not`."

lib/chai/core/assertions.js
```diff
@@ -2030,43 +2968,129 @@ module.exports = function (chai, _) {
...
+   * to compare `subject[prop]`'s before and after values. But again, `.change`
+   * should typically only be used when it has been negated by adding `.not`
+   * earlier in the assertion chain.
+   *
+   *     var myObj = {dots: ''}
+   *       , noop = function () {};
+   *
+   *     expect(noop).to.not.change(myObj, 'dots');
+   *
+   * `.change` should rarely be used without negation. Whenever possible, it's
```
keithamus: 🛑 Should be something like "`.change` with two arguments should rarely..."

keithamus:
Okay @meeber I've reviewed all of this. Pretty much all great, I've got a few notes on some areas for concern though.

Largely I'm also sensing a theme - which is that these docs seem to suggest a lot of the assertions we provide have very limited use cases. Is this something we should be worried about? Taking a long look at the horizon, should we be slowly pushing these assertions into plugins?

meeber:
Thanks for the review @keithamus. I know this was a tedious one to get through. I'll rebase once #947 is merged, go through all the feedback here, make changes, and update the PR.

meeber:
@keithamus @lucasfcosta @vieiralucas @shvaikalesh I just updated this PR. I made several passes through the whole thing to address concerns, improve readability, soften and de-formalize some language, move reference links closer to concepts, etc. If there are any major issues, let's fix them. But if all that's left is minor issues, then I vote we merge and proceed with the next canary release.

lib/chai/core/assertions.js
```diff
@@ -70,15 +76,34 @@ module.exports = function (chai, _) {
...
+   *
+   *     // Target object deeply (but not strictly) equals `{a: 1}`
+   *     expect({a: 1}).to.deep.equal({a: 1});
+   *     expect({a: 1}).to.not.equal({a: 1});
+   *
+   *     // Target array deeply (but not strictly) includes `{a: 1}`
+   *     expect([{a: 1}]).to.deep.include({a: 1});
+   *     expect([{a: 1}]).to.not.include({a: 1});
+   *
+   *     // Target object deeply (but not strictly) includes `x: {a: 1}`
```
shvaikalesh: Missing curly braces: `{x: {a: 1}}`

lib/chai/core/assertions.js
```diff
@@ -92,18 +117,19 @@ module.exports = function (chai, _) {
   /**
    * ### .nested
    *
-   * Sets the `nested` flag, later used by the `property` assertion.
+   * Enables dot- and bracket-notation in all `.property` and `.include`
```
shvaikalesh: Maybe we can link `pathval` here?

lib/chai/core/assertions.js
```diff
@@ -92,18 +117,19 @@ module.exports = function (chai, _) {
...
    *
-   *     expect({ foo: { bar: { baz: 'quux' } } })
-   *       .to.have.nested.property('foo.bar.baz', 'quux');
+   *     expect({a: {b: ['x', 'y']}}).to.have.nested.property('a.b[1]');
+   *     expect({a: {b: ['x', 'y']}}).to.nested.include({'a.b[1]': 'y'});
    *
-   * `.nested.property` special characters can be escaped by adding two slashes
-   * before the `.` or `[]`.
+   * If `.` or `[]` are part of an actual property name, they can be escaped by
+   * adding two backslashes before them.
```
shvaikalesh: Nit: second backslash is necessary only to escape the first one, unless one does
```js
String.raw`\.link`
```

If we are targeting beginner-level devs, wording is OK 👍

lib/chai/core/assertions.js
meeber: That's actually intentional. In the previous example, braces are included because `.include` is looking for an object `{a: 1}` inside the target array. But in that example, braces aren't included because `.include` is looking for a property `x` with a value of `{a: 1}` in the target object. I think it would take too much verbiage to clarify this distinction within the docs, without enough gain, so I vote just leave as-is.

lib/chai/core/assertions.js
meeber: Not sure there's enough value with linking `pathval`. The value in linking `deep-eql` and `type-detect` is that they both have separate documentation that list rules and behaviors that are relevant to someone writing an assertion using those flags, but the documentation for `pathval` just provides documentation on a bunch of functions that are used when using the module independently of Chai. Since we're moving in the direction of having a bunch of micro libraries, I think we should only link those that provide direct relevant information to the Chai user.

lib/chai/core/assertions.js
meeber: During my update, I didn't really see the need to get into the details of this failing due to an invalid constructor instance. I think it's sufficiently understood by the assertion itself that it only passes when the target is an instance of the constructor, without drawing any distinction between a valid-but-different-constructor and an invalid constructor in the case of failure.

lib/chai/core/assertions.js
meeber: Good point! I vote to keep as-is for now to be beginner friendly, but it might be interesting in the future to add another example using the single-slash alternative.

shvaikalesh:
LGTM. Incredible job, @meeber! Love the consistency (wording, param names). Agreed on moving forward with merging.

lib/chai/core/assertions.js
```diff
@@ -607,24 +900,46 @@ module.exports = function (chai, _) {
...
   /**
-   * ### .equal(value)
+   * ### .equal(val[, msg])
+   *
+   * Asserts that the target is strictly (`===`) equal to the given `val`.
+   *
+   *     expect(1).to.equal(1);
+   *     expect('foo').to.equal('foo');
+   * 
+   * Add `.deep` earlier in the chain to use deep equality instead. See the
```
lucasfcosta: Many users get confused about the behavior of `deep` when it comes to object's prototypes, maybe it would be cool for us to explain more about it here and avoid issues later.

lib/chai/core/assertions.js
```diff
@@ -751,25 +1109,44 @@ module.exports = function (chai, _) {
   Assertion.addMethod('greaterThan', assertAbove);
 
   /**
-   * ### .least(value)
+   * ### .least(n[, msg])
+   *
+   * Asserts that the target is a number greater than or equal to the given
+   * number `n`. However, it's often best to assert that the target is equal to
```
lucasfcosta: IMO it makes thinks strange to have a feature and then discourage users to use it unless we show them a good reason to do it.
Maybe we could add a simple explanation telling them that when they assert `something` is not `<other_thing>` this means that `something` can be `Inifinity - 1` things, which is too loose.

What do you think? I'm not sure this would be too much or not.

lib/chai/core/assertions.js
meeber: I think deep equality is mentioned too many times throughout the documentation to go into detail about the algorithm. I prefer linking to the deep-eql project page instead where those kinds of details can go.

lib/chai/core/assertions.js
meeber: There's about a dozen assertions that fall into this category. I intend on making the argument post-4.x that a lot of them should be removed from core. For now I kinda took the strategy of gently steering users away from misusing these assertions by using language like "it's often best to do this other thing" and showing an example. For any such assertions that we end up keeping, I'd be in favor of adding a bit more explanation on when it is an appropriate time to use it.

lucasfcosta: @meeber I'm convinced.

LGTM.

keithamus:
2 approvals makes a merge 😄

Good work, as always @meeber

PR #922, opened 2017-01-29
Remove implementation frames from stack trace

### LINKED ISSUE #878: Question regarding stack trace.

### LINKED ISSUE #904: Flag transfer problem in failed `.include` assertion
Proxy-related frames are showing up in the stack trace for the `.include` assertion when testing for a subset of properties in an object. Also, custom error messages aren't being displayed in this case.

Example:

```js
expect({x: {a: 1}}).to.include({x: {x: 9}}, "oh noooo");
```

Output:

>       AssertionError: expected { x: { a: 1 } } to have property 'x' of { x: 9 }, but got { a: 1 }
>       + expected - actual
> 
>        {
>       -  "a": 1
>       +  "x": 9
>        }
> 
>       at Proxy.<anonymous> (node_modules/chai/lib/chai/core/assertions.js:322:25)
>       at Array.forEach (native)
>       at Proxy.include (node_modules/chai/lib/chai/core/assertions.js:317:13)
>       at Proxy.chainableMethodWrapper (node_modules/chai/lib/chai/utils/addChainableMethod.js:93:49)
>       at Context.<anonymous> (test/index.js:8:26)

I suspect both problems are a lack of flags being transferred to the `.property` assertion that the `.include` assertion calls internally. Will investigate later, just noting for now.

meeber:
**Note to self:** Rebase after #899 is merged and update to turn `Error` into an `AssertionError` that properly retains ssfi. A follow up PR will properly retain the user's custom message here and elsewhere.

This PR is split into five commits:

1. Expand the role of the `keep_ssfi` flag so that it's easier to tell Chai not to overwrite an existing `ssfi` when it shouldn't, such as when an assertion is being called from within another assertion, or when an assertion is wrapped in a function such as every assertion on the `assert` interface.

2. Fix all issues on `expect`/`should` interfaces that were causing implementation frames to leak through. Also improve the internal Chai error test helper to validate stack traces.

3. Fix all issues on `assert` interface that were causing implementation frames to leak through. Further improve the internal Chai error test helper to validate stack traces.

4. Rename `keep_ssfi` flag to `lockSsfi`.

5. Clean up excessive test comments.

Closes #878
Closes #904

### DIFF
--- a/lib/chai/utils/addChainableMethod.js
+++ b/lib/chai/utils/addChainableMethod.js
@@ -87,17 +87,17 @@ module.exports = function addChainableMethod(ctx, name, method, chainingBehavior
           // frames from the stack trace of a failed assertion.
           //
           // However, we only want to use this function as the starting point if
-          // the `keep_ssfi` flag isn't set.
+          // the `lockSsfi` flag isn't set.
           //
-          // If the `keep_ssfi` flag is set, then this assertion is being
+          // If the `lockSsfi` flag is set, then this assertion is being
           // invoked from inside of another assertion. In this case, the `ssfi`
           // flag has already been set by the outer assertion.
           //
           // Note that overwriting a chainable method merely replaces the saved
           // methods in `ctx.__methods` instead of completely replacing the
           // overwritten assertion. Therefore, an overwriting assertion won't
-          // set the `ssfi` or `keep_ssfi` flags.
-          if (!flag(this, 'keep_ssfi')) {
+          // set the `ssfi` or `lockSsfi` flags.
+          if (!flag(this, 'lockSsfi')) {
             flag(this, 'ssfi', chainableMethodWrapper);
           }
 
--- a/lib/chai/utils/addChainableMethod.js
+++ b/lib/chai/utils/addChainableMethod.js
@@ -82,13 +82,24 @@ module.exports = function addChainableMethod(ctx, name, method, chainingBehavior
         chainableBehavior.chainingBehavior.call(this);
 
         var chainableMethodWrapper = function () {
-          // Use this chainable method wrapper as the starting point for
-          // removing implementation frames from the stack trace of a failed
-          // assertion. Note that this is the correct starting point even if
-          // this assertion has been overwritten since overwriting a chainable
-          // method merely replaces the saved methods in `ctx.__methods` instead
-          // of completely replacing the overwritten assertion.
-          flag(this, 'ssfi', chainableMethodWrapper);
+          // Setting the `ssfi` flag to `chainableMethodWrapper` causes this
+          // function to be the starting point for removing implementation
+          // frames from the stack trace of a failed assertion.
+          //
+          // However, we only want to use this function as the starting point if
+          // the `keep_ssfi` flag isn't set.
+          //
+          // If the `keep_ssfi` flag is set, then this assertion is being
+          // invoked from inside of another assertion. In this case, the `ssfi`
+          // flag has already been set by the outer assertion.
+          //
+          // Note that overwriting a chainable method merely replaces the saved
+          // methods in `ctx.__methods` instead of completely replacing the
+          // overwritten assertion. Therefore, an overwriting assertion won't
+          // set the `ssfi` or `keep_ssfi` flags.
+          if (!flag(this, 'keep_ssfi')) {
+            flag(this, 'ssfi', chainableMethodWrapper);
+          }
 
           var result = chainableBehavior.method.apply(this, arguments);
           if (result !== undefined) {
--- a/test/globalErr.js
+++ b/test/globalErr.js
@@ -56,7 +56,7 @@ describe('globalErr', function () {
 
     err(function () {
       err(function () { throw new Err('cat') }, {message: 'dog'});
-    }, 'expected \'cat\' to deeply equal \'dog\'');
+    }, 'expected \'cat\' to deeply equal \'dog\'', true);
   });
 
   it('should throw if fn does not throw', function () {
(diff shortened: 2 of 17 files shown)


meeber:
Pinging @keithamus @lucasfcosta @vieiralucas @shvaikalesh for review :D

lib/chai/utils/addChainableMethod.js
```diff
@@ -82,13 +82,24 @@ module.exports = function addChainableMethod(ctx, name, method, chainingBehavior
...
-          // removing implementation frames from the stack trace of a failed
-          // assertion. Note that this is the correct starting point even if
-          // this assertion has been overwritten since overwriting a chainable
-          // method merely replaces the saved methods in `ctx.__methods` instead
-          // of completely replacing the overwritten assertion.
-          flag(this, 'ssfi', chainableMethodWrapper);
+          // Setting the `ssfi` flag to `chainableMethodWrapper` causes this
+          // function to be the starting point for removing implementation
+          // frames from the stack trace of a failed assertion.
+          //
```
meeber: The `ssfi` flag should always contain a reference to the lowest internal Chai frame in the stack, because that frame and all frames above it will be filtered out. In other words, the very first Chai internal function that gets hit in an assertion should set the `ssfi` flag, and any functions after that shouldn't mess with the `ssfi` flag.

For example:

```js
expect({foo: 1}).to.include({foo: 1});
```

- When `expect({foo: 1})` gets called, `ssfi` is `undefined`.
- When `.to` gets called, `ssfi` gets set to either the `proxyGetter` or `propertyGetter` function for `.to`, depending on whether on not proxies are supported and enabled. Note that `.to` doesn't call any more internal Chai functions, so we're done with this step. But if it did happen to call more internal Chai functions, then it'd freeze it's `ssfi` before calling them since we want to keep `ssfi` pointing at the first internal Chai function that gets hit when `.to` is invoked. 
- When `.include` gets called, `ssfi` gets set to the `chainableMethodWrapper` function for `.include`. Now, `.include` calls the `.property` assertion internally, so we freeze `ssfi` before calling `.property` since we want to keep `ssfi` pointing at the first internal Chai function that gets hit when `.include` is invoked.

meeber:
Pushed a new version. Changes:

- Added new commit that renames `keep_ssfi` flag to `lockSsfi`. (Avoided "freeze" to avoid confusion with JavaScript's concept of a frozen object.)
- Modified existing commits to make the `globalErr` helper function throw a more description error message when implementation frames are detected in the stack trace of a failed assertion.
- Modified existing commits to add comments that make it more clear what exactly tests of the `globalErr` helper function are doing.

test/globalErr.js
```diff
@@ -108,4 +108,112 @@ describe('globalErr', function () {
...
+      // Skip tests if `err.stack` is unsupported
+      if (typeof err.stack === 'undefined') return;
+    }
+
+    // Note: `.to.not.throw` isn't used for the assertions that are expected to
+    // not throw an error because it'll pollute the very same stack trace which
+    // is being asserted on. Instead, if a test is failing, then Mocha will pick
+    // up the error that's thrown by `err`.
+    describe('falsey', function () {
+      // The inner `err` executes `myGetter` and catches the thrown error. But
```
lucasfcosta: I don't think we need these comments.

They are really well written, but I think it is very easy to understand what is happening just by reading the code. If we have too many comments on the code it just makes it more difficult to maintain because we gotta make sure we update all the related comments, otherwise they will tell a different story than the code does.

IMO the outer code is okay, but this one (inner one) seems too much.

Anyway, I won't reject this PR because of this, please let me know what you think about it. 😄

test/globalErr.js
```diff
@@ -108,4 +108,112 @@ describe('globalErr', function () {
...
+      // `skipStackTest` is truthy, it shouldn't see the word "Getter" in the
+      // stack trace of the thrown error, and thus shouldn't throw a new error
+      // (which would cause this test to fail).
+      it('shouldn\'t throw if "Getter" is in the stack trace', function () {
+        err(function myGetter () {
+          throw Error('my stack trace contains a fake implementation frame');
+        }, undefined, true);
+      });
+
+      // `err` executes `myWrapper` and catches the thrown error. Since
```
lucasfcosta: These comments are almost all the same, I think it adds a lot of noise to the code, especially because one of them would already be enough, but even more because now that we've got more descriptive error messages and awesome test case descriptions that already do the job of conveying the message we want it to.

meeber:
Pushed new version with excessive comments removed. (Changed it in a new commit instead of rebase in case we change mind in future).

snackycracky:
this article refers to this pr http://lucasfcosta.com/2017/02/17/JavaScript-Errors-and-Stack-Traces.html

keithamus:
How are we with this @meeber? Given:

> *Note to self*: Rebase after #899 is merged and update to turn Error into an AssertionError that properly retains ssfi. A follow up PR will properly retain the user's custom message here and elsewhere.

meeber:
@keithamus I made that note to myself yesterday :D 

I'll update this PR after work!

meeber:
@keithamus @shvaikalesh @vieiralucas @lucasfcosta Rebased since #899 was merged. As expected, some `instanceof` tests started failing due to implementation frames in stack trace. Updated as planned. Tests now pass.

I think this PR is ready to go. After this is merged, there will be a follow-up PR related to messages.

lucasfcosta:
LGTM!
Gj!

meeber:
@keithamus @shvaikalesh @vieiralucas Need a second review on this one!

keithamus:
LGTM @meeber! Great work, it's not always glamorous to do this kind of work, but it's definitely needed. That's one of the many reasons you're such a valuable contributor 😘 👍 💯

PR #924, opened 2017-01-30
fix: make negated `.keys` consider size of sets

### LINKED ISSUE #919: `.not.have.keys` incorrectly failing

meeber:
When neither the `contains` nor `any` flags are set, the `.keys`
assertion implicitly means `.all.keys` and thus requires the target
and given sets to be the same size. Therefore, `not.keys` implicitly
means `.not.all.keys`, and should thus pass when the target and given
sets aren't the same size, even if the target set is a superset of the
given set. This commit enables this behavior.

Fixes #919

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -1461,7 +1461,7 @@ module.exports = function (chai, _) {
         });
       });
 
-      if (!flag(this, 'negate') && !flag(this, 'contains')) {
+      if (!flag(this, 'contains')) {
         ok = ok && keys.length == actual.length;
       }
     }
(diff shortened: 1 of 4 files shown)


lucasfcosta:
LGTM!

Seems more straightforward this way. Good job!

meeber:
This is a small one. Pinging @keithamus @vieiralucas @shvaikalesh

keithamus:
Okay, happy enough with this given the context.

PR #925, opened 2017-01-30
fix: wrong assertion types

### LINKED ISSUE #917: `.change`, `.increase`, and `.decrease` wrong assertion type
These three assertions are all registered as chainable method assertions despite not having any chainable behavior. It'd be okay for them to be chainable if they had non-specific meaning like "of" and "at", but that's not the case. Therefore, they should be changed to normal method assertions.

I suspect they were registered as chainable method assertions because of their interaction with the `by` assertion, e.g., `expect(targetFn).to.increase(subjectFn).by(2)`. However, that's still normal method assertion behavior; it'd only be chainable method assertion behavior if the non-invoked property-access version of the assertion had a side-effect.

meeber:
BREAKING CHANGE: `.change`, `.increase`, and `.decrease` changed from
Chainable Method Assertions to Method Assertions. They don't have any
chaining behavior, and there's no generic semantic benefit to chaining
them.

Fixes #917 

Note: If this is merged, #920 needs to be updated in order to remove verbiage about chaining behavior.

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -2083,8 +2083,8 @@ module.exports = function (chai, _) {
     );
   }
 
-  Assertion.addChainableMethod('change', assertChanges);
-  Assertion.addChainableMethod('changes', assertChanges);
+  Assertion.addMethod('change', assertChanges);
+  Assertion.addMethod('changes', assertChanges);
 
   /**
    * ### .increase(target[, property[, message]])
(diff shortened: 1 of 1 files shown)


meeber:
@vieiralucas Good point! I've updated the commit message.

lucasfcosta:
LGTM too! Well done @meeber!

Also, let's not forget to update #920.

Issue #923, opened 2017-01-29
Custom message not always respected


meeber:
As noted [here](https://github.com/chaijs/chai/issues/904#issuecomment-275867812), there's a problem with some assertions not always respecting a custom failed assertion error message. This problem is different (and more widespread) than the flag transfer problem mentioned in #904.

Here's an example:

```js
expect('blah').to.be.closeTo(2, 1, 'CUSTOM');  // Correctly includes custom error message
expect('blah', 'CUSTOM').to.be.closeTo(2, 1); // Incorrectly excludes custom error message
```

The culprit for the above example is [this line](https://github.com/chaijs/chai/blob/576a6f71530a158b014a97fc4bf9a152c1fbccd6/lib/chai/core/assertions.js#L1867). If the `msg` argument isn't provided, then it needs to look at the value of the `msg` flag instead.

Many other occurrences of this issue are similar to the above example. Also there are some assertions that fail to pass on the `msg` at all to internal assertions or errors, but #922 coincidentally fixes those occurrences.

keithamus:
Something I'd like to just throw out there: is it worth fixing these issues given that if we implement #585 anything like it currently looks, then messaging will be centralised and automated, and any efforts will be refactored away?

meeber:
I think it's worth it, it's a quick-n-easy fix, low-impact. Just waiting for #922 so don't have to mess with resolving merge conflicts.

PR #934, opened 2017-02-15
Custom message not always respected 

artsoroka:
Inside "closeTo" function "this" already references an Assertion with "msg" parameter previously being set, so no need to call new Assertion 

Fixes [#923](https://github.com/chaijs/chai/issues/923)


meeber:
Hi @artsoroka. Thanks for the PR. Although this solution does technically resolve the problem with `.closeTo`, it's not really compatible with #922, and would need to be refactored again as soon as #922 is merged. Therefore, the current plan is to wait until #922 is merged and then fix all the custom message issues in one go.

Issue #909, opened 2017-01-21
Custom Exception classes ignored


booleangate:
Unless I'm missing something, custom exception classes are being ignored when checked with `throw`.  Here's an example:

```js
class NotImplemented extends Error { }
expect(() => { throw new NotImplemented(); }).to.not.throw(NotImplemented);
```
This expectation passes although it shouldn't.

booleangate:
Did some more testing and this is probably due to Babel.  Changing the class definition to `Object.create` instead of `class ... extends` makes the expectation fail as expected.

```js
function NotImplemented() {
    Error.prototype.constructor.apply(this, arguments);
}

NotImplemented.prototype = Object.create(Error.prototype);
NotImplemented.prototype.name = 'NotImplemented';
NotImplemented.prototype.constructor = NotImplemented;

expect(() => { throw new NotImplemented(); }).to.not.throw(NotImplemented);
```

Maybe this is worth noting somewhere in the docs?

I'll leave this issue open so it can be seen, but feel free to close it.

keithamus:
Hey @booleangate thanks for the issue.

This is because Babel does not support extending native types - as some older browsers don't do this.

Take a look at some of these: https://babeljs.io/docs/usage/caveats/#classes, https://github.com/babel/babel/issues/4058, https://github.com/babel/babel/issues/106, https://www.npmjs.com/package/babel-plugin-transform-builtin-extend

I'll close this; while we _could_ document it I somewhat don't think it is not our responsibility to document issues with other tools and would be burdensome to do so. Babel has [some small documentation about this on their site](https://babeljs.io/docs/usage/caveats/#classes) - it could arguably be more prominent but thats not for us to dictate 😄 

Feel free to continue discussion here. If we come to a different resolution, we can always re-open this issue.

booleangate:
Indeed, documenting individual interoperability edge cases would be a nightmare.  Relying on Google to bring up this PR for people with my particular issue is probably enough.

asinner:
Just FYI if you're using Typescript and reading this, you will run into this issue trying to check the error type using `chai.throw()`. Here is the [Typescript doc](https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes#extending-built-ins-like-error-array-and-map-may-no-longer-work) on this particular issue.

In order to handle this I use the matching feature to ensure that the error it's throwing is the expected error (a similar approach will work in other compiled Javascript langs):

```ts
class NotFoundError extends Error {
[...]

Issue #773, opened 2016-08-18
throw() behaves unexpected with custom ES2015 errors


netzhuffle:
Code sample:

```
class MyError extends Error {
  constructor(message) {
    super(message);
    this.message = message;
    this.name = 'MyError';
  }
}

function myFunction() {
  throw new MyError();
}

expect(myFunction).to.throw(MyError);
```

Expected: The test should pass.

Actual:
> AssertionError: expected [Function: myFunction] to throw 'MyError' but 'MyError' was thrown

I assume this to be a bug as the message two mentioned classes are the same.

Note that the code sample works as expected when _not_ extending Error:

```
class MyError {
  constructor(message) {
    this.message = message;
    this.name = 'MyError';
  }
}
```

lucasfcosta:
Hi @netzhuffle, thanks for the issue!
Which version of Chai are you using?
We've done a big refactor of the `throws` assertion for `4.0.0`, I also looked at the code on the `3.5.0` tag and I couldn't see why your code could be failing, since it should be falling into the correct if/else branches according to what NodeJS console showed me.

I also tested the exact same code you provided on my machine and it seems to work fine. It even breaks when I comment the line which throws the error (as expected), therefore I conclude you are not using the latest released version (`3.5.0`).

Please let me know if you still have this issue after updating to the latest released version.
Thanks again for getting in touch :smile:

netzhuffle:
Thanks @lucasfcosta! I’m using Chai `3.5.0`.

But: I completely overlooked that I used babel whose transform-es2015-classes plugin is not compatible with chai’s `throws` (as it wraps classes in functions). So this is not a bug but an incompatibility.

Thanks for your help!

eddyerburgh:
This issue isn't resolved. I can replicate this in node without babel.

Running node 6.9.1

```
class CustomError extends Error {}

function throwError() {
    throw new CustomError('custom error');
}

it('Throws a custom error', () => {
    expect(throwError()).to.throw(errorTypes.CustomError);
});
```
Output: 
```
Error: custom error
```

This is my workaround : 
```
it('Throws a custom error', () => {
    Promise.resolve()
        .then(() => (throwError()))
        .then(() => {
            throw new Error('Should not resolve with invalid duration');
        })
        .catch((err) => {
            expect(err).to.be.a.instanceOf(CustomError);
        });
});
```

I'm willing to take a look at this and get a PR.

meeber:
[...]

Issue #596, opened 2016-01-24
Throw custom error (ES6)


theofidry:
Given a custom error:

```js
export default class DummyError extends Error {
    /**
     * @param {string} [message='']
     */
    constructor(message = '') {
        super(message);

        this.name = 'DummyError';
        this.message = message;

        if (Error.hasOwnProperty('captureStackTrace')) {
            Error.captureStackTrace(this, this.constructor);
        }

        this.stack = (new Error(message)).stack;
    }
}
```

And the following function:

```js
function throwDummyError() {
    throw new DummyError();
}
```

The following doesn't work:

```js
assert.throw(throwDummyError, DummyError);
```

And instead is throwing:

```
AssertionError: expected [Function: DummyError] to throw 'DummyError' but 'my error message' was thrown
```

keithamus:
Hey @theofidry thanks for the issue.

That should definitely work. The way we handle errors is getting better in `4.x.x`, so I'd first ask you to reproduce the error using the `4.x.x` branch. Could you try to reproduce using the tip of `4.x.x` please? Steps to do so:

1. `npm install chaijs/chai#4.x.x`
2. Run your tests
3. Report results :smile:

theofidry:
Indeed this was using the 3.4.1 version, should have said so from the beginning. I'm going to test it with the 4.x.x branch and come back to you.

keithamus:
Hey @theofidry have you had a chance to try out `4.x.x` yet? It'd be great to hear back from you :smile: :ear:

theofidry:
Unfortunately no, busy times :disappointed:. But think I will be able to do it today or during the weekend...

keithamus:
Great! I look forward to seeing how you go with it :smile:

theofidry:
Nope doesn't work :( https://github.com/haircvt/serializerjs/pull/14

theofidry:
Ha, looks like the `npm install` does not pass in Travis, locally I have:

```
 1) SerializerInterface As an interface, it should not implement anything:

      AssertionError: expected [Function: deserialize] to throw 'UnimplementedMethodError' but 'UnimplementedMethodError: Unimplemented "deserialize()" method' was thrown
      + expected - actual

      -UnimplementedMethodError: Unimplemented "deserialize()" method
      +UnimplementedMethodError

      at Function.assert.throws (node_modules/chai/lib/chai/interface/assert.js:1028:52)
[...]

PR #938, opened 2017-02-20
docs: update `instanceof`, `throws` docs to mention transpiler issues.

keithamus:
This documents problems users have run into in #596, #773, #909

This might be in conflict with #920 - feel free to merge this changes into #920 and close this one, or whatever.


meeber:
@keithamus Agreed, it's not ideal to have this kind of thing in our docs but at this point it seems less bad to have it than to not have it. Thanks for writing this up. I updated #920 with the changes. Closing this.

PR #947, opened 2017-03-26
fix: always honor custom message

### LINKED ISSUE #916: `.by` doesn't accept optional `msg` parameter
Noticed this while updating the docs. Seems like a minor oversight. Will fix later, just noting for now.

### LINKED ISSUE #923: Custom message not always respected

meeber:
Has #922 as a prequisite. Recommend reviewing that PR separately, and then only reviewing the diff from the final commit in this PR.

Closes #923 
Closes #916

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -299,9 +299,10 @@ module.exports = function (chai, _) {
   }
 
   function include (val, msg) {
+    if (msg) flag(this, 'message', msg);
+
     _.expectTypes(this, ['array', 'object', 'string'], flag(this, 'ssfi'));
 
-    if (msg) flag(this, 'message', msg);
     var obj = flag(this, 'object')
       , objType = _.type(obj).toLowerCase()
       , isDeep = flag(this, 'deep')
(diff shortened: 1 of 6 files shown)


keithamus:
Another great changeset @meeber! LGTM!

meeber:
Thanks @keithamus!

@lucasfcosta @vieiralucas @shvaikalesh Need one more review on this one! :D

Issue #905, opened 2017-01-18
Missing `.nested.include` and `.own.include`


meeber:
The `nested` and `own` flags currently work correctly with the `.include` assertion, but it's not documented or tested. Also, this functionality isn't currently exposed via the `assert` interface.

Note that I'm already making some documentation updates that include this, but the issue with the tests and `assert` interface are still outstanding.

meeber:
BDD docs and tests were updated to resolve most of this issue. Still outstanding: Exposing on `assert` interface and adding tests.

zetamorph:
I would like to do this and already read the contributor guide and familiarised myself with the coding style, but just to clarify, that task would also entail exposing the .deepNestedInclude, .deepOwnInclude(and the negated versions, respectively) methods on the assert interface?

meeber:
@zetamorph Yup, exactly. For tests, I'd recommend just duplicating the ones for each combination from the `expect` test suite, and then converting them over to `assert`-style. For example [here](https://github.com/chaijs/chai/blob/850bd7e016606ce52292c0477320279a8b9cf294/test/expect.js#L1839-L1864). It's okay to combine both the normal and negated versions into a single `it` block to make the conversion easier.

Thanks and feel free to reach out with any questions!

zetamorph:
@meeber Thanks for the heads-up. 
I have one question: does the coding style guide, specifically for nested arrays and objects, also strictly apply to examples in the documentation?

I find that the following:

```javascript
assert.deepNestedInclude(
  {a: 
    {b: 
      [
        {x: 1}
      ]
    }
  }, 
  {'a.b[0]': 
    {x: 1}
  }
);
```
is much harder to read than this:
```javascript
assert.deepNestedInclude({ a: { b: [ { x: 1 } ] } }, { 'a.b[0]': { x: 1 } });
```

meeber:
@zetamorph Coding examples can be written with a focus on readability. See [these examples](https://github.com/chaijs/chai/blob/850bd7e016606ce52292c0477320279a8b9cf294/lib/chai/core/assertions.js#L365-L381) from the updated BDD docs.

zetamorph:
@meeber Ok, thanks. 
I wasn`t sure because the style guide said that parts of the codebase still follow older styles. 
I submitted a pull request, https://github.com/chaijs/chai/pull/964.

PR #964, opened 2017-05-04
assert: add nestedInclude, deepNestedInclude, ownInclude and deepOwnInclude

zetamorph:
I added these methods and their negated versions to the assert interface and adapted the expect-style tests to assert-style as discussed in this issue: https://github.com/chaijs/chai/issues/905.

### DIFF
--- a/test/assert.js
+++ b/test/assert.js
@@ -763,7 +763,7 @@ describe('assert', function () {
     }, "blah: expected { a: { b: [ 'x', 'y' ] } } to have nested property 'a.b[1]' of 'x', but got 'y'");
 
     err(function () {
-      assert.nestedInclude({a: {b: ['x', 'y']}}, 'blah', {'a.b[1]': 'x'});
+      assert.nestedInclude({a: {b: ['x', 'y']}}, {'a.b[1]': 'x'}, 'blah');
     }, "blah: expected { a: { b: [ 'x', 'y' ] } } to have nested property 'a.b[1]' of 'x', but got 'y'");
 
     err(function () {
@@ -771,7 +771,7 @@ describe('assert', function () {
     }, "expected { a: { b: [ 'x', 'y' ] } } to have nested property 'a.c'");
 
     err(function () {
-      assert.nestedInclude({a: {b: ['x', 'y']}}, {'a.b[1]': 'y'});
+      assert.notNestedInclude({a: {b: ['x', 'y']}}, {'a.b[1]': 'y'});
     }, "expected { a: { b: [ 'x', 'y' ] } } to not have nested property 'a.b[1]' of 'y'");
   });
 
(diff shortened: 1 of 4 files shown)


test/assert.js
```diff
@@ -748,6 +748,104 @@ describe('assert', function () {
...
+    err(function () {
+      assert.nestedInclude({a: {b: ['x', 'y']}}, {'a.b[1]': 'x'}, 'blah');
+    }, "blah: expected { a: { b: [ 'x', 'y' ] } } to have nested property 'a.b[1]' of 'x', but got 'y'");
+
+    err(function () {
+      assert.nestedInclude({a: {b: ['x', 'y']}}, {'a.c': 'y'});
+    }, "expected { a: { b: [ 'x', 'y' ] } } to have nested property 'a.c'");
+
+    err(function () {
+      assert.notNestedInclude({a: {b: ['x', 'y']}}, {'a.b[1]': 'y'});
```
meeber: Let's add a third arg `'blah'` here so that `assert.notNestedInclude` has at least one test verifying that custom messages work. (I think it's an oversight that some of the existing test blocks like `.notDeepInclude` don't have that).

test/assert.js
```diff
@@ -748,6 +748,104 @@ describe('assert', function () {
...
+    err(function () {
+      assert.deepNestedInclude({a: {b: [{x: 1}]}}, {'a.b[0]': {y: 2}}, 'blah');
+    }, "blah: expected { a: { b: [ [Object] ] } } to have deep nested property 'a.b[0]' of { y: 2 }, but got { x: 1 }");
+
+    err(function () {
+      assert.deepNestedInclude({a: {b: [{x: 1}]}}, {'a.c': {x: 1}});
+    }, "expected { a: { b: [ [Object] ] } } to have deep nested property 'a.c'");
+
+    err(function () {
+      assert.notDeepNestedInclude({a: {b: [{x: 1}]}}, {'a.b[0]': {x: 1}});
```
meeber: Same thing here with adding a third arg `'blah'`.

test/assert.js
```diff
@@ -748,6 +748,104 @@ describe('assert', function () {
...
+    err(function () {
+      assert.deepOwnInclude({a: {b: 2}}, {a: {c: 3}}, 'blah');
+    }, "blah: expected { a: { b: 2 } } to have deep own property 'a' of { c: 3 }, but got { b: 2 }");
+
+    err(function () {
+      assert.deepOwnInclude({a: {b: 2}}, {'toString': Object.prototype.toString});
+    }, "expected { a: { b: 2 } } to have deep own property 'toString'");
+
+    err(function () {
+      assert.notDeepOwnInclude({a: {b: 2}}, {a: {b: 2}});
```
meeber: Same thing here with adding a third arg `'blah'`.

test/assert.js
```diff
@@ -748,6 +748,104 @@ describe('assert', function () {
...
+    err(function () {
+      assert.ownInclude({a: 1}, {a: 3}, 'blah');
+    }, "blah: expected { a: 1 } to have own property 'a' of 3, but got 1");
+
+    err(function () {
+      assert.ownInclude({a: 1}, {'toString': Object.prototype.toString});
+    }, "expected { a: 1 } to have own property 'toString'");
+
+    err(function () {
+      assert.notOwnInclude({a: 1}, {a: 1});
```
meeber: Same thing here with adding a third arg `'blah'`.

zetamorph:
Thanks! I added the custom messages. I will also do the same for the other tests.

meeber:
LGTM! Let's get another set of eyes on this before merging though: @vieiralucas @lucasfcosta @keithamus @shvaikalesh

shvaikalesh:
LGTM except for naming: maybe we should come up with something more descriptive than `object1` and `object2`? Personally I prefer `haystack` and `needle`.

zetamorph:
@shvaikalesh I was initially trying to come up with something specific to these methods that´s still not too long, but failed. You´re right in that haystack and needle are much more descriptive, I will change it, then the documentation for include-related methods is also more uniform.

meeber:
LGTM

Issue #890, opened 2016-12-19
Plugin issues with Chai v4


meeber:
I went through the popular Chai plugins to see how their test suites fair with Chai v4. (Actually, I used the #884 build instead of the current canary release). Overall, we're in good shape. Here are my findings:

1. chai-immutable
2. chai-jquery
3. chai-react

**Major issues**

- chai-immutable
  - Code had to be updated because of breaking change of `deep.property` to `nested.property`. Also because of change in capitalization in values returned by `type-detect`. Also because `.not.keys` was being treated implicitly as `.not.any.keys` instead of `.not.all.keys`. Further changes are likely needed due to change in `not.property`.
  - Updated code only works in Chai v4.0; therefore, these are breaking changes for end users
  - Tests had to be updated accordingly to change `deep.property` to `nested.property`, and to remove some of the `.not.keys` tests since they no longer made sense
  - Updated tests only work in Chai v4.0
  - PR submitted (https://github.com/astorije/chai-immutable/pull/69) but shouldn't be merged because there's more work to be done, and it shouldn't be merged until Chai v4 is released; `package.json` needs updating

- chai-things
  - Code was already broken prior to v4 due to `any` flag in Chai core
  - Tests had to be updated because of changed wording in failed property assertion
  - Updated tests only work in Chai v4.0; still one broken test because of `any` flag in core
  - No PR submitted yet because of major `any` issue
  - Independently of this, I have an open issue (#881) about possibly deprecating `any` in core, which would happen to resolve this problem

**Minor issues**

- chai-as-promised
  - Code had to be updated because in Chai v4, assertions return new Assertion objects with flags transferred over, and the promise was getting lost as a result
  - Updated code still works in Chai v3.5
  - One test had to be updated to change `deep.property` to `nested.property`
  - Updated test only works in Chai v4.0
  - PR submitted (https://github.com/domenic/chai-as-promised/pull/157) but shouldn't be merged until Chai v4 is released; `package.json` needs updating

- chai-jquery
  - Code had to be updated because in Chai v4, if `undefined` is explicitly passed as the second parameter to the `property` assertion, it asserts that the value value of the property equals `undefined`.
  - Updated code still works in Chai v3.5
  - Four tests had to be updated due to small change in failed assertion message from "have a property" to "have property"
  - PR submitted (https://github.com/chaijs/chai-jquery/pull/105) and can be merged now

- chai-http
  - No code had to be updated
  - Tests had to be updated to change `deep.property` to `nested.property`
[...]

PR #966, opened 2017-05-07
fix: PhantomJS 1.x incompatibility

meeber:
- Addresses https://github.com/chaijs/chai/issues/890#issuecomment-298168561
- Although I'm not crazy about adding a workaround in Chai for a 3rd party library, in this particular case it feels like a low-cost solution to allow Chai to continue supporting PhantomJS 1.x until plugins have had a chance to upgrade.


keithamus:
LGTM 👍. I'll leave the merge for another reviewer (+ @lucasfcosta, @shvaikalesh, @vieiralucas)

lucasfcosta:
LGTM. I don't think this will ever happen in any other environment, but does make things safer anyway.

PR #986, opened 2017-06-01
Revert get version package json

### LINKED ISSUE #981: webpack throwing a warning in 4.0.0
I'm seeing the following warning with Chai 4.0.0 that I didn't have with Chai 3.5.0
```
WARNING in ./~/chai/chai.js
Critical dependencies:
1:477-484 This seems to be a pre-built javascript file. Though this is possible, it's not recommended. Try to require the original source to get better results.
 @ ./~/chai/chai.js 1:477-484
```
The use case is browser based testing from within the "storybook-addon-specifications" module.

I suspect that the problem is the new `browser` field in the "package.json" file.  Webpack is now trying to load the "chai.js" webpacked file instead of original source "index.js".  Perhaps a solution is to add a `"webpack": "index.js"` line to package.json.

I'm no webpack expert though. I would defer to someone wiser in the ways of Webpack than myself.

lucasfcosta:
This fixes #981 by not requiring a `.json` file inside our code in order for webpack not to need `json-loader`. This change was introduced #819.

This commit also reverts changes made at #967, because now we still need to bump the version in the `lib/chai.js` file.

Finally, since this seems a very important thing to release ASAP given the number of people using webpack1, I've added a commit that does this release.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ chai.js: node_modules lib/*
 #
 
 define release
-	./node_modules/.bin/bump -y --$(1) package.json
+	./node_modules/.bin/bump -y --$(1) package.json lib/chai.js
 	make chai.js
 	git add --force chai.js lib/chai.js package.json bower.json
 	npm ls --depth=-1 --long . --loglevel silent | head -1 | git commit -F-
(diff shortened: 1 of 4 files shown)


vieiralucas:
Have we thought of doing something like `JSON.parse(fs.readFileSync('../package.json')).version`?

Anyways I think that this LGTM, just wanted to raise the `fs` approach

PR #989, opened 2017-06-06
Fix Out-Of-Order Variable Declarations

samlanning:
Following #988, I decided to see if lgtm.com would be useful for enforcing chai's coding style guidelines. I'm not sure if this is the sort of thing you'd be interested in, but I'd love to hear feedback either way.

I wrote a query for variable declaration order, and you can see the results it produces for chai (and mozilla's pdf.js for comparison) here: https://lgtm.com/query/1983330216/lang:javascript/

Currently it's quite a broad query, and has many results in test code, but I can make it more specific if you'd like, and exclude certain results, e,g:
 * if there's whitespace / extra lines between groups of variable declarations, consider them different groups.
 * exclude test code
 * exclude complicated variable declarations (for some definition of complicated)

I'm not sure whether you'd like to include this query in this repository, but it does have some interesting results, some of which I've addressed in a separate commit, and we could start to write further queries. We do have plans to allow projects to run their own custom queries as part of pull request integration soon, but we don't currently have a timeline for that.

If you like this sort of contribution, I can write queries for some of your other guidelines to see what we can find.

*Full Disclosure: I'm a core developer at lgtm.com*

### DIFF
--- a/.lgtm-queries/README.md
+++ b/.lgtm-queries/README.md
@@ -1,0 +1,8 @@
+# Chai Coding Style Queries
+
+This directory contains lgtm.com queries that help enforce our
+[coding style guidelines](https://github.com/chaijs/chai/wiki/Chai-Coding-Style-Guide).
+
+Currently these can only be run manually on the latest version that lgtm.com has
+analysed, but they plan to add support for custom queries in PR integration in
+the future.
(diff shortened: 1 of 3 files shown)


keithamus:
Thanks very much for your work here @samlanning. We're aware there are a few issues with the codebase, and we have some plans to fix them a bit later - we're very likely going to use `eslint` with a configuration which will catch what this PR provides, and more (we've been trialing this out in sub-projects; [example](https://github.com/chaijs/deep-eql/blob/master/package.json#L44)).

Because of that, I think we'll close this PR. If we end up _not_ choosing eslint, we can re-open this and have another look, but right now I'm very confident we'll be sticking with eslint.

Issue #585, opened 2016-01-05
Refactor plugin declarations


keithamus:
I'm going to put this document here, as the start of a design discussion around how plugins could be written for a future version of chai. Similar discussions have been had before, over in [Google Groups](https://groups.google.com/forum/#!msg/chaijs/TXtkOAGyetU/tE0iC_WJxboJ), and issues #117 and #457; but I wanted to start a new issue which can focus on the design of a new plugin architecture, and hopefully get some feedback.

> This comment gets updated regularly to reflect the discussion below, for older version of this proposal, see https://gist.github.com/keithamus/43f7d102ffbf441107a6ae19decaca23/revisions

### Current Implementation

So, just to get everyone on track - right now, you can make a chai plugin by writing a little bit of code, like this:

``` js
export default (chai, utils) => {
  chai.Assertion.addMethod('equal', function (val, msg) {
    this.assert(
      val === obj,
      'expected #{this} to equal #{exp}',
      'expected #{this} to not equal #{exp}',
      val,
      this._obj,
      true
    );
  });

}
```

### Motivation

Right now we have a nicely expressive interface for creating plugins - but it has a few problems, which can be addressed by making working on a new interface. The problems it has are:
- It caters to `expect` and `should` interfaces as first class, but does not create methods for the `assert` interface.
- Plugins define whether or not an assertion is a property or a method - there is `addMethod`, `addProperty`, and `addChainableMethod`. These in themselves aren't bad, but it does create a bit of an inconsistent interface, as it is an opt-in convention to use property assertions and chainable methods. Similarly, for users who only want to use method assertions, they have to use plugins like [dirty-chai](https://github.com/prodatakey/dirty-chai).
- Also around convention, it's up to the plugin author whether or not a method supports reassignment of the message. Chai's core assertions allow you to overwrite the assertion message by passing in the last argument (e.g. `expect(true).to.equal(true, 'True should be true!')`). Some plugins dont do this.
- Flags are imperative - and it's up to the methods as to whether or not they support the flag, and as such they work for some methods but not for others, and there is no way to determine which assertions support which flags.
- Many plugins, such as [chai-things](https://github.com/chaijs/Chai-Things) or [chai-as-promised](https://github.com/domenic/chai-as-promised) effectively add one or two "interceptor" flags - for example [chai-as-promised](https://github.com/domenic/chai-as-promised) adds `eventually` to every method, which means programmatically rewriting every function. We should have an expressive syntax for this.
[...]

PR #990, opened 2017-06-06
Allow dates for isBelow and isAbove assertions

### LINKED ISSUE #980: Error or not updated docs on assert.isAbove and isBelow
Hi.

`assert.isAbove` and `assert.isBelow` used to take Dates as possible inputs and worked fine, but since v4.0.0 it errors with `expected *date* to be a number` resulting in numerous failed test-cases throughout my codebase after an update from 3.5.

So my guess is either the documentation was not updated (as it currently lists {Mixed} type for an input) or it's a bug - either way there seems to be a problem. Just wanted to ask which one is it.

It's reproducible both in Mocha (Node) and in the browser (dev console on the official website with `4.0.0-canary.2` at the moment of this issue)

Is there is indeed a problem or have I misunderstood something?
And how can I help?

v1adko:
Hi guys.

This closes #980. I took #692 as an "inspiration" and extended the code to support dates as discussed in the issue.

I have encountered some questions along the way and left comments (off course will delete them once we decided what to do at those moments) - I will expand on them within GitHub, to make them easier to notice. So it's a WIP for now, just need a bit of your guidance.

This is my first attempt at actually PR-ing anything open-source, so sorry for mistakes.
I'd really like to contribute :)

Also need to fix up Phantom tests, but wanted to get some feedback and ask these questions first, to understand the general direction and the scope of this PR better.


v1adko:
I have probably made a newbie mistake deleting `/chai.js`. I'm not sure how that happened, the file is in gitignore, I'd blame 'make clean`, something definitely went wrong.

lib/chai/core/assertions.js
```diff
@@ -529,16 +529,35 @@ module.exports = function (chai, _) {
...
+      , doLength = flag(this, 'doLength')
+      , objType = _.type(obj)
+      , nType = _.type(n)
+      , msgPrefix = (msg) ? msg + ': ' : '';
 
     if (doLength) {
       new Assertion(obj, msg).to.have.property('length');
     } else {
-      new Assertion(obj, msg).is.a('number');
+      // Since utils/expectTypes is missing from 4.x.x and no better way to assert multiple types
```
v1adko: I could not find a better way to assert multiple types, on the original PR `expectTypes` was suggested, but it's no longer present in the utils

lib/chai/core/assertions.js
```diff
@@ -587,6 +606,7 @@ module.exports = function (chai, _) {
    * @api public
    */
 
+  // Should this also be updated?
```
v1adko: Originally we discussed only isAbove and isBelow methods to support dates, but these seem connected as well.

test/assert.js
```diff
@@ -792,13 +792,44 @@ describe('assert', function () {
...
   });
 
+  it('above (dates)', function() {
+    var now = new Date();
+    var oneSecondAgo = new Date(now.getTime() - 1);
+    assert.isAbove(now, oneSecondAgo, 'Now should be above 1 second ago');
+
+    err(function() {
+      assert.isAbove(oneSecondAgo, now);
+    }, 'expected ' + now.toUTCString() + ' to be above ' + oneSecondAgo.toString()); // TODO: Fix different date formatting
```
v1adko: One thing I noticed is that date formatting is not uniformal. While the first argument gets printed as a UTC string, second one is not. Probably should not be this way.

keithamus:
@v1adko thanks for this!

One general note, before I review this; could you please change the base branch to `master`. We need to delete the `4.x.x` branch but haven't gotten round to it yet!

lucasfcosta:
Hi @v1adko, thank you for this PR!

I was thinking if we should actually call these `isBefore` and `isAfter` to make it more semantic.

Maybe an alias would be enough, but I'd prefer separing those two into separate assertions.

What do you think?

meeber:
@lucasfcosta @vieiralucas #980 contains a discussion regarding overloading `isBelow` and `isAbove` versus creating separate assertions. As noted in that thread, my preference is to overload in this case (with the stipulation that both the target and given value must both be the same type), and perhaps open a separate PR to alias `isBefore` and `isAfter` to them. I think that offers sufficient protection against making a mistake with types, while respecting the notion that "before" and "below" can be used interchangeably when speaking of numbers (and when thinking of dates in numeric fashion), as can "above" and "after".

vieiralucas:
@meeber Having the target and given value must both be the same type looks enough to me

test/assert.js
vieiralucas: The error message is created by `util.getMessage` inside the `Assertion.prototype.assert` [here](https://github.com/chaijs/chai/blob/616cf8bf883afa67f99a952901b4b44fcabc0cfe/lib/chai/assertion.js#L139).

The `util.getMessage` uses [util.objectDisplay](https://github.com/chaijs/chai/blob/616cf8bf883afa67f99a952901b4b44fcabc0cfe/lib/chai/utils/getMessage.js#L46) which ends up calling [util.inspect](https://github.com/chaijs/chai/blob/616cf8bf883afa67f99a952901b4b44fcabc0cfe/lib/chai/utils/objDisplay.js#L28) that for Dates [will call toUTCString()](https://github.com/chaijs/chai/blob/616cf8bf883afa67f99a952901b4b44fcabc0cfe/lib/chai/utils/inspect.js#L124) 😄 

But the `.above | .below` assertions are just [concatenating](https://github.com/chaijs/chai/blob/616cf8bf883afa67f99a952901b4b44fcabc0cfe/lib/chai/core/assertions.js#L1101) the expected value to the message which just calls `.toString()`.

In order to fix that you need to update the call to `this.assert`.

Unfortunatelly github does not allow me to comment on unchanged lines so I'll try to explain here.

Basically you need to update this:
```js
this.assert(
  obj > n
  , 'expected #{this} to be above ' + n
  , 'expected #{this} to be at most ' + n
);
```
To this:
```js
this.assert(
  obj > n
  , 'expected #{this} to be above #{exp}'
  , 'expected #{this} to be at most #{exp}'
  , n
);
```
I believe you will need to do the same thing for the .below assertion too

Please let me know if you have any further questions 😄

lib/chai/core/assertions.js
```diff
@@ -529,16 +529,35 @@ module.exports = function (chai, _) {
...
     }
 
-    if (_.type(n) !== 'number') {
-      throw new Error('the argument to above must be a number');
+    if (!(nType === 'number' || nType === 'date')) {
+      var shoudBeType = (objType === 'number') 
+        ? 'number'
+        : (objType === 'date') 
+          ? 'date' 
+          : 'number or date';
```
v1adko: Sure, no problem :) I know that there are plans to introduce a linter, would definitely be beneficial for things such cases, no more "nits" and stuff.
Also, you guys can take a look at [prettier](https://github.com/prettier/prettier) as an alternative to standard linter.

lib/chai/core/assertions.js
```diff
@@ -651,16 +671,34 @@ module.exports = function (chai, _) {
...
     }
 
-    if (_.type(n) !== 'number') {
-      throw new Error('the argument to below must be a number');
+    if (!(nType === 'number' || nType === 'date')) {
+      var shoudBeType = (objType === 'number') 
+        ? 'number'
+        : (objType === 'date') 
+          ? 'date' 
+          : 'number or date';
```
v1adko: Where would be a better place for a function like this? Utils?

lib/chai/core/assertions.js
```diff
@@ -1151,6 +1183,7 @@ module.exports = function (chai, _) {
    * @api public
    */
 
+  // Should this also be updated? + assertMost and within
```
v1adko: Originally we discussed only `isAbove` and `isBelow` methods to support dates, but this seems connected as well. Same question goes for assertMost and within. [restored]

lib/chai/core/assertions.js
```diff
@@ -1069,18 +1069,49 @@ module.exports = function (chai, _) {
...
+          undefined,
+          ssfi
+        );
+      }
     }
 
-    if (typeof n !== 'number') {
-      flagMsg = flagMsg ? flagMsg + ': ' : '';
+    if (!(nType === 'number' || nType === 'date')) {
+      var shoudBeType = (objType === 'number' || objType === 'date' )
```
v1adko: @vieiralucas Instead of doing `else-if` I got rid of the nested ternary. If you think that it contradicts a style or obscures code, let me now and I'll change this too.

Restoring the question - if I were to separate this logic elsewhere, what's the best place? Utils?

v1adko: BTW, these functions (isAbove and isBelow) are very similar apart from several strings and conditions, so if you really want to make it DRY, these can be separated into a some kind of assertion constructor that would take a predicate and some other params (like the string 'below/above').

v1adko: I also know that there are plans to introduce a linter, would definitely be beneficial for such cases, no more "nits".
Also, you guys can take a look at [prettier](https://github.com/prettier/prettier) as an alternative to standard linter. [restored]

test/expect.js
```diff
@@ -622,6 +622,46 @@ describe('expect', function () {
...
+    }, "blah: expected null to be a number or a date");
+
+    err(function () {
+      expect(now).to.be.above(null, 'blah');
+    }, "blah: the argument to above must be a date");
+
+    // Is this right? Fails with the 'implementation frames error' (same as the 4 original ones in master)
+    err(function () {
+      expect(null).to.have.length.above(0, 'blah');
+    }, "Cannot read property 'length' of null");
```
v1adko: This test fails, along with 4 tests concerning instance checks (locally on clean `master`)
Any ideas?

v1adko:
@keithamus Changed to the master base.
Also fixed up some of the comments you guys made (like the UTC strings, thanks @vieiralucas)
Still having an issue with deleted `chai.js`, several questions and one failing test that I'm not sure how to address.

lib/chai/core/assertions.js
meeber: Yup all of those should be updated as well.

lib/chai/core/assertions.js
meeber: I think this type of utility function could just be local to the assertions file.

I wouldn't be opposed to a refactoring effort to further reduce duplication among these similar functions, as long as it didn't have a big impact on readability. That'd be a discussion for a separate issue/PR though.

test/expect.js
meeber: The `.property` assertion is ungracefully failing on [this line](https://github.com/chaijs/chai/blob/616cf8bf883afa67f99a952901b4b44fcabc0cfe/lib/chai/core/assertions.js#L1662) when the target is `null`. This causes Chai's implementation frames to not be properly removed from the stack trace, which in turn causes this test using the `err` helper function to fail. I guess this problem needs a separate issue.

meeber:
@v1adko Thanks for your work on this! A `git checkout chai.js` should restore the missing file. When the test suite completes successfully, it'll replace chai.js with an updated version. We don't want that until we release a new version, so you'll need to do `git checkout chai.js` before pushing after a successful test.

lib/chai/core/assertions.js
v1adko: After also updating atMost, atLeast and within there is a definite pattern and code duplication, not only in these lines. So, I'm not sure what parts exactly to separate out, I think it's mostly the same code, just different conditions, so easier would be to create a generic assertion and "plug in" the differences.

lib/chai/core/assertions.js
```diff
@@ -1412,26 +1536,64 @@ module.exports = function (chai, _) {
...
+
+    if (!doLength && (objType !== startType || objType !== finishType)) {
+      throw new AssertionError(
+        msgPrefix + 'type mismatch, expected to within values to be ' + objType + 's',
+        undefined,
+        ssfi
+      );
+    }
+
+    // Should it check if before <= after?
```
v1adko: I noticed that within method does not check weather the two given params make a valid range. Should it? Or it's assumed that it's gonna always throw and the user would see it from the generic error message?

lib/chai/utils/expectTypes.js
```diff
@@ -30,10 +30,11 @@ module.exports = function expectTypes(obj, types, ssfi) {
   flagMsg = flagMsg ? flagMsg + ': ' : '';
 
   obj = flag(obj, 'object');
+  // TODO: Fix - results in undefined when passed date
```
v1adko: if `obj` is a date, after a flag it returns undefined, do we care?)

test/assert.js
```diff
@@ -1961,13 +1991,43 @@ describe('assert', function () {
...
+
+    assert.isAtLeast(now, oneSecondAgo, 'Now should be above one second ago');
+    assert.isAtLeast(now, now, 'Now should be equal to now');
+
+    err(function() {
+      assert.isAtLeast(now, oneSecondAfter, 'blah');
+    }, 'blah: expected ' + now.toUTCString() + ' to be at least ' + oneSecondAfter.toUTCString());
+
+    // These fail with 'implementation frames not properly filtered'
+    err(function() {
```
v1adko: Sadly I still don't know how to approach this problem.

test/assert.js
```diff
@@ -1998,13 +2088,44 @@ describe('assert', function () {
...
+      assert.isAtMost(now, 1, 'blah');
+    }, 'blah: type mismatch, expected to most value to be a date');
+
+    err(function() {
+      assert.isAtMost(1, now, 'blah');
+    }, 'blah: type mismatch, expected to most value to be a number');
+  });
+
+  // There's no assert.isWithin(n, a, b)?
```
v1adko: While should an expect implement have this options, it looks like there's no simple assert.isWithin()
Should we add an alias?

v1adko:
@meeber Updated the other methods + tests too. And checkout chai worked, thanks 👍 
Added several comments to discuss. From the previous issues there are only failing tests and a discussion on a best way to make it DRY and refactor.

Also I'm going to be out of town for the weekend without a computer, so, unfortunately, won't be able to work on this. Will gladly pick it up next week.

meeber:
@v1adko: I opened #992 to address the issue that's causing the failing tests due to implementation frames. We'll get that one merged first, and then you can rebase this branch with the change.

test/assert.js
meeber: I'd say leave that for a separate PR.

test/assert.js
meeber: Should be fixed in #992

lib/chai/core/assertions.js
meeber: Let's leave it as-is for now. Can always readdress later.

lib/chai/utils/expectTypes.js
meeber: It looks to me like `expectTypes` can only be used on assertion objects in order to make sure the target of the assertion (i.e., the value stored in the `object` flag) is one of the allowed types. It can't be used to check that arguments passed to an assertion are of the correct type; the `object` flag would be `undefined` regardless of the argument type. So you're doing the correct thing by checking the argument type manually.

v1adko: Got it, non-issue then, got confused about the use-case of this method.

v1adko:
@meeber All right, I'll wait till then

cletusw:
If we could get this in a patch update ASAP that'd be great since this fixes an unnecessary breaking change (#980) from 3.x

v1adko:
@meeber, @vieiralucas  Hi :) Don't want to rush you guys or anything, just interested in how those PRs are moving along because they look green and as I understood I need to wait for them to get merged first.

v1adko:
Ok guys, I've updated the branch - tests are passing, thank you.
Cleaned up the comments and will open issues on some of the points brought up in this PR after it get's through (isWithin alias, code duplication, extra checks, etc.)
I think this is ready for the review.

lib/chai/core/assertions.js
```diff
@@ -1156,18 +1187,48 @@ module.exports = function (chai, _) {
...
+    }
+
+    if (doLength && nType === 'date') {
       throw new AssertionError(
-        flagMsg + 'the argument to least must be a number',
+        msgPrefix + 'the argument to least must be a number',
+        undefined,
+        ssfi
+      );
+    }
```
keithamus: Same as above, I think we need to reduce the complexity here.

lib/chai/core/assertions.js
```diff
@@ -1242,18 +1304,48 @@ module.exports = function (chai, _) {
...
+    if (!(nType === 'number' || nType === 'date')) {
+      var shoudBeType = (objType === 'number' || objType === 'date' )
+        ? objType
+        : 'number or date';
+      throw new AssertionError(
+        msgPrefix + 'the argument to below must be a ' + shoudBeType,
+        undefined,
+        ssfi
+      );
+    }
```
keithamus: And again, as above.

lib/chai/core/assertions.js
```diff
@@ -1328,23 +1421,53 @@ module.exports = function (chai, _) {
...
+
+    if (!doLength && objType !== nType) {
+      throw new AssertionError(
+        msgPrefix + 'type mismatch, expected to most value to be a ' + objType,
+        undefined,
+        ssfi
+      );
+    }
```
keithamus: And again, as above.

lib/chai/core/assertions.js
```diff
@@ -1412,21 +1536,57 @@ module.exports = function (chai, _) {
...
+
+    if (doLength && (startType === 'date' || finishType === 'date')) {
+      throw new AssertionError(
+        msgPrefix + 'the arguments to within must be numbers',
+        undefined,
+        ssfi
+      );
+    }
```
keithamus: And again, but we need to check both arguments to ensure all 3 (obj, start, finish) are the same type.

test/expect.js
```diff
@@ -531,18 +531,72 @@ describe('expect', function () {
...
     err(function () {
-      expect(1).to.have.lengthOf.within(5,7, 'blah');
+      expect(1).to.have.lengthOf.within(5, 7, 'blah');
     }, "blah: expected 1 to have property 'length'");
   });
 
+  it('within(start, finish) (dates)', function(){
+    var now = new Date();
+    var oneSecondAgo = new Date(now.getTime() - 1);
+    var oneSecondAfter = new Date(now.getTime() + 1);
```
keithamus: And here.

test/expect.js
```diff
@@ -622,6 +676,45 @@ describe('expect', function () {
     }, "blah: expected 1 to have property 'length'");
   });
 
+  it('above(n) (dates)', function(){
+    var now = new Date();
+    var oneSecondAgo = new Date(now.getTime() - 1);
+    var oneSecondAfter = new Date(now.getTime() + 1);
```
keithamus: And here.

test/expect.js
```diff
@@ -787,11 +880,53 @@ describe('expect', function () {
     }, "blah: expected 1 to have property 'length'");
   });
 
+  it('below(n) (dates)', function(){
+    var now = new Date();
+    var oneSecondAgo = new Date(now.getTime() - 1);
+    var oneSecondAfter = new Date(now.getTime() + 1);
```
keithamus: And here.

test/expect.js
```diff
@@ -874,6 +1009,59 @@ describe('expect', function () {
     }, "blah: expected 1 to have property 'length'");
   });
 
+  it('most(n) (dates)', function(){
+    var now = new Date();
+    var before = new Date(now.getTime() - 1);
+    var after = new Date(now.getTime() + 1);
```
keithamus: And here.

test/should.js
```diff
@@ -545,6 +545,64 @@ describe('should', function() {
     }, "blah: expected 1 to have property 'length'");
   });
 
+  it('within(start, finish) (dates)', function(){
+    var now = new Date();
+    var before = new Date(now.getTime() - 1);
+    var after = new Date(now.getTime() + 1);
```
keithamus: And here.

test/should.js
```diff
@@ -592,6 +650,49 @@ describe('should', function() {
     }, "blah: expected 1 to have property 'length'");
   });
 
+  it('above(n) (dates)', function(){
+    var now = new Date();
+    var oneSecondAgo = new Date(now.getTime() - 1);
+    var oneSecondAfter = new Date(now.getTime() + 1);
```
keithamus: And here.

test/should.js
```diff
@@ -676,6 +777,49 @@ describe('should', function() {
     }, "blah: expected 1 to have property 'length'");
   });
 
+  it('below(n) (dates)', function(){
+    var now = new Date();
+    var oneSecondAgo = new Date(now.getTime() - 1);
+    var oneSecondAfter = new Date(now.getTime() + 1);
```
keithamus: And here.

test/should.js
```diff
@@ -721,6 +865,59 @@ describe('should', function() {
     }, "blah: expected 1 to have property 'length'");
   });
 
+  it('most(n) (dates)', function(){
+    var now = new Date();
+    var before = new Date(now.getTime() - 1);
+    var after = new Date(now.getTime() + 1);
```
keithamus: And here.

test/assert.js
```diff
@@ -1989,13 +2048,43 @@ describe('assert', function () {
...
+    }, 'blah: expected null to be a number or a date');
 
     err(function() {
       assert.isBelow(1, null, 'blah');
     }, 'blah: the argument to below must be a number');
   });
 
+  it('below (dates)', function() {
+    var now = new Date();
+    var oneSecondAgo = new Date(now.getTime() - 1);
```
keithamus: And here.

test/assert.js
```diff
@@ -2006,13 +2095,42 @@ describe('assert', function () {
...
 
     err(function() {
       assert.isAtMost(1, null, 'blah');
     }, 'blah: the argument to most must be a number');
   });
 
+  it('atMost (dates)', function() {
+    var now = new Date();
+    var oneSecondAgo = new Date(now.getTime() - 1);
+    var oneSecondAfter = new Date(now.getTime() + 1);
```
keithamus: And here.

test/assert.js
```diff
@@ -1969,13 +1999,42 @@ describe('assert', function () {
...
 
     err(function() {
       assert.isAtLeast(1, null, 'blah');
     }, 'blah: the argument to least must be a number');
   });
 
+  it('atLeast (dates)', function() {
+    var now = new Date();
+    var oneSecondAgo = new Date(now.getTime() - 1);
+    var oneSecondAfter = new Date(now.getTime() + 1);
```
keithamus: Same here.

test/assert.js
```diff
@@ -1952,13 +1952,43 @@ describe('assert', function () {
...
+    }, 'blah: expected null to be a number or a date');
 
     err(function() {
       assert.isAbove(1, null, 'blah');
     }, 'blah: the argument to above must be a number');
   });
 
+  it('above (dates)', function() {
+    var now = new Date();
+    var oneSecondAgo = new Date(now.getTime() - 1);
```
keithamus: These will actually be 1ms after. We should change the `1` to `1000` to make them actually 1 second different.

keithamus:
@v1adko thanks for this. I've made a few comments - really just 2 changes - but in multiple places. Could you please address these? Then I think we're good to merge!

v1adko:
@keithamus Hah, dumb mistake with milliseconds - fixed it, thanks. :)
Also I've refactored one of the methods - [here](https://github.com/chaijs/chai/pull/990/files#diff-6f8be24f0197a0249fde03c55e7397dbR1080)
If you're OK with this approach - I'm gonna quickly update the rest of the methods in the same manner.
The reason I had to do the checks outside of `(doLength) {} else { ... }` is that obviously it's always false there.
You're right that this looks much cleaner, I just didn't want to break previous tests (when null is passed), and we've lost some possibly useful information like [here](https://github.com/chaijs/chai/pull/990/files#diff-110fc4dec7399d3597e41d0966029d3cL572), but all in all I think it's worth it.

keithamus:
I think I could live with that refactor 😄. Hopefully #585 will eventually get us to a point where this kind of manual type checking isn't necessary any more.

test/expect.js
```diff
@@ -454,53 +454,53 @@ describe('expect', function () {
...
 
     err(function () {
-      expect([ 1, 2, 3 ]).to.have.lengthOf.within(5,7, 'blah');
+      expect([ 1, 2, 3 ]).to.have.lengthOf.within(5, 7, 'blah');
     }, "blah: expected [ 1, 2, 3 ] to have a length within 5..7");
 
     err(function () {
       expect(null).to.be.within(0, 1, 'blah');
-    }, "blah: expected null to be a number");
+    }, "blah: the arguments to within must be numbers or dates");
```
v1adko: @keithamus This doesn't seem right to me. Technically here both arguments to within are numbers, argument to expect isn't so the message is not very accurate.

v1adko:
@keithamus I also feel like [L1082-L1094](https://github.com/v1adko/chai/blob/allow-below-dates-assertion/lib/chai/core/assertions.js#L1082-L1094) can be extracted into a separate function like `checkNumberOrDate`, what's your take on that?
It's gonna reduce duplication in 4 assertion functions (within is custom).
What's your take on this?

test/expect.js
keithamus: Yes this message is a little vague. Can we add an extra `if` to just throw saying something like "blah: expected null to be a number or date"

keithamus:
Thanks @v1adko. While the simplicity of creating assertions using the helper is definitely somewhere we want to be in future, I think for now having the repetitive yet easily understandable code in the PR as it stands now works well. In addition, #585 grants us the scope for these kinds of optimisations, where we can truly focus on getting chai assertions to be very easy to write.

Let's just fix up the small issue you mentioned above, then get this PR merged!

keithamus:
Perfect! Thanks @v1adko this looks good to me!

@meeber, @lucasfcosta @vieiralucas as you've commented, do you all want to approve before we move this to merge?

v1adko:
@keithamus Done

test/assert.js
```diff
@@ -1952,13 +1952,43 @@ describe('assert', function () {
...
+
+    err(function() {
+      assert.isAbove(now, 1);
+    }, 'the argument to above must be a date');
+
+    err(function() {
+      assert.isAbove(1, now, 'blah');
+    }, 'blah: the argument to above must be a number');
+  });
+
```
meeber: Let's add the custom message (`blah`) testing to the other failure cases here.

test/assert.js
```diff
@@ -1989,13 +2048,43 @@ describe('assert', function () {
...
+    err(function() {
+      assert.isBelow(now, 1);
+    }, 'the argument to below must be a date');
+
+    err(function() {
+      assert.isBelow(1, now, 'blah');
+    }, 'blah: the argument to below must be a number');
+  });
+
```
meeber: Let's add the custom message (`blah`) testing to the other failure cases here.

test/should.js
```diff
@@ -545,6 +545,64 @@ describe('should', function() {
...
+
+    err(function () {
+      (now).should.not.be.within(null, now, 'blah');
+    }, "blah: the arguments to within must be dates");
+
+    err(function () {
+      (now).should.not.be.within(now, null, 'blah');
+    }, "blah: the arguments to within must be dates");
+  });
+
```
meeber: Don't need parenthesis around `<var name>` whenever it's `(<var name>).should`; still need parenthesis for the values plugged directly, like `(0).should`

test/should.js
```diff
@@ -592,6 +650,49 @@ describe('should', function() {
...
+    err(function () {
+      (1).should.not.be.above(now, 'blah');
+    }, "blah: the argument to above must be a number");
+
+    err(function () {
+      ([]).should.have.length.above(now, 'blah');
+    }, "blah: the argument to above must be a number");
+  });
+
```
meeber: Don't need parenthesis around `<var name>` whenever it's `(<var name>).should`; still need parenthesis for the values plugged directly, like `(0).should`

test/should.js
```diff
@@ -676,6 +777,49 @@ describe('should', function() {
...
+    err(function () {
+      (now).should.not.be.below(null, 'blah');
+    }, "blah: the argument to below must be a date");
+
+    err(function () {
+      ('string').should.have.length.below(now, 'blah');
+    }, "blah: the argument to below must be a number");
+  });
+
```
meeber: Don't need parenthesis around `<var name>` whenever it's `(<var name>).should`; still need parenthesis for the values plugged directly, like `(0).should`

test/should.js
```diff
@@ -721,6 +865,59 @@ describe('should', function() {
...
+    err(function () {
+      (1).should.not.be.at.most(now, 'blah');
+    }, "blah: the argument to most must be a number");
+
+    err(function () {
+      (now).should.not.be.at.most(undefined);
+    }, "the argument to most must be a date");
+  });
+
```
meeber: Don't need parenthesis around `<var name>` whenever it's `(<var name>).should`; still need parenthesis for the values plugged directly, like `(0).should`

lib/chai/core/assertions.js
```diff
@@ -1197,7 +1215,7 @@ module.exports = function (chai, _) {
   /**
    * ### .below(n[, msg])
    *
-   * Asserts that the target is a number less than the given number `n`.
+   * Asserts that the target is a number or a date less than the given number or a date `n` respectively.
```
v1adko: @meeber I don't see how what you wrote is different from what's already there (green).
Or is it supposed to be the same? Sorry, I'm a bit confused. What's would be the suggestions to make it read better?)

meeber: @v1adko In almost every case, the suggestion is "given number or date" instead of "given number or a date". In addition to that, the one for `.above` is also missing an "a" at the start.

v1adko: @meeber So no articles then?

test/should.js
v1adko: Ok, if I understood correctly, just omit parenthesis and do `now.should` instead of `(now).should`?

test/assert.js
v1adko: @meeber All of them or some of them? I've seen this message omitted in some other tests.

lib/chai/core/assertions.js
meeber: Articles for the first half (either "the target is a number or a date", or, "the target is a number or date"). But no articles for the second half ("than the given number or date").

test/should.js
meeber: Yup!

v1adko:
@meeber Done. As for the commits - I think it's better to squash them via GitHub.

test/assert.js
meeber: At the minimum, it'd be good for each unique error case in an assertion (e.g., "must be a number" is one case and "must be a date" is another) to have a test that verifies that custom messages are respected. As for the message being omitted in other tests, some were inadvertently missed in #947 which is where many such tests were added (after being excluded in the original, historical PRs).

meeber:
@v1adko The tests for `assert.isAbove` need that same custom message loving, but otherwise this looks good to go!

v1adko:
@meeber Missed that :)
Looks like we're waiting for @lucasfcosta to sign off and it's done. 🎉

keithamus:
🎉 ✨

cletusw:
Awesome! Can we get a release for this? Thanks.

lucasfcosta:
@cletusw Let's see what other people think, but IMO we should.

In order to avoid having to wait for too long for a release like we did last time I think it would be awesome for us to keep releasing new versions with `patch` and `minor` changes as soon as possible. This also improves feedback time and we can fix things more quickly.

What do you think?

keithamus:
Eventually we'll hopefully end up adding semantic-release to this repo like the others; in the meantime anyone is welcome to raise a PR of the output of `make release`, and we can merge & tag it ✨

PR #992, opened 2017-06-10
fix: check target's type in `.property` assertion

meeber:
Previously, the `.property` assertion failed ungracefully if the target was `null` or `undefined`. This commit causes an `AssertionError` to be thrown instead so that the ssfi flag and custom error messages are respected.

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -1644,6 +1644,7 @@ module.exports = function (chai, _) {
     var isNested = flag(this, 'nested')
       , isOwn = flag(this, 'own')
       , flagMsg = flag(this, 'message')
+      , obj = flag(this, 'object')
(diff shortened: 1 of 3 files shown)


keithamus:
Do we want to consider this a bugfix or a breaking change?

meeber:
@keithamus I vote bugfix. The `.property` assertion was already incompatible with `null` and `undefined` because it was throwing a `TypeError` on [this line](https://github.com/chaijs/chai/blob/616cf8bf883afa67f99a952901b4b44fcabc0cfe/lib/chai/core/assertions.js#L1662). This PR just replaces the `TypeError` with an `AssertionError`.

vieiralucas:
I vote bugfix too, nobody should be relying on this throwing a `TypeError` anyways.

PR #994, opened 2017-06-21
Add ES6 collection support to include()

### LINKED ISSUE #970: expect(...).to.include should support ES2015 Sets and Maps
It seems like Chai supports key assertions on ES2015 sets & maps (introduced in this PR: https://github.com/chaijs/chai/pull/633), but it does not support simple "inclusion" assertions.

Consider the following testcase:
```javascript
it("should detect item existence in Set", function () {
    let container = new Set();

    container.add(123);

    expect(container).to.include(123);
});
```

This test should obviously pass, as the ``container`` has ``123`` value added to it. Instead, it fails with the following error:
```
AssertionError: object tested must be an array, an object, or a string, but set given
```

Same thing happens if we change ``Set`` to ``Map`` (``container.set("somekey", 123)``).

---

As far as I know, this problem occurs both in ``3.5.0`` version, as well as the latest ``4.0.0-canary.2`` pre-release version.

shvaikalesh:
Closes #970

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -453,21 +453,16 @@ module.exports = function (chai, _) {
     flag(this, 'contains', true);
   }
 
-  function isDeepIncluded (arr, val) {
-    return arr.some(function (arrVal) {
-      return _.eql(arrVal, val);
-    });
-  }
-
   function include (val, msg) {
     if (msg) flag(this, 'message', msg);
 
-    _.expectTypes(this, ['array', 'object', 'string']);
+    _.expectTypes(this, [
+      'array', 'object', 'string',
+      'map', 'set', 'weakmap', 'weakset',
+    ]);
 
     var obj = flag(this, 'object')
-      , objType = _.type(obj).toLowerCase()
-      , isDeep = flag(this, 'deep')
-      , descriptor = isDeep ? 'deep ' : '';
+      , objType = _.type(obj).toLowerCase();
 
     // This block is for asserting a subset of properties in an object.
     if (objType === 'object') {
@@ -504,10 +499,42 @@ module.exports = function (chai, _) {
       return;
     }
 
-    // Assert inclusion in an array or substring in a string.
+    var isDeep = flag(this, 'deep')
+      , descriptor = isDeep ? 'deep ' : ''
+      , included = false;
+
+    switch (objType) {
+      case 'string':
+        included = obj.indexOf(val) !== -1;
+        break;
+
+      case 'weakmap':
+      case 'weakset':
+        if (isDeep) {
+          var flagMsg = flag(this, 'message')
+            , ssfi = flag(this, 'ssfi');
+          flagMsg = flagMsg ? flagMsg + ': ' : '';
+
+          throw new AssertionError(
+            flagMsg + 'unable to use .deep.include with weak collection',
+            undefined,
+            ssfi
+          );
+        }
+
+        included = obj.has(val);
+        break;
+
+      default:
+        var isEql = isDeep ? _.eql : function(a, b) { return a === b; };
+        obj.forEach(function (item) {
+          included = included || isEql(item, val);
+        });
+    }
+
+    // Assert inclusion in collection or substring in a string.
     this.assert(
-        objType === 'string' || !isDeep ? ~obj.indexOf(val)
-                                        : isDeepIncluded(obj, val)
+      included
       , 'expected #{this} to ' + descriptor + 'include ' + _.inspect(val)
       , 'expected #{this} to not ' + descriptor + 'include ' + _.inspect(val));
   }
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -462,7 +462,7 @@ module.exports = function (chai, _) {
 
     _.expectTypes(this, [
       'array', 'object', 'string',
-      'map', 'set', 'weakmap', 'weakset',
+      'map', 'set', 'weakset',
     ]);
 
     var obj = flag(this, 'object')
@@ -512,7 +512,6 @@ module.exports = function (chai, _) {
         included = obj.indexOf(val) !== -1;
         break;
 
-      case 'weakmap':
       case 'weakset':
         if (isDeep) {
           var flagMsg = flag(this, 'message')
@@ -520,7 +519,7 @@ module.exports = function (chai, _) {
           flagMsg = flagMsg ? flagMsg + ': ' : '';
 
           throw new AssertionError(
-            flagMsg + 'unable to use .deep.include with weak collection',
+            flagMsg + 'unable to use .deep.include with WeakSet',
             undefined,
             ssfi
           );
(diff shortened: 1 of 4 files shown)


lib/chai/core/assertions.js
```diff
@@ -504,10 +499,42 @@ module.exports = function (chai, _) {
...
+            undefined,
+            ssfi
+          );
+        }
+
+        included = obj.has(val);
+        break;
+
+      default:
+        var isEql = isDeep ? _.eql : function(a, b) { return a === b; };
```
shvaikalesh: I feel like we should use `SameValue` here, but it is minor breaking change.

lib/chai/core/assertions.js
```diff
@@ -504,10 +499,42 @@ module.exports = function (chai, _) {
...
+      , included = false;
+
+    switch (objType) {
+      case 'string':
+        included = obj.indexOf(val) !== -1;
+        break;
+
+      case 'weakmap':
+      case 'weakset':
+        if (isDeep) {
```
shvaikalesh: I think we should consider containing code like this into a private method: it is easy to forget to put `flagMsg`, pass `ssfi`, or skip second argument.

keithamus: Agreed, but I'd like to see focus on #585 to get the assertion functions to be pure functions returning objects, so we can wrap them up in assertionerrors

lib/chai/core/assertions.js
keithamus: Well this is a new feature right? So no one would have hit this path before. It would be a small gotcha to say:

> If you pass in (weak)?(map|set) then it'll use samevalue, otherwise ===

keithamus: Yes - we should look to change many of these instances for chai@5 i think

keithamus:
I'm very happy with this, but let's get another review to sign off before merge. @meeber @lucasfcosta?

lib/chai/core/assertions.js
```diff
@@ -504,10 +499,42 @@ module.exports = function (chai, _) {
...
+        }
+
+        included = obj.has(val);
+        break;
+
+      default:
+        var isEql = isDeep ? _.eql : function(a, b) { return a === b; };
+        obj.forEach(function (item) {
+          included = included || isEql(item, val);
+        });
```
vieiralucas: Can we do something here to optimize this by breaking out the loop once `included` is true?

I realize that then we would need to have 3 implementations, one for `Array`, another for `Map` and another for `Set`.

Do you think it is worth it?

shvaikalesh: We can optimize away looping through `map` and `set` for non-deep `include` (cool) and use `some` + `indexOf` on arrays. Code:

```diff
+  case 'map':
+  case 'set':
+    if (isDeep) {
+      obj.forEach(function (item) {
+        included = included || _.eql(item, val);
+      });
+   } else {
+     included = obj.has(val);
+   }
+   break;
+
+  case 'array':
+    if (isDeep) {
+     included = obj.some(function (item) {
+       return _.eql(item, val);
+     })
+    } else {
+      included = obj.indexOf(val) !== -1;
+    }
+    break;

-  default:
-    var isEql = isDeep ? _.eql : function(a, b) { return a === b; };
-    obj.forEach(function (item) {
-      included = included || isEql(item, val);
-    });
```

shvaikalesh:
One more thing: should we throw when `deep.include` is called on string? Currently we just ignore the flag, but the assertion does not make sense.

lib/chai/core/assertions.js
keithamus: This looks good but I'd be a little careful here with IE11 - which has terrible support for Map/Set. It probably will work but I've often been surprised how seemingly obvious features are missing from IE11's Map/Set

keithamus:
@shvaikalesh I think we should - as a general future feature consideration - start adding warnings for wonky assertion chains, rather than explicitly erroring. Let's leave it out of this PR and discuss it in https://github.com/chaijs/chai/issues/996

shvaikalesh:
I've updated the docs and dropped `WeakMap` support -- we can't iterate over it's values to make an assertion.

Regarding `deep` flag: point raised by @meeber in https://github.com/chaijs/chai/issues/996#issuecomment-310481585 is valid for `WeakSet` targets too. I am fine with silently ignoring it.

lib/chai/core/assertions.js
```diff
@@ -504,10 +513,62 @@ module.exports = function (chai, _) {
...
+            ssfi
+          );
+        }
+
+        included = obj.has(val);
+        break;
+
+      case 'map':
+        var isEql = isDeep ? _.eql : SameValueZero;
+        obj.forEach(function (item) {
```
lucasfcosta: As @vieiralucas said, we might be able to optimize this by getting out of the loop when we found the included item.

In order to support IE and older versions of it we can `throw` an error inside the `forEach` callback and catch it.

This is a cross-platform compatible implementation I've made on `Sinon` for the `every` method. We can inspire our `some` implementation in that. However, even though I'd like it to be done, I won't block merging this PR because of that.

lib/chai/core/assertions.js
```diff
@@ -330,6 +330,16 @@ module.exports = function (chai, _) {
    *
    *     expect({a: 1, b: 2, c: 3}).to.include({a: 1, b: 2});
    *
+   * When the target is a Set or WeakSet, `.include` asserts that the given `val` is a
+   * member of the target. SameValueZero equality algorithm is used.
+   *
+   *     expect(new Set([1, 2])).to.include(2);
+   *
+   * When the target is a Map, `.include` asserts that the given `val` is one of
+   * the values of the target. SameValueZero equality algorithm is used.
```
lucasfcosta: I think it would be good to explain how the `SameValueZero` algorithm works. I think many people could get confused by that and would end up either having to read the code to understand it or just avoiding using this assertion because they're not sure of what it does.

meeber: @lucasfcosta Fair point. In this case, I would suggest moving the phrase "SameValueZero equality algorithm is used" from where it is currently down to the separate paragraph where `strict` and `deep` equality` is discussed. This would make the doc style more consistent anyway. And then rather than explain what "SameValueZero equality" is, we could provide a link to an explanation, just like we do with the "deep equal" algorithm.

lib/chai/core/assertions.js
meeber: @lucasfcosta I think this is worth exploring. I'd prefer this kind of optimization technique be implemented in a separate PR as a refactor.

lib/chai/core/assertions.js
keithamus: We could just refer to is as it is more commonly known; [`Object.is`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is) or if you'd like to we coud keep the phrasing but link to [the equality article on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness) which mentions `SameValueZero`.

lib/chai/core/assertions.js
lucasfcosta: @meeber agreed.

LGTM! Let's get this merged.

PR #1005, opened 2017-07-09
test: skip failing `.include` tests in IE11

meeber:
Contrary to spec, IE11 uses SameValue instead of SameValueZero
equality for Sets.

### DIFF
--- a/test/assert.js
+++ b/test/assert.js
@@ -657,7 +657,11 @@ describe('assert', function () {
 
       assert.include(set, val);
       assert.include(set, 2);
-      assert.include(set, 0);
+      if (set.has(0)) {
+        // This test is skipped in IE11 because (contrary to spec) IE11 uses
+        // SameValue instead of SameValueZero equality for sets.
+        assert.include(set, 0);
+      }
       assert.include(set, NaN);
     }
 
--- a/test/expect.js
+++ b/test/expect.js
@@ -1910,7 +1910,11 @@ describe('expect', function () {
       expect(set).to.not.include([{a: 1}]);
       expect(set).to.include(2);
       expect(set).to.not.include(3);
-      expect(set).to.include(0);
+      if (set.has(0)) {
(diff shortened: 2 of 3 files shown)


meeber:
@keithamus @lucasfcosta @vieiralucas Hey guys, are any of you around? I'd like to get a second approval on this small PR so we can (hopefully) have a passing build before releasing v4.1.

keithamus:
This feels wrong to me. The point of having an assertion library is to smooth out these bugs. If IE11 misbehaves we should be fixing it. If we feel like we want to get 4.1 out, then we can merge this for now - but I'd expect us to quickly be releasing a 4.1.1 to fix this properly.

meeber:
@keithamus Here is the problem (using plain JavaScript, with Chai removed from the equation):

```js
var blah = new Set();
blah.add(-0);

blah.has(0); // IE11: false; others: true

blah.add(0);

blah.size; // IE11: 2, others: 1
```

In the example above, the answer to these two questions is different between IE11 and other browsers:

- Is `0` included in the Set after `-0` is added?
- What's the size of the Set after both `0` and `-0` are added?

From Chai's perspective, we seem to have two options:

- Accurately reflect whether or not `0` or `-0` is included in the given set based on the current environment
- Accurately reflect whether or not `0` or `-0` is included in the given set in most browsers, but lie about it in IE11

We're faced with the same dilemma once we add a `.size` assertion for checking the size of Sets. Should such a Chai assertion consider the size of the Set above to be 1 or 2 in IE11?

My currently feeling is that Chai should accurately reflect what's happening in the engine. This would give a tester a notice that their code is behaving differently (in an unexpected way) between different engines, and that they likely need to change it.

keithamus:
> My currently feeling is that Chai should accurately reflect what's happening in the engine. This would give a tester a notice that their code is behaving differently (in an unexpected way) between different engines, and that they likely need to change it.

Yes. This rebuttal convinces me then. Frustrating how even modern IEs still plague us with issues 😞

PR #1008, opened 2017-07-22
fix: `.instanceof` to allow DOM interfaces in IE11

### LINKED ISSUE #1000: chai4 regression: instanceOf validation fails for native HTMLElement instances on IE11
Starting with chai v4, instanceOf now validates the target as a constructor function.
The assertion resides in: https://github.com/chaijs/chai/blob/master/lib/chai/core/assertions.js#L1604
```
    var validInstanceOfTarget = constructor === Object(constructor) && (
        typeof constructor === 'function' ||
        (typeof Symbol !== 'undefined' &&
         typeof Symbol.hasInstance !== 'undefined' &&
         Symbol.hasInstance in constructor)
    );
```
This validation fails on IE11, IE10, and Safari 7 when the constructor is HTMLInputElement.
I'm guessing it's broken for other HTMLElement constructors, and other (old) environments as well.

Sample code which shows the failure:
https://github.com/wix/wix-react-components/blob/master/test/components/number-input.spec.tsx#L17

Failure log:
https://travis-ci.org/wix/wix-react-components/builds/248446200

Notes:
- `typeof HTMLInputElement === "object"` on IE11.
- `<number-input-dom-instance> instanceof HTMLInputElement === true` on IE11
- I'm loading core-js/shim to the client, but I've tested the validation a a clean IE11 (about:blank) and it fails even without the shim.

Same assertion worked with chai 3, as there was no validation.

meeber:
Fixes #1000

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -1601,28 +1601,25 @@ module.exports = function (chai, _) {
     var target = flag(this, 'object')
     var ssfi = flag(this, 'ssfi');
     var flagMsg = flag(this, 'message');
-    var validInstanceOfTarget = constructor === Object(constructor) && (
-        typeof constructor === 'function' ||
-        (typeof Symbol !== 'undefined' &&
-         typeof Symbol.hasInstance !== 'undefined' &&
-         Symbol.hasInstance in constructor)
-    );
 
-    if (!validInstanceOfTarget) {
-      flagMsg = flagMsg ? flagMsg + ': ' : '';
-      var constructorType = constructor === null ? 'null' : typeof constructor;
-      throw new AssertionError(
-        flagMsg + 'The instanceof assertion needs a constructor but ' + constructorType + ' was given.',
-        undefined,
-        ssfi
-      );
+    try {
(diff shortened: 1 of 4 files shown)


keithamus:
@shvaikalesh @lucasfcosta @vieiralucas want to put a second 👍 on this?

lib/chai/core/assertions.js
```diff
@@ -1601,28 +1601,26 @@ module.exports = function (chai, _) {
...
-        flagMsg + 'The instanceof assertion needs a constructor but ' + constructorType + ' was given.',
-        undefined,
-        ssfi
-      );
+    try {
+      var isInstanceOf = target instanceof constructor;
+    } catch (err) {
+      if (err instanceof TypeError) {
+        flagMsg = flagMsg ? flagMsg + ': ' : '';
+        var type = constructor === null ? 'null' : typeof constructor;
```
shvaikalesh: maybe we can do `_.type(constructor)` here?

meeber:
@shvaikalesh Updated!

lucasfcosta:
LGTM too! Good job on that commenting explaining this change.

PR #1012, opened 2017-07-30
fix: `.include` to work with all objects

### LINKED ISSUE #1009: deep.equal() and deep.include() doesn't work with Error
`deep.equal()` and `deep.include()` doesn't compare the object keys as I expect it to. I found the issue when upgrading to v4 as the `deep.equal()` works fine in v3.5. It stopped working in v4+.

Here is a code to confirm the issue: https://jsfiddle.net/wLzrucae/3/.

I tried adding `error` to the allowed types for `include()` assertion, but it still failed on comparison. Any ideas?

meeber:
Overly strict type-checking was causing `.include` to reject `Error`
objects and objects with a custom `@@toStringTag`.

Fixes #1009

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -469,53 +469,18 @@ module.exports = function (chai, _) {
 
   function include (val, msg) {
     if (msg) flag(this, 'message', msg);
-
-    _.expectTypes(this, [
-      'array', 'object', 'string',
-      'map', 'set', 'weakset',
-    ]);
-
+    
     var obj = flag(this, 'object')
-      , objType = _.type(obj).toLowerCase();
+      , objType = _.type(obj).toLowerCase()
+      , flagMsg = flag(this, 'message')
+      , negate = flag(this, 'negate')
+      , ssfi = flag(this, 'ssfi')
+      , isDeep = flag(this, 'deep')
+      , descriptor = isDeep ? 'deep ' : '';
 
-    // This block is for asserting a subset of properties in an object.
-    if (objType === 'object') {
-      var props = Object.keys(val)
-        , negate = flag(this, 'negate')
-        , firstErr = null
-        , numErrs = 0;
+    flagMsg = flagMsg ? flagMsg + ': ' : '';
 
-      props.forEach(function (prop) {
-        var propAssertion = new Assertion(obj);
-        _.transferFlags(this, propAssertion, true);
-        flag(propAssertion, 'lockSsfi', true);
-
-        if (!negate || props.length === 1) {
-          propAssertion.property(prop, val[prop]);
-          return;
-        }
-
-        try {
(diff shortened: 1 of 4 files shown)


lib/chai/core/assertions.js
```diff
@@ -469,65 +469,37 @@ module.exports = function (chai, _) {
...
-      // to throw an assertion error if all of the properties are included,
-      // in which case we throw the first property assertion error that we
-      // encountered.
-      if (negate && props.length > 1 && numErrs === props.length) throw firstErr;
+    flagMsg = flagMsg ? flagMsg + ': ' : '';
 
-      return;
+    // `_.expectTypes` isn't used here because `.include` should work with
+    // objects with a custom `@@toStringTag`.
+    if ((objType !== 'string' && objType !== 'object') || objType === 'null') {
```
shvaikalesh: Maybe we should move this check to `default` to eliminate `!== 'string'` check and use `val === Object(val)` to avoid `=== 'null'` stuff and allow functions?

meeber:
@shvaikalesh @vieiralucas Updated.

shvaikalesh:
Great job, @meeber. LGTM.

metareven:
@meeber 

This does not work. The following object still gives the error message: 

> AssertionError: object tested must be an array, a map, an object, a set, a string, or a weakset, but object given

When using:
`expect(object).to.include('notReady')`

```
{ id: 674,
  title: 'El Globo',
  floors: [ 4 ],
  topFloors: 0,
  ui: 
   { statusBarMenuVisible: false,
     statusBarMenuInitial: true,
     loggedInUsersMenuVisible: false,
     loggedInUsersMenuInitial: true,
     lockLabelOn: false },
  status: 
   { isPublishable: true,
     isBeingPublished: false,
     isDirty: false,
     published: false,
     stored: false },
  itemId: 1,
  createdBy: 'kurator2',
  createdDate: 1476885553000,
  latest: true,
  live: false,
  parentId: undefined,
  notReady: 
   { isHouseLockedByAnother: true,
     isAlreadyPublished: false,
     isAlreadyStored: false,
     isSingleFloor: false,
     isTemporaryFloor: false,
     isSingleRoom: false },
  type: undefined 
}
```

Tested on both chai 4.1.1 and chai 4.1.2

meeber:
@metareven According to [the docs](http://www.chaijs.com/api/bdd/#method_include), I don't think you can use `.include` in this way. Since your target is an object, the value you pass has to be an object too, not a string. This should work instead for your usage: `expect(object).to.have.property('notReady');`.

With that said, something is clearly not right with this error message and this PR in general. It's been quite awhile since I looked at the Chai codebase, but I'm thinking `if (val !== Object(val)) {` should be `if (obj !== Object(obj)) {` instead, and then a separate check is needed afterward to make sure `val` is an object as well with a more precise error message if it's not.

Issue #930, opened 2017-02-09
Chai needds a better way to assert the existence of an error


gabegattis:
I often find myself doing this:
```js
should.exist(err);
err.should.be.an.instanceOf(Error);
err.message.should.equal('this is an error');
```

This feels pretty awkward. Is there a better way to do this?

If not, something like
```js
should.existError('this is an error');
```
would be great.

keithamus:
@gabegattis thanks for the issue.

You could use chaining here to make this look a little nicer, and also read more like natural language:

```js
should.exist(err).and.be.an.instanceOf(Error).with.property('message', 'this is an error');
```

Hope this is sufficient for you. I'll close this for now, but let's continue the discussion if you have any more thoughts 😄

gabegattis:
Doing it that way would put everything in one statement, but it is still about the same amount of typing I have to do. That is one of the main things that I am concerned with.

Having a convenience method for this case would make things easier. I understand that we don't want a convenience method for every conceivably use case, but this kind of error checking seems like a very common use case, so I think it is justified.

It would also encourage devs to do more thorough error checking. I have seen tests where someone just did `should.exist(err)` without checking type and message, and the test eventually concealed bugs where `err` was either not an error or was an unexpected error.

I would be happy to make a pull request if you are willing to merge something like this.

meeber:
Can your tests be refactored to use the `throw` assertion (documented [here](https://github.com/chaijs/chai/blob/f937ee5be0f7e468e195522a37bc41cd7329a294/lib/chai/core/assertions.js#L1505-L1612))? Or are you provided the `err` object in a different way?

gabegattis:
I do use the `throw` assertion a lot if I am testing a function that I expect to throw. That is not what this issue is about though. This issue is mainly for functions that callback an error.

Something like
```js
it('should error if given an invalid param', function(done) {
  database.findAccount('thisIsAnInvalidParam', function(err, account) {
    should.exist(err);
    err.should.be.an.instanceOf(Error);
    err.message.should.equal('this is an error');
  });
});
```

meeber:
Gotcha. It makes sense to me for Chai to have an assertion that's exactly like `.throw` except the target is an existing error object. And then `.throw` can use that assertion internally, so there's no duplication between the two.

```js
except(myErr).to.be.an.error();
except(myErr).to.be.an.error(TypeError);
except(myErr).to.be.an.error(myErr); // No point in doing this instead of .equal, but supported by .throws
[...]

PR #1015, opened 2017-08-02
WIP: Update `.throw` and add `.error`

meeber:
# **DO NOT MERGE; THIS IS A WORK IN PROGRESS!!**

I've been doing a lot of work on #930. This turned out to be a lot harder than expected; `.throw` couldn't be updated to simply call `.error` internally due to the difference in failed assertion messages between the two. I suspect this is the same kind of issue causing problems in plugins like `chai-as-promised` when they try to adapt `.throw`.

The more I looked at our current `.throw` implementation, the more I realized that the complexity wasn't because of negation as we originally thought but rather because of how the failed assertion messages were being composed. As a result, this PR turned into a pretty big refactor of `.throw` to reduce complexity. This also required changing some of `.throw`'s failed assertion messages to be more consistent.

Some notes/thoughts:

- This PR currently only adds `.error` to the BDD interface. I'll add the Assert interface later.
- Is changing a failed assertion message considered a breaking change? On the one hand, it's likely to break the tests of any plugins that do string matching against failed assertion messages. On the other hand, it's unlikely to break the actual functionality of the plugin from the end user's perspective, or any of the end user's tests. I'm currently learning toward "no".
- The `matchError` and `decribeExpectedError` helper functions should probably be moved to the `check-error` module.
- Part of what `.throw` does is assert that a function throws, without taking into account what it throws. This means it will work if the target function throws something besides an `Error` object (e.g., a string, a number, `null`, or even `undefined`). I think it's important that the `.throw` logic retain this base functionality. However, when it comes to adding arguments to the `.throw` function to perform additional assertions on the value that was thrown, I think we should consider requiring that the object thrown be an `Error` object. Currently, the behavior of the various `check-error` functions is a little sketchy if the object being tested isn't an `Error`. This problem carries over into the `matchError` helper function as well.

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -2380,20 +2380,24 @@ module.exports = function (chai, _) {
    * @returns Boolean
    */
   function matchError(err, errLike, errMsgMatcher) {
-    var isMatchingErrLike = true;
-    var isMatchingMessage = true;
+    if (!err) return false;
 
-    if (errLike instanceof Error) {
-      isMatchingErrLike = _.checkError.compatibleInstance(err, errLike);
-    } else if (errLike) {
-      isMatchingErrLike = _.checkError.compatibleConstructor(err, errLike);
+    if (errLike instanceof Error
+        && !_.checkError.compatibleInstance(err, errLike)) {
+      return false;
     }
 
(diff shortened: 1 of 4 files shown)


meeber:
Didn't mean to close this. I think I broke it; the reopen button is grayed out for me. I'll create a new PR later if it doesn't magically fix itself.

Issue #1017, opened 2017-08-04
Wrong signature in assertions documentation


JpOnline:
After the [`.isExtensibleObject`](https://github.com/chaijs/chai/blob/master/lib/chai/interface/assert.js#L2900), all the methods miss the optional message parameter in the documentation.

keithamus:
Thanks for the issue @JpOnline! Want to make a PR to fix this?

Hpauric:
Hey, I can make those fixes, if that suits? Looks like `.ifError(object)` is missing the optional message parameter also.

keithamus:
Sounds great 👌. Look forward to seeing the PR @Hpauric

meeber:
Fixed by #1023

PR #1023, opened 2017-08-09
fix: added missing assert parameters in documentation

Hpauric:
Fix for issue #1017 

Optional message parameter added to documentation for:

- `.ifError`
- `.isExtensible`
- `.isNotExtensible`
- `.isFrozen`
- `.isNotFrozen`
- `.isEmpty`
- `.isNotEmpty`

### DIFF
--- a/chai.js
+++ b/chai.js
@@ -6967,7 +6967,7 @@ module.exports = function (chai, util) {
   }
 
   /*!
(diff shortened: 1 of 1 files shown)


keithamus:
Just noticed this PR edited `chai.js` - so we'll need to rework this to the right places.

PR #1030, opened 2017-08-21
Add `configs` to the use function

Bamieh:
Added `configs` parameter to the `use` function.

Example:
```
import myChaiLibrary from '...';
chai.use(myChaiLibrary, configs)
```
The `used` function now can accept a 3rd parameter, which is the passed settings:

```
function myChaiLibrary(_chai, utils, configs) {

}
```

This adds a nice way to pass configs, rather than writing use then adding `chai.myChaiLibrary.someConfig = true` for example.

Additionally, sometimes, like in my library case, i have some required configs i need during the `use` function call, now i have to create a factory function which accepts the configs, then i have to call the use on the result of the setup. I prefer this way.

Cheers!

### DIFF
--- a/lib/chai.js
+++ b/lib/chai.js
@@ -10,7 +10,7 @@ var used = [];
  * Chai version
  */
 
-exports.version = '4.1.1';
+exports.version = '4.2.1';
 
 /*!
  * Assertion Error
@@ -34,9 +34,9 @@ var util = require('./chai/utils');
  * @api public
  */
 
-exports.use = function (fn) {
+exports.use = function (fn, configs) {
   if (!~used.indexOf(fn)) {
-    fn(exports, util);
+    fn(exports, util, configs);
     used.push(fn);
   }
 


lib/chai.js
```diff
@@ -10,7 +10,7 @@ var used = [];
  * Chai version
  */
 
-exports.version = '4.1.1';
+exports.version = '4.2.1';
```
astorije: I don't think you meant this. Either way, I think it's set at deploy time.

astorije:
Can you make your library so that it's a function that accepts the config as an argument? Something like:

```js
chai.use(myChaiLibrary(configs));
```

Your library would return a function that returns what you currently have in it, applying the passed config (or defaults).

I think this would be nicer than adding a `config` param, and be a more functional-looking. Since Chai does not care/need that config, you only pass it to what needs it (your library).

vieiralucas:
I'm with @astorije here
i'm gonna close this for now.

@Bamieh feel free to continue the discussion here or file an issue if you strongly disagree

Bamieh:
I have no counter argument, its a matter of preference.

PR #1040, opened 2017-09-06
Make tests pass with `--use_strict`

### LINKED ISSUE #1034: `errorMessage` leaked to global scope in non-strict mode, `ReferenceError` in strict mode
If I run `node --use_strict -e 'require("chai").expect("42").to.be.above(0)'` I get:

> ReferenceError: errorMessage is not defined

Instead of:

> AssertionError: expected '42' to be a number or a date

If I run `node -e 'try { require("chai").expect("42").to.be.above(0) } catch (ignore) { require("chai").expect(global.errorMessage).to.be.undefined }'`

I get:

> AssertionError: expected 'expected \'42\' to be a number or a date' to be undefined

* * *

There are probably other cases of variables being used without declaration; this just happens to be the one I ran into. Using `--use_strict` or `--check-leaks` in Mocha testing can help spot these, by the way!

shvaikalesh:
Closes #1034
Related to https://github.com/chaijs/assertion-error/pull/9

### DIFF
--- a/test/should.js
+++ b/test/should.js
@@ -455,20 +455,12 @@ describe('should', function() {
       new Foo().should.be.an.instanceof(undefined);
     }, "The instanceof assertion needs a constructor but undefined was given.");
 
-    // Different browsers may have different error messages
-    var expectedError;
-    try {
-      t instanceof Thing;
-    } catch (err) {
-      errMsg = '[object Object] instanceof function Thing(){} failed: ' + err.message + '.';
-    }
-
     err(function(){
       function Thing(){};
       var t = new Thing();
       Thing.prototype = 1337;
       t.should.be.an.instanceof(Thing);
-    }, expectedError, true);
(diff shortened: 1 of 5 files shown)


test/should.js
```diff
@@ -455,20 +455,12 @@ describe('should', function() {
...
-      errMsg = '[object Object] instanceof function Thing(){} failed: ' + err.message + '.';
-    }
-
     err(function(){
       function Thing(){};
       var t = new Thing();
       Thing.prototype = 1337;
       t.should.be.an.instanceof(Thing);
-    }, expectedError, true);
+    }, 'The instanceof assertion needs a constructor but function was given.', true);
```
shvaikalesh: I think the message is a bit misleading: `TypeError` is thrown because `fn.prototype` is primitive.

astorije:
Should we add `"use strict"` at the top of our files then?

keithamus:
Not yet - we'll do a push for linting very soon though I think.

PR #1042, opened 2017-09-07
Fix typos and remove trailing whitespaces

simonewebdesign:


lucasfcosta:
Hey folks, what about merging this?

LGTM and it's one less PR for us to keep here and cleanup the tracker.

brutalcrozt:
@simonewebdesign May i ask, How you do the removing white space? If use tools, please tell me which one is used.

simonewebdesign:
@brutalcrozt I set my editor of choice (Sublime Text) to automatically trim white space on save, the setting is `"trim_trailing_white_space_on_save": true`.

astorije:
@brutalcrozt, I personally use [Atom](https://atom.io) which trims all trailing whitespaces on save by default, along with extraneous newlines at the end of files. I personally find this very useful to keep things clean :)

PR #1044, opened 2017-09-09
property assertion should only accept strings if nested, fixes #1043

### LINKED ISSUE #1043: Type check for parsePath variable would be nice to have
When calling the parsePath function (https://github.com/chaijs/chai/blob/master/chai.js#L10117) type checking can help users identify incorrect usage of API.

Problematic line:
```
function parsePath(path) {
   var str = path.replace(/([^\\])\[/g, '$1.[');
```

When running this (slightly incorrect) code:
```
expect(data).to.have.nested.property({'a':'1'})
```

The following error is thrown:
```
TypeError: path.replace is not a function
```

Because path in this case is an object `{'a':'1'}`
Would be nice to throw a more descriptive error :)

solodynamo:
### Description of the Change

Adds case check when property assertion is not string and nested flag is set.

### Why should this be in core?

Adds required check case when non-string param is passed with nested flag set.

### Benefits

It will provide more descriptive error to the devs on invalid passed params.

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -1770,11 +1770,19 @@ module.exports = function (chai, _) {
 
     if (isNested) {
       if (nameType !== 'string') {
-        throw new AssertionError(flagMsg + 'the argument to property must be a string when using nested syntax');
+        throw new AssertionError(
+          flagMsg + 'the argument to property must be a string when using nested syntax',
+          undefined,
+          ssfi
+        );
       }
     } else {
       if (nameType !== 'string' && nameType !== 'number' && nameType !== 'symbol') {
-        throw new AssertionError(flagMsg + ' the argument to property must be a string, number, or symbol');
+        throw new AssertionError(
+          flagMsg + 'the argument to property must be a string, number, or symbol',
+          undefined,
+          ssfi
+        );
       }
     }
 
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -1765,14 +1765,13 @@ module.exports = function (chai, _) {
       , obj = flag(this, 'object')
       , ssfi = flag(this, 'ssfi');
 
+    flagMsg = flagMsg ? flagMsg + ': ' : '';
+
     if (typeof name !== 'string' && isNested) {
-      var msgPrefix = flag(this, 'message')
-      msgPrefix = msgPrefix ? msgPrefix + ': ' : ''
-      throw new AssertionError(msgPrefix + 'the argument to `property` must be a string')
+      throw new AssertionError(flagMsg + 'the argument to `property` must be a string')
     }
 
     if (isNested && isOwn) {
-      flagMsg = flagMsg ? flagMsg + ': ' : '';
       throw new AssertionError(
         flagMsg + 'The "nested" and "own" flags cannot be combined.',
         undefined,
@@ -1781,7 +1780,6 @@ module.exports = function (chai, _) {
     }
 
     if (obj === null || obj === undefined) {
-      flagMsg = flagMsg ? flagMsg + ': ' : '';
       throw new AssertionError(
         flagMsg + 'Target cannot be null or undefined.',
         undefined,
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -1770,11 +1770,11 @@ module.exports = function (chai, _) {
 
     if (isNested) {
       if (nameType !== 'string') {
-        throw new AssertionError(flagMsg + 'the argument to `property` must be a string');
+        throw new AssertionError(flagMsg + 'the argument to property must be a string when using nested syntax');
       }
     } else {
       if (nameType !== 'string' && nameType !== 'number' && nameType !== 'symbol') {
-        throw new AssertionError(flagMsg + 'the argument to `property` must be either of type string, number or symbol');
+        throw new AssertionError(flagMsg + ' the argument to property must be a string, number, or symbol');
       }
     }
 
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -1763,12 +1763,19 @@ module.exports = function (chai, _) {
       , isOwn = flag(this, 'own')
       , flagMsg = flag(this, 'message')
       , obj = flag(this, 'object')
-      , ssfi = flag(this, 'ssfi');
+      , ssfi = flag(this, 'ssfi')
+      , nameType = typeof name;
 
     flagMsg = flagMsg ? flagMsg + ': ' : '';
 
-    if (typeof name !== 'string' && isNested) {
-      throw new AssertionError(flagMsg + 'the argument to `property` must be a string')
+    if (isNested) {
+      if (nameType !== 'string') {
+        throw new AssertionError(flagMsg + 'the argument to `property` must be a string');
+      }
+    } else {
+      if (nameType !== 'string' && nameType !== 'number' && nameType !== 'symbol') {
+        throw new AssertionError(flagMsg + 'the argument to `property` must be either of type string, number or symbol');
+      }
     }
 
     if (isNested && isOwn) {
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -1765,6 +1765,12 @@ module.exports = function (chai, _) {
       , obj = flag(this, 'object')
       , ssfi = flag(this, 'ssfi');
 
+    if (typeof name !== 'string' && isNested) {
+      var msgPrefix = flag(this, 'message')
+      msgPrefix = msgPrefix ? msgPrefix + ': ' : ''
+      throw new AssertionError(msgPrefix + 'the argument to `property` must be a string')
+    }
+
     if (isNested && isOwn) {
       flagMsg = flagMsg ? flagMsg + ': ' : '';
       throw new AssertionError(
--- a/test/assert.js
+++ b/test/assert.js
@@ -1460,13 +1460,9 @@ describe('assert', function () {
       assert.propertyVal(dummyObj, 'a', '2', 'blah');
     }, "blah: expected { a: '1' } to have property 'a' of '2', but got '1'");
 
-    assert.doesNotThrow(function () {
-      assert.nestedProperty({a:1}, '{a:1}');
-    }, Error, 'the argument to property must be a string when using nested syntax');
-
-    assert.throws(function () {
-      assert.nestedProperty({a:1}, {'a':'1'});
-    }, Error, 'the argument to property must be a string when using nested syntax');
+    err(function () {
+      assert.nestedProperty({a:1}, {'a':'1'}, 'blah');
+    }, 'blah: the argument to property must be a string when using nested syntax');
   });
 
   it('deepPropertyVal', function () {
--- a/test/assert.js
+++ b/test/assert.js
@@ -1,6 +1,5 @@
 describe('assert', function () {
   var assert = chai.assert;
-  var expect = chai.expect;
 
   it('assert', function () {
     var foo = 'bar';
@@ -1461,13 +1460,13 @@ describe('assert', function () {
       assert.propertyVal(dummyObj, 'a', '2', 'blah');
     }, "blah: expected { a: '1' } to have property 'a' of '2', but got '1'");
 
-    expect(function () {
+    chai.expect(function () {
       assert.nestedProperty({a:1}, '{a:1}');
-    }).to.not.throw('the argument to `property` must be a string');
+    }).to.not.throw('the argument to property must be a string when using nested syntax');
 
-    expect(function () {
+    chai.expect(function () {
       assert.nestedProperty({a:1}, {'a':'1'});
-    }).to.throw('the argument to `property` must be a string');
+    }).to.throw('the argument to property must be a string when using nested syntax');
   });
 
   it('deepPropertyVal', function () {
--- a/test/assert.js
+++ b/test/assert.js
@@ -1460,13 +1460,13 @@ describe('assert', function () {
       assert.propertyVal(dummyObj, 'a', '2', 'blah');
     }, "blah: expected { a: '1' } to have property 'a' of '2', but got '1'");
 
-    chai.expect(function () {
+    assert.doesNotThrow(function () {
       assert.nestedProperty({a:1}, '{a:1}');
-    }).to.not.throw('the argument to property must be a string when using nested syntax');
+    }, Error, 'the argument to property must be a string when using nested syntax');
 
-    chai.expect(function () {
+    assert.throws(function () {
       assert.nestedProperty({a:1}, {'a':'1'});
-    }).to.throw('the argument to property must be a string when using nested syntax');
+    }, Error, 'the argument to property must be a string when using nested syntax');
   });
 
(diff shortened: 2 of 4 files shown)


keithamus:
Great work thanks for this @solodynamo. Do you think you could also add a test case to verify this behaviour? The property tests can be found here: https://github.com/chaijs/chai/blob/dbeae08fe526cbe79634bae07613fc427a8a53aa/test/expect.js#L1416-L1465

meeber:
Thanks for working on this @solodynamo. I think we need to expand this a bit further to also perform a type check if not `isNested`. For example:

```js
var nameType = typeof name;
if (isNested) {
  if (nameType !== 'string') {
    <error>
  }
} else {
  if (nameType !== 'string' && nameType !== 'number' && nameType !== 'symbol') {
    <error>
  }
}
```

As @keithamus mentioned, it would be good to have tests confirming the new behavior. Whatever tests that are added to `chai/test/expect.js` should be replicated for the other interfaces in `chai/test/should.js` and `chai/test/assert.js`.

lib/chai/core/assertions.js
```diff
@@ -1765,6 +1765,12 @@ module.exports = function (chai, _) {
       , obj = flag(this, 'object')
       , ssfi = flag(this, 'ssfi');
 
+    if (typeof name !== 'string' && isNested) {
+      var msgPrefix = flag(this, 'message')
```
meeber: We don't need to define `msgPrefix` here; the value of `flag(this, 'message')` is already stored in the `flagMsg` variable from line 1764.

lib/chai/core/assertions.js
```diff
@@ -1765,6 +1765,12 @@ module.exports = function (chai, _) {
...
+      msgPrefix = msgPrefix ? msgPrefix + ': ' : ''
```
meeber: The logic here on line 1770 is repeated on lines 1769 and 1778 (except with `flagMsg`); I'm thinking we can reduce duplication by removing all three of these and instead performing the operation a single time near the top after line 1766 where the variable declarations take place.

solodynamo: Yes, there is little duplication of logic here . We can remove msgPrefix and use flagMsg variable in its place and move the `flagMsg = flagMsg ? flagMsg + ': ' : ''`  near the top as you suggested.

solodynamo:
Hey @keithamus @meeber updated the PR with suggested changes .

meeber:
@solodynamo Thanks for your continued work on this! Please see [this comment](https://github.com/chaijs/chai/pull/1044#issuecomment-328307113) for a couple more suggested changes.

solodynamo:
Hey @meeber  incorporated more suggested changes .

lib/chai/core/assertions.js
```diff
@@ -1763,10 +1763,22 @@ module.exports = function (chai, _) {
...
       , obj = flag(this, 'object')
-      , ssfi = flag(this, 'ssfi');
+      , ssfi = flag(this, 'ssfi')
+      , nameType = typeof name;
+
+    flagMsg = flagMsg ? flagMsg + ': ' : '';
+
+    if (isNested) {
+      if (nameType !== 'string') {
+        throw new AssertionError(flagMsg + 'the argument to `property` must be a string');
```
vieiralucas: What about telling that it must be string because the nested flag is on?

Something like:
```js
throw new AssertionError(flagMsg + 'the argument to `property` with nested flag must be a string');
```

I'm afraid that it will be confusing to sometimes say that it can be a `string`, `number` or `symbol` and sometimes just saying that it must be a `string`

Thoughts? @keithamus @solodynamo @meeber

solodynamo: But rather than writing `nested flag` we should think something else as end user may not know about internal logic of `isNested`.  Suggestions ? @vieiralucas @meeber @keithamus

keithamus: Agreed, we can safely assume to get here they've written `.nested.propery()` so we can use that instead.

meeber: A user could reach this error via the `assert` interface too, though, so `nested.property` wouldn't always be entirely accurate.

test/expect.js
```diff
@@ -1462,6 +1462,18 @@ describe('expect', function () {
...
+      expect({a:1}).to.have.nested.property('{a:1}');
+    }).to.not.throw('the argument to `property` must be a string');
+
+    expect(function () {
+      expect({a:1}).to.have.nested.property({'a':'1'});
+    }).to.throw('the argument to `property` must be a string');
+
+    expect(function () {
+      expect({a:1}).to.have.property(null);
+    }).to.throw('the argument to `property` must be either of type string, number or symbol');
```
meeber: Let's repeat these three new error tests in the `should` and `assert` interface test files.

solodynamo: Okay , so just duplicating them in both the files.

solodynamo: @keithamus @meeber `isNested` is true when `name` is `undefined` , Shouldn't this be false

lib/chai/core/assertions.js
vieiralucas: I'm having a hard time thinking on a good message

```
the argument to `property` when nesting must be a `string`

the argument to `property` for a nested assertion must be a `string`
```

test/expect.js
meeber: @solodynamo I'm not sure what you mean.

On a separate note, I just noticed a couple of things:

- Let's remove the `.not.throw` test on lines 1466-1468; we already test successful `.nested.property` assertions in the section that starts on line 1754.
- Let's relocate the test on lines 1470-1472 to the end of the `nested.property` section (that starts on line 1754).

solodynamo: @meeber I meant to ask 
` expect(function () {
      ({a:1}).should.have.nested.property(undefined);
    }).to.throw('the argument to `property` must be either of type string, number or symbol');` 
Above test is throwing error `'the argument to `property` must be a string'`. I debugged and found `isNested` is true in above case , so i meant to ask shouldn't `isNested` be false. Please correct me if i misunderstood something.

solodynamo:
@meeber @keithamus @vieiralucas look through updated changes, if looks fine please merge.

test/assert.js
```diff
@@ -1,5 +1,6 @@
 describe('assert', function () {
   var assert = chai.assert;
+  var expect = chai.expect;
```
meeber: Let's not add `expect` to the `assert` interface tests. It should be possible to test errors without `expect` by using the same style that's used elsewhere in this file.

test/should.js
```diff
@@ -1,5 +1,6 @@
 describe('should', function() {
   var should = chai.Should();
+  var expect = chai.expect;
```
meeber: Same thing here. Let's not add `expect` to the `should` interface tests.

lib/chai/core/assertions.js
meeber: How about:

**Normal:** the argument to `property` must be a `string`, `number`, or `symbol`
**Nested:** the argument to `property` must be a `string` when using `nested` syntax

solodynamo:
@meeber updated PR .

test/assert.js
```diff
@@ -1461,13 +1460,13 @@ describe('assert', function () {
...
       assert.nestedProperty({a:1}, '{a:1}');
-    }).to.not.throw('the argument to `property` must be a string');
+    }).to.not.throw('the argument to property must be a string when using nested syntax');
 
-    expect(function () {
+    chai.expect(function () {
       assert.nestedProperty({a:1}, {'a':'1'});
-    }).to.throw('the argument to `property` must be a string');
+    }).to.throw('the argument to property must be a string when using nested syntax');
   });
```
solodynamo: we are using `chai.expect` on line 18 too.

vieiralucas: It's shouldn't be wrong to use `expect` in this test, because here we are actually using `chai` to assert, we are not calling `assertions` and expecting them to work, we are testing that the assertion `throws`.

But I agree with @meeber that we must try and be able to use own `interface` to test it.

You can replace the ones you added on this PR with `assert.throws` and maybe remove the others in another PR.

solodynamo: Sure @vieiralucas . I have replaced them and also surely will open a seperate PR in future to replace the remaining.

solodynamo:
@meeber updated PR .

test/assert.js
```diff
@@ -1454,6 +1455,18 @@ describe('assert', function () {
...
+      assert.propertyVal(dummyObj, 'a', '2', 'blah');
+    }, "blah: expected { a: '1' } to have property 'a' of '2', but got '1'");
+
+    assert.doesNotThrow(function () {
+      assert.nestedProperty({a:1}, '{a:1}');
+    }, Error, 'the argument to property must be a string when using nested syntax');
+
+    assert.throws(function () {
+      assert.nestedProperty({a:1}, {'a':'1'});
+    }, Error, 'the argument to property must be a string when using nested syntax');
```
vieiralucas: @solodynamo I'm sorry for making a mess, but I realized that what @meeber meant to suggest is for you to use the `err` function just like on the tests above.

Notice that when you do that it will start to throw `implementation frames not properly filtered from stack trace:`.
Thats because you forgot to pass the `ssfi` argument to the constructor of the `AssertionError`
You are doing this:
```js
throw new AssertionError(flagMsg + 'the argument to property must be a string when using nested syntax');
```
But you need to do this:
```js
throw new AssertionError(
  flagMsg + 'the argument to property must be a string when using nested syntax',
  undefined,
  ssfi
);
```

I'm sorry again for the confusion I made.

lib/chai/core/assertions.js
```diff
@@ -1763,10 +1763,22 @@ module.exports = function (chai, _) {
...
+
+    flagMsg = flagMsg ? flagMsg + ': ' : '';
+
+    if (isNested) {
+      if (nameType !== 'string') {
+        throw new AssertionError(flagMsg + 'the argument to property must be a string when using nested syntax');
+      }
+    } else {
+      if (nameType !== 'string' && nameType !== 'number' && nameType !== 'symbol') {
+        throw new AssertionError(flagMsg + ' the argument to property must be a string, number, or symbol');
```
vieiralucas: It looks like this has an extra `space`. `flagMsg` variable already has a trailing space

test/assert.js
```diff
@@ -1454,6 +1455,18 @@ describe('assert', function () {
...
       assert.property(undefined, 'a', 'blah');
     }, "blah: Target cannot be null or undefined.");
+
+    err(function () {
+      assert.propertyVal(dummyObj, 'a', '2', 'blah');
+    }, "blah: expected { a: '1' } to have property 'a' of '2', but got '1'");
+
+    assert.doesNotThrow(function () {
+      assert.nestedProperty({a:1}, '{a:1}');
+    }, Error, 'the argument to property must be a string when using nested syntax');
```
vieiralucas: You don't think you need to add this test actually.
There is already a test above on line [1404](https://github.com/chaijs/chai/pull/1044/files#diff-9c6c3cd6ffcef4f57ac6e0c793b31b85R1404) for the `assert.nestedProperty` passing a String.

test/assert.js
vieiralucas: Also, please pass a third argument to test if the custom message works.
Your test should look like this:
```js
err(function () {
  assert.nestedProperty({a:1}, {'a':'1'}, 'blah');
}, 'blah: the argument to property must be a string when using nested syntax');
```

test/expect.js
```diff
@@ -1462,6 +1462,10 @@ describe('expect', function () {
     err(function () {
       expect(undefined, 'blah').to.have.property("a");
     }, "blah: Target cannot be null or undefined.");
+
+    expect(function () {
+      expect({a:1}).to.have.property(null);
+    }).to.throw('the argument to property must be a string, number, or symbol');
```
meeber: This should use the `err` helper function, just like the test above it.

test/expect.js
```diff
@@ -1755,6 +1759,10 @@ describe('expect', function () {
       expect({ 'foo.bar': 'baz' })
         .to.have.nested.property('foo.bar');
     }, "expected { 'foo.bar': 'baz' } to have nested property 'foo.bar'");
+
+    expect(function () {
+      expect({a:1}).to.have.nested.property({'a':'1'});
+    }).to.throw('the argument to property must be a string when using nested syntax');
```
meeber: This should use the `err` helper function.

test/should.js
```diff
@@ -1407,6 +1407,16 @@ describe('should', function() {
...
 
+    ({a:1}).should.have.nested.property('a');
+
+    (function () {
+      ({a:1}).should.have.nested.property('{a:1}');
+    }).should.not.throw('the argument to property must be a string when using nested syntax');
+
+    (function () {
+      ({a:1}).should.have.nested.property({'a':'1'});
+    }).should.throw('the argument to property must be a string when using nested syntax');
```
meeber: This should use the `err` helper function, just like the test below it.

test/assert.js
```diff
@@ -1454,6 +1455,14 @@ describe('assert', function () {
...
       assert.property(undefined, 'a', 'blah');
     }, "blah: Target cannot be null or undefined.");
+
+    err(function () {
+      assert.propertyVal(dummyObj, 'a', '2', 'blah');
+    }, "blah: expected { a: '1' } to have property 'a' of '2', but got '1'");
+
+    err(function () {
+      assert.nestedProperty({a:1}, {'a':'1'}, 'blah');
+    }, 'blah: the argument to property must be a string when using nested syntax');
```
meeber: One more test should be added here to the `assert` interface tests in order to verify the error message "the argument to property must be a string, number, or symbol". This test can closely resemble the `expect({a:1}).to.have.property(null);` test that you added on the `expect` interface, except using the `assert` interface instead of `expect`. (And using the `err` helper function, just like you've done in the previous two tests on the `assert` interface).

test/should.js
```diff
@@ -1407,6 +1407,16 @@ describe('should', function() {
 
     ({ 'foo.bar[]': 'baz'}).should.have.nested.property('foo\\.bar\\[\\]');
 
+    ({a:1}).should.have.nested.property('a');
+
+    (function () {
+      ({a:1}).should.have.nested.property('{a:1}');
+    }).should.not.throw('the argument to property must be a string when using nested syntax');
```
meeber: Let's remove this test. We already test that `.nested.property` works properly elsewhere in the suite. Adding this test with ".should.not.throw` doesn't add any value.

test/should.js
```diff
@@ -1407,6 +1407,16 @@ describe('should', function() {
...
+
+    (function () {
+      ({a:1}).should.have.nested.property({'a':'1'});
+    }).should.throw('the argument to property must be a string when using nested syntax');
+
```
meeber: One more test should be added here to the `should` interface tests in order to verify the error message "the argument to property must be a string, number, or symbol". This test can closely resemble the `expect({a:1}).to.have.property(null);` test that you added on the `expect` interface, except using the `should` interface this time instead of `expect`. (And using the `err` helper function).

solodynamo:
@keithamus @meeber @vieiralucas

meeber:
LGTM! @vieiralucas ?

vieiralucas:
LGTM! Thank you so much for this PR @solodynamo 
I'm sorry that it took so long, but you did it! I hope you had a good experience contributing to chai. :smile:

PR #1049, opened 2017-09-12
Remove unnecessary code

abetomo:
* Remove unnecessary line feeds
* Remove unused variable declaration
* Remove meaningless processing (lib/chai/utils/inspect.js)
    * `numLinesEst` does not seem to be used anywhere after this.

It is not a remove, but changed to a single quote unified to other coding styles. (`'number'` of lib/chai/core/assertions.js)


abetomo:
@astorije Thanks for the review.
I squashed the commitment.

PR #1052, opened 2017-09-14
Remove unnecessary `else` and clean up the code

abetomo:
Sorry frequently with a little modify.
I also fixed it because I noticed some.

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -2612,7 +2612,7 @@ module.exports = function (chai, _) {
       var isCompatibleConstructor = _.checkError.compatibleConstructor(caughtErr, errorLike);
       if (isCompatibleConstructor === negate) {
         if (everyArgIsDefined && negate) {
-            errorLikeFail = true;
+          errorLikeFail = true;
         } else {
           this.assert(
               negate
@@ -2635,7 +2635,7 @@ module.exports = function (chai, _) {
       var isCompatibleMessage = _.checkError.compatibleMessage(caughtErr, errMsgMatcher);
       if (isCompatibleMessage === negate) {
         if (everyArgIsDefined && negate) {
-            errMsgMatcherFail = true;
+          errMsgMatcherFail = true;
         } else {
           this.assert(
             negate
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -1147,7 +1147,7 @@ module.exports = function (chai, _) {
     if (doLength) {
       new Assertion(obj, flagMsg, ssfi, true).to.have.property('length');
     }
-    
+
     if (!doLength && (objType === 'date' && nType !== 'date')) {
       errorMessage = msgPrefix + 'the argument to above must be a date';
     } else if (nType !== 'number' && (doLength || objType === 'number')) {
@@ -1436,7 +1436,7 @@ module.exports = function (chai, _) {
     if (doLength) {
       new Assertion(obj, flagMsg, ssfi, true).to.have.property('length');
     }
-    
+
     if (!doLength && (objType === 'date' && nType !== 'date')) {
       errorMessage = msgPrefix + 'the argument to most must be a date';
     } else if (nType !== 'number' && (doLength || objType === 'number')) {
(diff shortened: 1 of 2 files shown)


astorije:
> Remove unnecessary `else`
> `else` is unnecessary as it has been returned by the previous if statement.

I know this is a thing now, we're actually using this in another project I maintain and I'm definitely not a fan of this syntax. Yes it saves 1 line (`} else {`) but in practice I find the reading flow affected by it. Several times it has caught me off-guard, because by overlooking the `return` statement above, I missed that what I'm reading is an `else` block in disguise.

But that is opinion really, I don't have data or facts to back this up, just personal experience, so I'll be happy to go with the flow if other @chaijs/chai strongly disagree :)
Either way, all other commits are 👌, thanks!

abetomo:
Thank you for review!
I see. I will fix it according to chai's policy!
I am waiting for the review.

astorije:
> I am waiting for the review.

Yes, please wait before wasting your time updating your PR, because maybe I'm the only one thinking that, or maybe others will have better reasons. Either way, I'll go with the flow :)

lib/chai/core/assertions.js
```diff
@@ -1017,16 +1017,15 @@ module.exports = function (chai, _) {
...
-      );
     }
+    this.assert(
+        val === obj
+      , 'expected #{this} to equal #{exp}'
+      , 'expected #{this} to not equal #{exp}'
+      , val
+      , this._obj
+      , true
+    );
```
vieiralucas: In this case it makes more sense to leave the `else` and remove the `return` inside the `if`
Or leave the `return` alone in a new line after calling `this.eql` and keep doing this early returning (I'm personally a big fan of early returns btw)

Thats because this function should return `undefined` and its currently sometimes returning the call of `this.eql` and sometimes just `undefined`.

I know `this.eql` will eventually return `undefined` but I need to read `this.eql` implementation to know that.

If I would refactor this code my code would look like this
Using early return:
```js
function assertEqual (val, msg) {
  if (flag(this, 'deep')) {
    this.eql(val);
    return;
  }

  this.assert(
      val === obj
    , 'expected #{this} to equal #{exp}'
    , 'expected #{this} to not equal #{exp}'
    , val
    , this._obj
    , true
  );
}
```
Or just removing the `return`:
```js
function assertEqual (val, msg) {
  if (flag(this, 'deep')) {
    this.eql(val);
  } else {
    this.assert(
        val === obj
      , 'expected #{this} to equal #{exp}'
      , 'expected #{this} to not equal #{exp}'
      , val
      , this._obj
      , true
    );
  }
}
```

lib/chai/core/assertions.js
```diff
@@ -1017,16 +1017,15 @@ module.exports = function (chai, _) {
     var obj = flag(this, 'object');
     if (flag(this, 'deep')) {
       return this.eql(val);
-    } else {
```
vieiralucas: In this case it makes more sense to leave the `else` and remove the `return` inside the `if`
Or leave the `return` alone in a new line after calling `this.eql` and keep doing this early returning (I'm personally a big fan of early returns btw)

Thats because this function should return `undefined` and its currently sometimes returning the call of `this.eql` and sometimes just `undefined`.

I know `this.eql` will eventually return `undefined` but I need to read `this.eql` implementation to know that.

If I would refactor this code my code would look like this
Using early return:
```js
function assertEqual (val, msg) {
  if (flag(this, 'deep')) {
    this.eql(val);
    return;
  }

  this.assert(
      val === obj
    , 'expected #{this} to equal #{exp}'
    , 'expected #{this} to not equal #{exp}'
    , val
    , this._obj
    , true
  );
}
```
Or just removing the `return`:
```js
function assertEqual (val, msg) {
  if (flag(this, 'deep')) {
    this.eql(val);
  } else {
    this.assert(
        val === obj
      , 'expected #{this} to equal #{exp}'
      , 'expected #{this} to not equal #{exp}'
      , val
      , this._obj
      , true
    );
  }
}
```

vieiralucas:
So, besides the one that I commented on these changes looks fine to me.
But I think we should start linting chai's code to fix this for real

lib/chai/core/assertions.js
abetomo: Thanks for the review.
I am also a fan of early return, so I fixed it like that.

abetomo:
Thank you. I fixed it.
I think that there are things that can be helpful in code linting, so I'm happy if I can contribute in the future.

abetomo:
Thank you for review.
Since time has passed, I will close it.

PR #1061, opened 2017-10-10
fix: simplify chai module export

keithamus:
There's an issue around chai using the `exports` keyword in a dynamic way that means the main module cannot be used in a way with module systems that statically infer what the exports are. 

This change simply refactors the export code to ensure `chai` is a single module that gets exported once, and doesn't refer to `module.exports` or `exports` inside of any function calls.

See https://github.com/josh/selector-observer/pull/4

### DIFF
--- a/lib/chai.js
+++ b/lib/chai.js
@@ -5,18 +5,19 @@
  */
 
 var used = [];
+var chai = {};
 
 /*!
  * Chai version
  */
 
-exports.version = '4.1.2';
+chai.version = '4.1.2';
 
 /*!
  * Assertion Error
  */
 
-exports.AssertionError = require('assertion-error');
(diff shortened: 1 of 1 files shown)


astorije:
@keithamus, tests seem to be completely lost with your changes though 😅

meeber:
The problem looks to be related to circular dependencies. The problem doesn't manifest until exporting a new object in `lib/chai.js` instead of the default `module.exports`.

In `lib/chai.js`:

```js
var util = require('./chai/utils');
chai.util = util;
```

For example, in `lib/chai/utils/addProperty.js` (which is being required in `lib/chai/utils/index.js`):

```js
var chai = require('../../chai');
console.log(chai);  // Prints: {}
```

Issue #1070, opened 2017-10-27
Expect to throw (number) doesn't seem to work


SamStonehouse:
const failingCall = function() {
      throw 10;
    };

    expect(failingCall).to.throw(10);

Results in 'AssertionError: expected [Function: failingCall] to throw 10 but 10 was thrown'

    const failingCall = function() {
      throw 'test';
    };

    expect(failingCall).to.throw('test');

Passes as normal

And...

    const failingCall = function() {
      throw 0;
    };

    expect(failingCall).to.throw(0);

Fails with 'AssertionError: expected [Function: failingCall] to throw an error'

keithamus:
Hey @SamStonehouse thanks for the issue.

This seems to be a problem with our type checking for that assertion. We should throw a better error message saying that `throw(instanceof Number)` is not supported. 

PRs welcome 😄

bryanbraun:
I put together a PR for this issue over at https://github.com/chaijs/chai/pull/1071. Let me know if you have any feedback.

keithamus:
Hey @SamStonehouse thanks for the issue.

We've added this to our Roadmap https://github.com/chaijs/chai/projects/2! We'll be releasing chai 5 soon, but for now I'll close this issue because it is tracked on our roadmap.

PR #1071, opened 2017-10-29
Improve type checking / error when numbers are passed to .throw()

bryanbraun:
Addresses chaijs/chai#1070, and includes a test. Let me know if you have any questions. 😄

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -2566,6 +2566,14 @@ module.exports = function (chai, _) {
       , negate = flag(this, 'negate') || false;
     new Assertion(obj, flagMsg, ssfi, true).is.a('function');
 
+    if (typeof errorLike === 'number') {
+      throw new AssertionError(
+        'the argument to `.throw()` cannot be an number',
+        undefined,
+        ssfi
+      );
+    }
+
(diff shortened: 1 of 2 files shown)


bryanbraun:
@meeber, thanks for the feedback. I'll add tests for those other interfaces, and work on handling those multi-parameter cases as well. 👍

bryanbraun:
(I was going to update this but I saw the other work happening in https://github.com/chaijs/chai/pull/1021, so I'm holding off for now and keeping an eye on that one. If that doesn't end up working out, I'm happy to jump back in here and make the changes that were recommended)

keithamus:
Hey @bryanbraun! Thanks for making this PR. I'm sitting next to @lucasfcosta right now and we've been discussing where to go with this PR. While we're keen to get this merged, so we can register your contribution to chai - the reality is we're not going to be making a release any time soon, until we've cleaned up a lot of stuff. So merging this would just give a false sense that we're going to release it - but that won't be true. This feature will eventually land in chai 5, but chai 5 will be a while off.

Please don't take that as a sign that your contribution wasn't worthwhile! We hugely appreciate the time you've put into this and please, don't let this dissuade you from making more PRs! I really hope we see more contributions from you 😃

PR #1078, opened 2017-11-02
Feat(build seq) codecov, use bash instead of npm module.

### LINKED ISSUE #1077: Improvement build sequence (codecov)
Well i am too excited to contribute to chai, so previous work #1073 seems have better way.
It's just 300kb , but in my point of view it save developers time to install it and save small disk space :smile: .
image
The better way is via [bash](https://docs.codecov.io/docs/about-the-codecov-bash-uploader#section-via-flag).

brutalcrozt:
Fix #1077

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,8 +11,8 @@ node_js:
 
 script:
   - make test-travisci
-  - codecov
-
+after_success:
+  - bash <(curl -s https://codecov.io/bash)
 env:
   global:
   - secure: h4eLQ2jkR1p2O5BEyscLtbNT/84LDGFnj977SeKmg3flVTpRR+ZiYb5YvIUIYfnwsw1jlr11xv0cya23nJjJA5DQjrVS1/kg4sWrLD9Gs6TTqpSNSk0HrFpxhrHZ/uSSPfV9fbyxfDLxgq72fc55WCuY5Osy1kXmsPdJX79/cOU=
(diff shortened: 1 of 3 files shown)


keithamus:
Hey @brutalcrozt thanks for the issue.

I feel _really_ uncomfortable downloading and executing aribtrary bash on my local machine, and on CI (where we have secret tokens like account passwords). The 300k is only downloaded for developers/ci (where it is also heavily cached), so I think it's adequete to keep it. I'll close this issue, you did the right thing in the original PR, and changing to running arbitrary bash would be a bad move IMO.

astorije:
Something I often see is adding the dependency in the build config itself and not `package.json`:

```
install:
  - npm install -g codecov
script:
  - ...
  - codecov
```

That would solve what you were trying to solve, @brutalcrozt, without relying on a bash download.

---

In fact, considering `codecov` is currently _not_ installed `-g`lobally, how come running `codecov` works? I'd expect to see `./node_modules/.bin/codecov` in there instead, unless Travis runs their `script` section from `./node_modules/.bin` itself.

keithamus:
I believe Travis have `./node_modules/.bin` in the `PATH`. We could make travis install codecov independantly, but I would like to draw complexity _out_ of our travis scripts and into tooling that can work agnostic of CI - especially as I'd like to add appveyor into our toolchain for testing IE without saucelabs, so package.json feels right to me. Other should comment though, we should reach a consensus.

meeber:
I like the self-documenting nature of keeping the dev dependency in `package.json`. I'd like to keep it as-is.

Issue #1084, opened 2017-11-09
Increase assert interface coverage


brutalcrozt:
List of targets:
- [changes with getter function without message](https://codecov.io/gh/chaijs/chai/src/6e9bfc1329f79fa25c7c60fc801a405e5c2ca077/lib/chai/interface/assert.js#L2497...2498) ok
- [changesBy with getter function without message](https://codecov.io/gh/chaijs/chai/src/6e9bfc1329f79fa25c7c60fc801a405e5c2ca077/lib/chai/interface/assert.js#L2525...2527) 
fail with expecting getter function to have `.value`
- [doesNotChange with getter function without message](https://codecov.io/gh/chaijs/chai/src/6e9bfc1329f79fa25c7c60fc801a405e5c2ca077/lib/chai/interface/assert.js#L2557...2558) ok
- [changesButNotBy with getter function without message](https://codecov.io/gh/chaijs/chai/src/6e9bfc1329f79fa25c7c60fc801a405e5c2ca077/lib/chai/interface/assert.js#L2586...2588)
fail with expecting getter function to have `.value`
- increases and decreases with getter function without message notOk
  > decreasesBy fail with expecting getter function to have `.value`

  > decreasesButNotBy fail with expecting getter function to have `.value`

  > so do with the increasesBy and increasesButNotBy

I seems that i need some help with explaining or pointing why it expecting getter function to have specific property.

keithamus:
Hey @brutalcrozt thanks for this issue.

We're going to work a bunch on chai soon. We'll be looking at increasing test coverage of all of our assertions. I'll close this for now because there's nothing specifically actionable from this issue!

PR #1085, opened 2017-11-09
test(assert): increase coverage

brutalcrozt:
related issue #1084

### DIFF
--- a/test/assert.js
+++ b/test/assert.js
@@ -2086,7 +2086,7 @@ describe('assert', function () {
     err(function() {
       assert.isAbove(now, now, 'blah');
     }, 'blah: expected ' + now.toUTCString() + ' to be above ' + now.toUTCString());
-    
+
     err(function() {
       assert.isAbove(null, now);
     }, 'expected null to be a number or a date');
@@ -2251,18 +2251,23 @@ describe('assert', function () {
         heroes = ['spiderman', 'superman'],
         fn     = function() { obj.value += 5 },
         fnDec  = function() { obj.value -= 20 },
+        getterFn = function() { return obj.value },
         bangFn = function() { obj.str += '!' },
         smFn   = function() { 'foo' + 'bar' },
         batFn  = function() { heroes.push('batman') },
         lenFn  = function() { return heroes.length };
 
     assert.changes(fn, obj, 'value');
+    assert.changes(fn, getterFn, 'changes via getter function');
     assert.changesBy(fn, obj, 'value', 5);
     assert.changesBy(fn, obj, 'value', -5);
+    assert.changesBy(fn, getterFn, 5);
     assert.changesBy(fnDec, obj, 'value', 20);
 
     assert.doesNotChange(smFn, obj, 'value');
+    assert.doesNotChange(smFn, getterFn, 'value');
     assert.changesButNotBy(fnDec, obj, 'value', 1);
+    assert.changesButNotBy(fnDec, getterFn, 1);
 
     assert.changes(bangFn, obj, 'str');
 
@@ -2270,7 +2275,7 @@ describe('assert', function () {
     assert.changesButNotBy(batFn, lenFn, 2);
 
     err(function () {
-      assert.changes(smFn,obj, 'value', 'blah');
+      assert.changes(smFn, obj, 'value', 'blah');
     }, "blah: expected .value to change");
 
     err(function () {
@@ -2302,17 +2307,26 @@ describe('assert', function () {
         lenFn = function() { return arr.length },
         incFn = function() { obj.value += 2 },
         decFn = function() { obj.value -= 3 },
+        getterFn = function() { return obj.value },
         smFn  = function() { obj.value += 0 };
 
     assert.decreases(decFn, obj, 'value');
+    assert.decreases(decFn, getterFn, 'decreases via getter function');
     assert.doesNotDecrease(smFn, obj, 'value');
+    assert.doesNotDecrease(smFn, getterFn, 'value');
     assert.decreasesBy(decFn, obj, 'value', 3);
+    assert.decreasesBy(decFn, getterFn, 3);
     assert.decreasesButNotBy(decFn, obj, 'value', 10);
+    assert.decreasesButNotBy(decFn, getterFn, 10);
 
     assert.increases(incFn, obj, 'value');
+    assert.increases(incFn, getterFn, 'increases via getter function');
     assert.doesNotIncrease(smFn, obj, 'value');
+    assert.doesNotIncrease(smFn, getterFn, 'value');
     assert.increasesBy(incFn, obj, 'value', 2);
+    assert.increasesBy(incFn, getterFn, 2);
     assert.increasesButNotBy(incFn, obj, 'value', 1);
+    assert.increasesButNotBy(incFn, getterFn, 1);
 
     assert.decreases(popFn, lenFn);
     assert.doesNotDecrease(pFn, lenFn);


brutalcrozt:
Since the `msg` is empty on [changesBy](https://github.com/chaijs/chai/blob/master/lib/chai/interface/assert.js#L2525), [changesButNotBy](https://github.com/chaijs/chai/blob/master/lib/chai/interface/assert.js#L2586), [increasesBy](https://github.com/chaijs/chai/blob/master/lib/chai/interface/assert.js#L2647), [decreasesBy](https://github.com/chaijs/chai/blob/master/lib/chai/interface/assert.js#L2769), [doesNotDecreaseBy](https://github.com/chaijs/chai/blob/master/lib/chai/interface/assert.js#L2830). I want to refactor them to not use `tmpMsg`. Please let me know if it's a wrong turn. I will rely on travis test only when refactor them.

Seems [doesNotIncrease](https://github.com/chaijs/chai/blob/master/lib/chai/interface/assert.js#L2679) do it correctly. Since the author of #621 is @lucasfcosta, i like to listen to his opinion as well.

lib/chai/interface/assert.js
```diff
@@ -1012,49 +1012,49 @@ module.exports = function (chai, util) {
 
   /**
    * ### .nestedInclude(haystack, needle, [message])
-   * 
-   * Asserts that 'haystack' includes 'needle'. 
-   * Can be used to assert the inclusion of a subset of properties in an 
+   *
+   * Asserts that 'haystack' includes 'needle'.
+   * Can be used to assert the inclusion of a subset of properties in an
```
brutalcrozt: trailing space are removed by atom i guest. Would this affect documentation? If yes i will carefully add space on `every` next line. So next line will be `*<space><space>...`

test/assert.js
```diff
@@ -2251,26 +2251,31 @@ describe('assert', function () {
...
+    assert.changes(fn, getterFn, 'value');
     assert.changesBy(fn, obj, 'value', 5);
     assert.changesBy(fn, obj, 'value', -5);
+    // assert.changesBy(fn, getterFn, 'value', 5);
     assert.changesBy(fnDec, obj, 'value', 20);
 
     assert.doesNotChange(smFn, obj, 'value');
+    assert.doesNotChange(smFn, getterFn, 'value');
     assert.changesButNotBy(fnDec, obj, 'value', 1);
+    // assert.changesButNotBy(fnDec, getterFn, 'value', 1);
```
keithamus: Why are these assertions commented out? Do they not work?

lib/chai/interface/assert.js
```diff
@@ -2767,8 +2766,8 @@ module.exports = function (chai, util) {
   assert.decreasesBy = function (fn, obj, prop, delta, msg) {
     if (arguments.length === 4 && typeof obj === 'function') {
       var tmpMsg = delta;
-      delta = prop;
       msg = tmpMsg;
+      prop = null;
```
lucasfcosta: I think this is not correct.

When we have four arguments and the type of `obj` is a function, we cannot set `prop` to null because the user might desire to check for the value of a `prop` in a `function`, since a `function` in JavaScript can have properties.

Please correct me if I'm wrong.

lib/chai/interface/assert.js
```diff
@@ -2860,9 +2859,8 @@ module.exports = function (chai, util) {
 
   assert.decreasesButNotBy = function (fn, obj, prop, delta, msg) {
     if (arguments.length === 4 && typeof obj === 'function') {
-      var tmpMsg = delta;
-      delta = prop;
-      msg = tmpMsg;
+      msg = 'getter function';
+      prop= null;
```
lucasfcosta: Same here, if we have four arguments to this I think that `prop` must not be `null`.

Also, a space is missing between `prop` and the `=` sign.

lib/chai/interface/assert.js
lucasfcosta: Well, I would not worry about this that much even though we've already got #1042.

test/assert.js
```diff
@@ -2251,26 +2251,31 @@ describe('assert', function () {
...
         fn     = function() { obj.value += 5 },
         fnDec  = function() { obj.value -= 20 },
+        getterFn = function() { return obj.value },
         bangFn = function() { obj.str += '!' },
         smFn   = function() { 'foo' + 'bar' },
         batFn  = function() { heroes.push('batman') },
         lenFn  = function() { return heroes.length };
 
     assert.changes(fn, obj, 'value');
+    assert.changes(fn, getterFn, 'value');
```
lucasfcosta: I also think that this is incorrect.

When writing `assert.changes(fn, getterFn, 'value')` I understand that you expect `fn` to change the property `value` of `getterFn`.

It's redundant to specify a `key` if you already have a getter function. If we did this it would not be possible to check for property in `function` objects as well.

lib/chai/interface/assert.js
brutalcrozt: Alright, the overall explanation make it clearer. I will process when i had chance :smile: , glad to see you back. :love_letter:

test/assert.js
brutalcrozt: Yes Cap, it doesn't work. That's why i made #1090 , :smile:

lib/chai/interface/assert.js
brutalcrozt: OMG, he notice :smile: , indeed that's my mistake. Did we need [eslint](https://eslint.org/)?

test/assert.js
brutalcrozt: I think the misunderstanding is located [here](https://github.com/chaijs/chai/issues/1090#issuecomment-345946379).
> I found it awkward when we use **assert(fn, getterFn, null, 1);**

Which should be `assert.changesBy(fn, getterFn, deltaValue);`

I will continue discussing on issue 1090, kindly re-validate with your point of view.

lib/chai/interface/assert.js
brutalcrozt: I will try to exclude this commit, focusing on testing only. Either we both wrong or none of us is wrong.

test/assert.js
```diff
@@ -2251,26 +2251,33 @@ describe('assert', function () {
         heroes = ['spiderman', 'superman'],
         fn     = function() { obj.value += 5 },
         fnDec  = function() { obj.value -= 20 },
+        getterFn = function() { return obj.value },
+        getter= function() {},
```
lucasfcosta: Missing a space between `=` and the identifier.

test/assert.js
```diff
@@ -2251,26 +2251,33 @@ describe('assert', function () {
...
         bangFn = function() { obj.str += '!' },
         smFn   = function() { 'foo' + 'bar' },
         batFn  = function() { heroes.push('batman') },
         lenFn  = function() { return heroes.length };
 
+    getter.value= obj.value;
```
lucasfcosta: Missing a space here too. But this line will probably end up being deleted due to never using `getter`.

test/assert.js
lucasfcosta: It also seems like this function is never used.

test/assert.js
```diff
@@ -2251,26 +2251,33 @@ describe('assert', function () {
...
     assert.changes(fn, obj, 'value');
+    assert.changes(fn, getterFn, 'value');
```
lucasfcosta: This does not seem correct.

If we want to test properties on functions we must create an `fn` that changes the value of a prop in the `getterFn`. What you've done here is create another `getter` function which has a `value` prop that is never used.

test/assert.js
```diff
@@ -2251,26 +2251,33 @@ describe('assert', function () {
...
     assert.changesBy(fn, obj, 'value', 5);
     assert.changesBy(fn, obj, 'value', -5);
+    assert.changesBy(fn, getterFn, 5);
```
lucasfcosta: Please take a look at the comment I've made on the first assertion test because that is valid to all additions below, such as this one.

brutalcrozt: > so basically resets all the feature changes right?

Yes, sir.

> Let's make sure all the code is correctly indented too.

Agreed.

---
Sorry that i have made unnecessary notification and work for you. I already take a look at your slack status about 5-10 min, because i already had bad experience with `git rebase`. So i ask Keith instead, but still i [get lost](https://slack-files.com/T0901SDRC-F84725QE4-b55bdbbf46) on my own confusion. With your aid i found some [clarity](https://runkit.com/adi/5a16d0989cd06300128397b2), but it seems i found it while you do the last review.

Due to unexpected rain, i made those dirty line. I have to save really quick and forgot to clean up, i will make sure it won't happen again. I will watch video that you mention, when i get lost again about `git rebase`.1. I am also not sure are you still busy on real world or not 2. Based on previous response time, i assume that this PR will get a bit of delay. Somehow i miss your last review on this PR. 

I really felt bad about this, hope on tomorrow. It should bring fresh state.

lucasfcosta: Hey buddy, there's no need to feel bad about it. We're very happy that you want to contribute. Please don't worry about that 😃 

Basically, when doing a `rebase -i` you will be prompted with a list of commits and a few actions to do with them.

Whenever you want to drop a commit you can just use the `drop` word in front of it.

In order to ammend a commit into the last one you can use `fixup` and you can also use `squash` to ammend commits and change their messages as well.

After doing it a few times you will see it's simple, you just gotta try, it's totally normal to make mistakes the first times you try it.

test/assert.js
lucasfcosta: Hi @brutalcrozt, thanks for the answer.

I'm sorry but I still didn't understand how this could be correct, since `delta` cannot be a `string`, which is what you're passing by using `'value'` as the last argument.

Can anyone else please review & confirm just for us to make sure we're not missing any detail?

brutalcrozt: Hey brother, i think no need to ask third person point of view. I assume there is misunderstanding, i can say it correct just because run node test resulting green and travis also green. Have i show you [this](https://runkit.com/adi/5a16d0989cd06300128397b2) ?
Sorry i also make mistake again here, i mislead you. you are reviewing the `assert.changes` but i am comment about `assert.changesBy`.

---
I think that `assert.changes` pass because the key or property is be null when reaching this [line](https://github.com/chaijs/chai/blob/master/lib/chai/interface/assert.js#L2498).
do i need make additional change about this `assert.changes` ?
If you don't mind, please addressing `assert.changesBy` on other line. When you feel still necessary :smile: .

lucasfcosta: Ah, I understand this now. You were right about the tests passing.

The reason why I was confused was because of [this line](https://github.com/chaijs/chai/blob/master/lib/chai/interface/assert.js#L2496), which uses the third argument as a message when the second one is a `function` and we only have three arguments.

However, would you mind just changing the `value` string to `example message` or something like that? Just to make sure the test looks clear.

Great work, btw! Thanks 😄

lucasfcosta:
I got it now, I just had a minor comment about the `value` string.

Also, IMO we should change this API in the future to make it more consistent. I think that checking `props` on `functions` is a valid and not uncommon use case and therefore should be supported on this interface in the same way it is supported on the others.

Right now we need to pass a `message` that is `truthy` in order to do that in this interface.

test/assert.js
brutalcrozt: Done, let me know if this still require some change.

---
Agreed, we should add it to next milestone. 4.2.0 ?
Kinda related to #1090, but unfortunately i still have issue to formulate the word. Let's continue there :wink:

brutalcrozt:
😄 ahh a burst of happiness. Finally i got the point of rebase, also i can address my own misunderstanding.
Thanks for being patient 💘

lucasfcosta:
@brutalcrozt I'm afraid I won't be able to help regarding that since I don't use the web interface.

I'm also not sure I understand your question correctly.

But please make sure to check your `git` configurations. If you take a look at your commit you will be able to see that it contains an `invalid-mail-address` as one of the committers. Maybe [this link](https://help.github.com/articles/setting-your-commit-email-address-in-git/) can help you.

brutalcrozt:
Nvm, the question is not really important.
Is it the reason you don't approve it?
As I remember the first commit is when I use live CD OS. I am in a rush so not configure the email properly.
I am not sure I know how to change the first commit email. Should I reopen new PR instead? :wink:

lucasfcosta:
@brutalcrozt please try this: https://stackoverflow.com/questions/3042437/change-commit-author-at-one-specific-commit

brutalcrozt:
Okay, now it's green? :heart_eyes:  . Thanks for believe in me.

lucasfcosta:
Yay! You're almost there buddy, you just gotta change the commit message following these guidelines: https://conventionalcommits.org/

brutalcrozt:
done

meeber:
@lucasfcosta Any chance you could take a look at this? You have an open review blocking merge :D

lucasfcosta:
Oh, I'm so sorry I missed this for so long!

PR #822, opened 2016-10-03
add typescript definitions to chai

goenning:
I've been using chai with all my TypeScript projects and having the Chai core team maintaining the definitions would be perffect.

@keithamus are you still interested on it?

This PR includes both the definitions and test cases. It's basically a copy from https://github.com/DefinitelyTyped/DefinitelyTyped/tree/types-2.0/chai

This PR should close chaijs/chai/issues/650 and types/npm-chai/issues/14

PR #1092, opened 2017-11-23
Copied typings from DefinitelyTyped

JoshuaKGoldberg:
Included them as `lib/chai.d.ts`, with `--noEmit` tests run in `Makefile` against `test/typings/*.ts`. TypeScript@2.6.1 is now a dev dependency.

Closes https://github.com/chaijs/chai#650 and https://github.com/types/npm-chai#14.

I intentionally copy & pasted what's now used under `@types/chai` despite it being a little... sloppy in some parts (why the Resharper comment??). Improving the typings would be great as followup issues IMO.

Non-merge-conflict version of https://github.com/chaijs/chai/pull/822.

~~_(note: I don't have `make` locally, so I'm using AppVeyor/Travis as my build machine. Ignore the force pushes until it works!)_~~ AppVeyor failures look unrelated. Can this get merged anyway? :smile:

### DIFF
--- a/lib/chai.d.ts
+++ b/lib/chai.d.ts
@@ -1,0 +1,1516 @@
+declare namespace Chai {
+
+    interface ChaiStatic {
+        expect: ExpectStatic;
+        should(): Should;
+        /**
+         * Provides a way to extend the internals of Chai
+         */
+        use(fn: (chai: any, utils: any) => void): ChaiStatic;
+        assert: AssertStatic;
+        config: Config;
+        AssertionError: typeof AssertionError;
+        version: string;
+    }
+
+    export interface ExpectStatic extends AssertionStatic {
+        fail(actual?: any, expected?: any, message?: string, operator?: Operator): void;
+    }
+
+    export interface AssertStatic extends Assert {
+    }
+
+    export interface AssertionStatic {
+        (target: any, message?: string): Assertion;
+    }
+
+    export type Operator = string; // "==" | "===" | ">" | ">=" | "<" | "<=" | "!=" | "!==";
+
+    export type OperatorComparable = boolean | null | number | string | undefined | Date;
+
+    interface ShouldAssertion {
+        equal(value1: any, value2: any, message?: string): void;
+        Throw: ShouldThrow;
+        throw: ShouldThrow;
+        exist(value: any, message?: string): void;
+    }
+
+    interface Should extends ShouldAssertion {
+        not: ShouldAssertion;
+        fail(actual: any, expected: any, message?: string, operator?: Operator): void;
+    }
+
+    interface ShouldThrow {
+        (actual: Function): void;
+        (actual: Function, expected: string|RegExp, message?: string): void;
+        (actual: Function, constructor: Error|Function, expected?: string|RegExp, message?: string): void;
+    }
+
+    interface Assertion extends LanguageChains, NumericComparison, TypeComparison {
+        not: Assertion;
+        deep: Deep;
+        ordered: Ordered;
+        nested: Nested;
+        any: KeyFilter;
+        all: KeyFilter;
+        a: TypeComparison;
+        an: TypeComparison;
+        include: Include;
+        includes: Include;
+        contain: Include;
+        contains: Include;
+        ok: Assertion;
+        true: Assertion;
+        false: Assertion;
+        null: Assertion;
+        undefined: Assertion;
+        NaN: Assertion;
+        exist: Assertion;
+        empty: Assertion;
+        arguments: Assertion;
+        Arguments: Assertion;
+        equal: Equal;
+        equals: Equal;
+        eq: Equal;
+        eql: Equal;
+        eqls: Equal;
+        property: Property;
+        ownProperty: OwnProperty;
+        haveOwnProperty: OwnProperty;
+        ownPropertyDescriptor: OwnPropertyDescriptor;
+        haveOwnPropertyDescriptor: OwnPropertyDescriptor;
+        length: Length;
+        lengthOf: Length;
+        match: Match;
+        matches: Match;
+        string(string: string, message?: string): Assertion;
+        keys: Keys;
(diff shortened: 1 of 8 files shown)


demurgos:
@keithamus 
This looks like a good start effectively, but I am not sure if it is mergeable yet due to the type reference to `assertion-error`.

lib/chai.d.ts
```diff
@@ -0,0 +1,1518 @@
+// <reference types="assertion-error"/>
```
demurgos: How does it deal with this type reference? Does it forces the user to install the types for `assertion-error` itself? This may pose issues due to versions mismatch between the version actually used by chai and the one uplifted to the consumer project.

lucasfcosta: Can't we bundle an specific version of [this type definition](https://github.com/types/npm-assertion-error) to be used with Chai? This way we could always ensure it would be right (instead of depending on peer-deps).

(I don't understand much about TypeScript, so please let me know if I misunderstood what you've just said)

demurgos: The link you posted refers to the types for the _typings_ tool. This tool was used before Typescript 2, because typescript did not support types from npm. These repos are currently (slowly) maintained for older tools but the recommended way is to publish the types to the library's repository. If this is impossible, the solution is to use [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped). Definitions in this repo are automatically published to the `@types` npm organization.

There are two solutions for the current issue:
- Adding type definitions to `assertion-error` (better)
- adding [`@types/assertion-error`](https://www.npmjs.com/package/@types/assertion-error) as a normal dependency to `package.json` (quicker)

lucasfcosta: The first option seems more adequate indeed.

[As I've seen in the `assertion-error` repo](https://github.com/chaijs/assertion-error), it's not that big, so it might be a simple issue to attack.

JoshuaKGoldberg: Oh, I had a note to deal with this and forgot about it...

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/assertion-error/index.d.ts

+1 to adding type definitions to `assertion-error` and -1 to adding any extra dependencies to `package.json`.  I'll tackle that now.

JoshuaKGoldberg: Actually, just looking through this code, it looks like `AssertionError` is manually copy & pasted below. The comment above is `//` instead of `///` and so doesn't do anything. I'll remove it.

JoshuaKGoldberg:
@demurgos I've removed the reference... but looks like coverage is calculated as decreased on the PR.

demurgos:
Thanks for update, I'll take a look at it.

Regarding coverage, I don't see any change: https://codecov.io/gh/chaijs/chai/pull/1092 Is it the right link? (Or do you only see the reduction in local ?)

JoshuaKGoldberg:
Huh, I was seeing a coveralls integration saying it was reduced. Never mind!

demurgos:
I am having trouble to run the tests in your branch. Do you just use `npm install && test-typescript`? (module `chai` not found)

demurgos:
Hi,
I added some commits on top of your branch: https://github.com/demurgos/chai/commits/types.

Here are the main changes:
- Fix the Typescript tests: solve type definitions resolution and global namespace modification (should)
- Add TS options enforcing strict rules
- Reexport `AssertionError` from `assertion-error`: **This currently relies on the unpublished [`JoshuaKGoldberg/assertion-error#types`](https://github.com/JoshuaKGoldberg/assertion-error/tree/types)** branch.

demurgos:
The types for chai [were merged](https://github.com/chaijs/assertion-error/pull/11) and [the version was bumped](https://github.com/chaijs/assertion-error/pull/12). We're waiting for the new version to be published to npm but it should happen soon.

I think I'll update my branch and send a new PR superseding this one once the types are on npm. (See previous message)

meeber:
I don't have any experience with TypeScript, and don't have the time to get smart on it, so I'll defer to other contributors and maintainers on this PR. Question though: What does merging this PR mean in terms of future changes to Chai? In particular, do all future new assertions and changes to existing assertions require corresponding updates to these new TypeScript-specific files before they can be safely merged?

demurgos:
Additions to the Javascript code will need to be documented in the type definitions to be usable from Typescript.

You have two ways to approach this:
- Require type definitions to always document the current JS code (enforce synchronization)
- Wait for Typescript users to send PRs to update the types when they need a feature.

meeber:
What I'm most concerned about is breaking changes to existing assertions. If a breaking change is made to an existing assertion, is there an expectation that the type definitions need to be synchronized prior to the next semver major release, or is it generally understood and accepted by the community that such synchronization can happen in a patch after the major release?

demurgos:
Right, I mainly addressed additions.
First of all, types should be ideally synced so what I am describing below are trade-offs that violate the purpose of types or semver in favor of easier maintainance. I don't like it personally but will try to provide you the best description:

- Incomplete types are the less harmful: if a function is not defined, TS users cannot use it but they'll just send a PR and it can be treated as patch. This happens when a feature is added and the types are not updated.
- Types with extra symbols because of removal in the code without updating the types are more misleading. Still, they are usually found rather quickly and consumers can simply avoid to use them until the types are updated. This may be tolerated.
- Invalid types due to changes in the function signatures or types of exported symbols are the worst and should be avoided. If a function requires a `string` argument but then is changed to a `number` argument then TS users will have to manually bypass typechecking around this function to use it. This is usually not tolerated.

---

From a practical point of view, the goal of types is to catch some errors before running the code. Because of this, I'd say that updating the types when breaking changes occur is when it's the most important to help with migration: types become very valuable because they allow to immediately find the lines that need to be updated (or most of them).

demurgos:
I updated my branch to use the new `assertion-error` version. I submitted [a new PR](#1119) where we can continue the discussion (based on this PR, with fixes and changes discussed above).

PR #1095, opened 2017-11-29
fix: Avoid repeated String#slice calls in stringDistance

bmeurer:
The `stringDistance` function calls `strA.slice(0, -1)` and `strB.slice(0, -1)`
multiple times, which is a bit of a waste of time here. JavaScript
engines cannot generally eliminate the duplicated calls easily, so
it's better to avoid the redundant calls altogether.

This improves the chai test on the [web-tooling-benchmark](https://github.com/v8/web-tooling-benchmark) by
around 8% when run with upcoming V8 6.4.

### DIFF
--- a/lib/chai/utils/proxify.js
+++ b/lib/chai/utils/proxify.js
@@ -112,10 +112,12 @@ function stringDistance(strA, strB, memo) {
     if (strA.length === 0 || strB.length === 0) {
       memo[strA.length][strB.length] = Math.max(strA.length, strB.length);
     } else {
+      var sliceA = strA.slice(0, -1);
+      var sliceB = strB.slice(0, -1);
       memo[strA.length][strB.length] = Math.min(
-        stringDistance(strA.slice(0, -1), strB, memo) + 1,
-        stringDistance(strA, strB.slice(0, -1), memo) + 1,
-        stringDistance(strA.slice(0, -1), strB.slice(0, -1), memo) +
+        stringDistance(sliceA, strB, memo) + 1,
+        stringDistance(strA, sliceB, memo) + 1,
+        stringDistance(sliceA, sliceB, memo) +
           (strA.slice(-1) === strB.slice(-1) ? 0 : 1)
       );
     }


lucasfcosta:
The code LGTM!
Thanks for your insight into this, it's great to have this kind of input in Chai's codebase 😄 

Btw, I've taken a look at the failing checks:
* `validate-commit-msg` - Fails because the commit message does not follow the rules specified here: https://conventionalcommits.org/. For this case I think we can use the `fix:` prefix
* `continuous-integration/appveyor/pr` - Fails because the build would run as a Visual Studio project. Maybe setting this build as `Script` as suggested would be enough, but since we already run a reliable build on TravisCI, I think this is redundant. Is there any significant change between running it on `AppVeyor` and `TravisCI`?

keithamus:
Sorry - the appveyor stuff is a red herring in this project, I accidentally set it up for this rather than another project. I'll try to disable it.

bmeurer:
So what about the validate-commit-msg? Where can I see why it fails?

lucasfcosta:
Hi @bmeurer,

You can see the accepted prefixes on this link: https://conventionalcommits.org/

I think that using `perf:` for this case would be perfect.

keithamus:
BTW `validate-commit-msg` shouldn't block the merge. If someone reviews this code we can squash-and-merge.

lucasfcosta:
Great! This LGTM too.

If you want to rebase it I'll be waiting to merge, otherwise just leave as comment and we'll merge as soon as we see it.

Thanks!

bmeurer:
Please just merge the change and make the appropriate changes to the commit message.

sophiebits:
Does this even need to use slice at all? It's easy to convert this code to just pass around lengths – but better yet, it can just be a loop without recursion which I just wrote:

```js
function stringDistance(strA, strB, memo) {
  var memo = [];
  // `memo` is a two-dimensional array containing distances.
  // memo[i][j] is the distance between strA.slice(0, i) and
  // strB.slice(0, j).
  for (var i = 0; i <= strA.length; i++) {
    memo[i] = Array(strB.length + 1).fill(0);
  }

  for (var i = 0; i <= strA.length; i++) {
    for (var j = 0; j <= strB.length; j++) {
      if (i === 0) {
        memo[i][j] = j;
      } else if (j === 0) {
        memo[i][j] = i;
      } else {
        memo[i][j] = Math.min(
          memo[i - 1][j] + 1,
          memo[i][j - 1] + 1,
          memo[i - 1][j - 1] +
            (strA.charCodeAt(i - 1) === strB.charCodeAt(j - 1) ? 0 : 1)
        );
      }
    }
  }

  return memo[strA.length][strB.length];
}
```

I don't have the benchmark on my machine so it's hard for me to compare easily. I verified it does seem to give the same results though which should not be surprising. @bmeurer Do you want to run the benchmark?

keithamus:
@sophiebits Running a benchmark would be awesome! If it works out faster I'd be very happy with a new PR being made to reflect those changes.

sophiebits:
I actually see some more opportunities for improvement here so I'll try to get the benchmark set up locally.

sophiebits:
(Posted #1098.)

PR #1098, opened 2017-12-02
perf: Optimize proxify and stringDistance

sophiebits:
- Fill 2D array with ints upfront to reduce property access cost
- Change from recursive to simpler iterative (DP) solution
- Add cap parameter to stringDistanceCapped to limit computation
- Make candidate generation use a simple loop to avoid allocating unnecessary arrays and to call stringDistance only once on each pair of strings instead of every time in the sort callback

This improves chai perf by about 13% on @bmeurer's https://github.com/v8/web-tooling-benchmark.

### DIFF
--- a/lib/chai/utils/proxify.js
+++ b/lib/chai/utils/proxify.js
@@ -49,19 +49,31 @@ module.exports = function proxify(obj, nonChainableMethodName) {
             nonChainableMethodName + '".');
         }
 
-        var orderedProperties = getProperties(target).filter(function(property) {
-          return !Object.prototype.hasOwnProperty(property) &&
-            builtins.indexOf(property) === -1;
-        }).sort(function(a, b) {
-          return stringDistance(property, a) - stringDistance(property, b);
+        // If the property is reasonably close to an existing Chai property,
+        // suggest that property to the user. Only suggest properties with a
+        // distance less than 4.
+        var suggestion = null;
+        var suggestionDistance = 4;
+        getProperties(target).forEach(function(prop) {
+          if (
+            !Object.prototype.hasOwnProperty(prop) &&
+            builtins.indexOf(prop) === -1
+          ) {
+            var dist = stringDistanceCapped(
+              property,
+              prop,
+              suggestionDistance
+            );
+            if (dist < suggestionDistance) {
+              suggestion = prop;
+              suggestionDistance = dist;
+            }
+          }
         });
 
-        if (orderedProperties.length &&
-            stringDistance(orderedProperties[0], property) < 4) {
-          // If the property is reasonably close to an existing Chai property,
-          // suggest that property to the user.
+        if (suggestion !== null) {
           throw Error('Invalid Chai property: ' + property +
-            '. Did you mean "' + orderedProperties[0] + '"?');
+            '. Did you mean "' + suggestion + '"?');
         } else {
           throw Error('Invalid Chai property: ' + property);
         }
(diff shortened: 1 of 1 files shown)


sophiebits:
I didn't try that but it makes sense. I did try hoisting the 2D array as a static 41x41 array (and bailing out on names longer than 40 chars) and it didn't seem to help much. Not sure where the bottleneck is now.

lib/chai/utils/proxify.js
```diff
@@ -89,36 +101,44 @@ module.exports = function proxify(obj, nonChainableMethodName) {
...
-        stringDistance(sliceA, strB, memo) + 1,
-        stringDistance(strA, sliceB, memo) + 1,
-        stringDistance(sliceA, sliceB, memo) +
-          (strA.slice(-1) === strB.slice(-1) ? 0 : 1)
+  var memo = [];
+  // `memo` is a two-dimensional array containing distances.
+  // memo[i][j] is the distance between strA.slice(0, i) and
+  // strB.slice(0, j).
+  for (var i = 0; i <= strA.length; i++) {
+    memo[i] = Array(strB.length + 1).fill(0);
```
mathiasbynens: `new Array(…)` is faster than `Array(…)` in V8

mathiasbynens: Consider using `new Array(…)` instead of `Array(…)`. The `[[Call]]` syntax lacks [elements kind tracking](https://v8project.blogspot.com/2017/09/elements-kinds-in-v8.html) in V8 and is therefore less efficient.

bmeurer:
In this case it doesn't matter, since the arrays contain only small integers.

lib/chai/utils/proxify.js
```diff
@@ -89,36 +101,44 @@ module.exports = function proxify(obj, nonChainableMethodName) {
...
-  if (!memo) {
-    // `memo` is a two-dimensional array containing a cache of distances
-    // memo[i][j] is the distance between strA.slice(0, i) and
-    // strB.slice(0, j).
-    memo = [];
-    for (var i = 0; i <= strA.length; i++) {
-      memo[i] = [];
-    }
+function stringDistanceCapped(strA, strB, cap) {
+  if (Math.abs(strA.length - strB.length) >= cap) {
```
lucasfcosta: This was a good catch 👍

lucasfcosta:
Whoah! I'm really glad that we have so many awesome contributors that are also performance specialists.

I've taken a look through the code and the logic LGTM but since I don't feel like I could give any other valuable insights when it comes to performance I'll be happy to review this again when you are happy with it.

Thanks everyone for your collaboration, you are awesome 😄

sophiebits:
I don't think I'll get a chance to look back at this so feel free to merge! Making it `2 x n` and replacing `1 - i` for `i - 1` should work though if you want to test that yourself.

bmeurer:
Thanks a lot @sophiebits and @keithamus! Excited to see this happening. Anyone interested to explore the [memory efficient variant of the Levenshtein algorithm](https://www.codeproject.com/Articles/13525/Fast-memory-efficient-Levenshtein-algorithm)? That should give another nice boost. 👍

not-an-aardvark:
This improvement looks great, although I'm confused about why it's affecting performance so much -- ostensibly the `stringDistance` function only gets called when an invalid property has been accessed, which would usually be a result of programmer error. Is there a reason that `stringDistance` ends up getting called in the benchmark?

bmeurer:
The `chai` test in the [web-tooling-benchmark](https://github.com/v8/web-tooling-benchmark) was derived from the chai test suite itself. So it tests the `chai` library using it's own assert/expect machinery. See the [`chai-benchmark.js` file](https://github.com/v8/web-tooling-benchmark/blob/master/src/chai-benchmark.js) for the exact test cases.

meeber:
`stringDistance` is being called in the benchmark due to the "proxify" tests. Search "pizza" in the benchmark file for all occurrences of this. As definitely-an-aardvark suggested, those tests are centered around programmer error, most likely hunger-induced. Although delicious, the pizza tests have the side-effect of causing the benchmarks to overstate the impact of perf improvements in the `stringDistance` function on a typical Chai workload. I don't normally advocate for the removal of pizza, but it may be warranted in the case of benchmarks.

bmeurer:
@meeber But those are only 7 tests?

meeber:
@bmeurer Yup, just verified those are the only tests calling `stringDistance`. Based on an admittedly small sample size, removing those tests seems to have a noticeable impact on the benchmark in Chai 4.1.2, but not very noticeable in the Chai master branch (which suggests that these perf improvements did their job!) Either way, those tests involve a code path that's only hit when a developer writes a non-existent assertion (e.g., `expect(true).to.be.ture`), so they're not really meaningful in regard to benchmarking actual Chai workloads. On the other hand, faster is better in every code path, and it's valuable to know if some change to the JavaScript engine negatively impacts performance in even a rarely hit code path.

bmeurer:
@meeber Thanks for the explanation. Interesting to see that it made such a big difference. I'll think we still keep the 7 tests following your reasoning, especially since in the near future when we roll a new version of Chai, the impact will be less noticeable.

keithamus:
Yes I suppose it is worth noting - on a typical users test suite, _passing tests_ won't hit these code paths. This likely makes most of a difference to a users development machine, where they'll see this paths hit from typos.

Recently I've been considering how we can improve test times for passing tests, so that tests in typical workloads (e.g. CI) run as quick as possible. One example of slowdown that is instric to the architecture of chai at current - is that we format error messages even for passing tests.

bmeurer:
@keithamus This is a good point. We've observed that error message formatting is indeed quite significant for chai.

Issue #1116, opened 2018-01-05
The `assert.fail([message])` format should be supported for compatibility with NodeJS assert interface


s-leroux:
The NodeJS assert interface supports being called with 1 (or 0) arguments:
https://nodejs.org/api/assert.html#assert_assert_fail_message

But currently Chai's `assert.fail` only supports being called with 2 (or more arguments) leading to that idiom to fail with a custom error message:

    assert.fail(0,1, "Custom error message");

The interface should be modified so we can write that instead:

    assert.fail("Custom error message");

PR #1117, opened 2018-01-05
feat: Add the assert.fail([message]) interface

s-leroux:
In relation to #1116 the `assert.fail` interface should accept only 1 arguments to fail with a custom message.

### DIFF
--- a/lib/chai/interface/assert.js
+++ b/lib/chai/interface/assert.js
@@ -41,9 +41,17 @@ module.exports = function (chai, util) {
   };
 
   /**
+   * ### .fail([message])
    * ### .fail(actual, expected, [message], [operator])
    *
    * Throw a failure. Node.js `assert` module-compatible.
+   *
+   *     assert.fail();
+   *     assert.fail("custom error message");
+   *     assert.fail(1, 2);
+   *     assert.fail(1, 2, "custom error message");
+   *     assert.fail(1, 2, "custom error message", ">");
+   *     assert.fail(1, 2, undefined, ">");
    *
    * @name fail
    * @param {Mixed} actual
@@ -55,6 +63,13 @@ module.exports = function (chai, util) {
    */
 
   assert.fail = function (actual, expected, message, operator) {
+    if (arguments.length < 2) {
+        // Comply with Node's fail([message]) interface
+
+        message = actual;
+        actual = undefined;
+    }
+
     message = message || 'assert.fail()';
     throw new chai.AssertionError(message, {
         actual: actual
--- a/lib/chai/interface/expect.js
+++ b/lib/chai/interface/expect.js
@@ -10,9 +10,17 @@ module.exports = function (chai, util) {
   };
 
   /**
+   * ### .fail([message])
    * ### .fail(actual, expected, [message], [operator])
    *
    * Throw a failure.
+   *
+   *     expect.fail();
+   *     expect.fail("custom error message");
+   *     expect.fail(1, 2);
+   *     expect.fail(1, 2, "custom error message");
+   *     expect.fail(1, 2, "custom error message", ">");
+   *     expect.fail(1, 2, undefined, ">");
    *
    * @name fail
    * @param {Mixed} actual
@@ -24,6 +32,11 @@ module.exports = function (chai, util) {
    */
 
   chai.expect.fail = function (actual, expected, message, operator) {
+    if (arguments.length < 2) {
+        message = actual;
+        actual = undefined;
+    }
+
     message = message || 'expect.fail()';
     throw new chai.AssertionError(message, {
         actual: actual
--- a/lib/chai/interface/should.js
+++ b/lib/chai/interface/should.js
@@ -42,9 +42,18 @@ module.exports = function (chai, util) {
     var should = {};
 
     /**
+     * ### .fail([message])
      * ### .fail(actual, expected, [message], [operator])
      *
      * Throw a failure.
+     *
+     *     should.fail();
+     *     should.fail("custom error message");
+     *     should.fail(1, 2);
+     *     should.fail(1, 2, "custom error message");
+     *     should.fail(1, 2, "custom error message", ">");
+     *     should.fail(1, 2, undefined, ">");
+     *
      *
      * @name fail
      * @param {Mixed} actual
@@ -56,6 +65,11 @@ module.exports = function (chai, util) {
      */
 
     should.fail = function (actual, expected, message, operator) {
+      if (arguments.length < 2) {
+          message = actual;
+          actual = undefined;
+      }
+
       message = message || 'should.fail()';
       throw new chai.AssertionError(message, {
           actual: actual
--- a/test/assert.js
+++ b/test/assert.js
@@ -14,10 +14,24 @@ describe('assert', function () {
     }, "expected foo to equal `bar`");
   });
 
-  it('fail', function () {
-    chai.expect(function () {
-      assert.fail(0, 1, 'this has failed');
-    }).to.throw(chai.AssertionError, /this has failed/);
+  describe("fail", function() {
+    it('should accept a message as the 3rd argument', function () {
+      err(function() {
+        assert.fail(0, 1, 'this has failed');
+      }, /this has failed/);
+    });
+
+    it('should accept a message as the only argument', function () {
+      err(function() {
+        assert.fail('this has failed');
+      }, /this has failed/);
+    });
+
+    it('should produce a default message when called without any arguments', function () {
+      err(function() {
+        assert.fail();
+      }, /assert\.fail()/);
+    });
   });
 
   it('isTrue', function () {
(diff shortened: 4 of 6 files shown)


s-leroux:
OK @keithamus, I will change `should.fail` and `expect.fail` too.

s-leroux:
Done.

I noticed `chai.js` was automatically modified while running the tests. But I left it out of the commit since it incorporated changes that were not caused by my own modifications.

lib/chai/interface/assert.js
```diff
@@ -41,6 +41,7 @@ module.exports = function (chai, util) {
   };
 
   /**
+   * ### .fail([message])
    * ### .fail(actual, expected, [message], [operator])
```
meeber: @chaijs/chai Anyone know how the website generation script handles listing multiple function signatures in this way? Or the proper way to do this with jsdoc in general? I struggled with this problem quite a bit in https://github.com/chaijs/check-error/pull/18, and ended up including a single function signature at the top (with optional parameters) but documenting the alternate signatures in the jsdoc body. Not sure what the best approach is.

lib/chai/interface/expect.js
```diff
@@ -10,6 +10,7 @@ module.exports = function (chai, util) {
   };
 
   /**
+   * ### .fail([message])
    * ### .fail(actual, expected, [message], [operator])
```
meeber: Same question here regarding function signatures.

lib/chai/interface/should.js
```diff
@@ -42,6 +42,7 @@ module.exports = function (chai, util) {
     var should = {};
 
     /**
+     * ### .fail([message])
      * ### .fail(actual, expected, [message], [operator])
```
meeber: Same question here regarding function signatures.

test/assert.js
```diff
@@ -20,6 +20,18 @@ describe('assert', function () {
     }).to.throw(chai.AssertionError, /this has failed/);
   });
 
+  it('fail', function () {
+    chai.expect(function () {
```
meeber: Rater than using `chai.expect.to.throw`, I think we can use the same `err()` helper function for these `assert` interface tests that are used in the `expect` and `should` interface tests. Doing so has the advantage of the test making sure implementation frames aren't in the Assertion Error's stack trace.

test/assert.js
```diff
@@ -20,6 +20,18 @@ describe('assert', function () {
     }).to.throw(chai.AssertionError, /this has failed/);
   });
 
+  it('fail', function () {
```
meeber: I'd like the test descriptions to be more descriptive than "fail", even if it's as simple as "fail (1 arg)" and "fail (no args)". Another option would be to create a new `describe` block for `fail`, and then put descriptive `it` blocks inside of it.

test/assert.js
```diff
@@ -20,6 +20,18 @@ describe('assert', function () {
...
+    chai.expect(function () {
+      assert.fail('this has failed');
+    }).to.throw(chai.AssertionError, /this has failed/);
+  });
+
+  it('fail', function () {
+    chai.expect(function () {
+      assert.fail();
+    }).to.throw(chai.AssertionError, /assert\.fail/);
+  });
```
meeber: Same comments as above regarding description and `err` helper function.

test/expect.js
```diff
@@ -239,7 +239,19 @@ describe('expect', function () {
   it('fail', function () {
     err(function() {
       expect.fail(0, 1, 'this has failed');
-    }, /this has failed/);
+    }, 'this has failed');
+  });
+
+  it('fail', function () {
```
meeber: Same comment as the `assert` interface test regarding `it` block description.

test/expect.js
```diff
@@ -239,7 +239,19 @@ describe('expect', function () {
...
+    err(function() {
+      expect.fail('this has failed');
+    }, 'this has failed');
+  });
+
+  it('fail', function () {
```
meeber: Same comment as the `assert` interface test regarding `it` block description.

test/should.js
```diff
@@ -239,6 +239,18 @@ describe('should', function() {
     }, 'this has failed');
   });
 
+  it('fail', function () {
```
meeber: Same comment as the `assert` interface test regarding `it` block description.

test/should.js
```diff
@@ -239,6 +239,18 @@ describe('should', function() {
...
+    err(function() {
+      should.fail('this has failed');
+    }, 'this has failed');
+  });
+
+  it('fail', function () {
```
meeber: Same comment as the `assert` interface test regarding `it` block description.

keithamus:
> Also, my initial feeling is that this constitutes a breaking change, since it's changing how parameter handling is performed on an existing assertion. 

My feeling is this is a feature. My reasoning is that the behaviour of `expect.fail` with one argument, certainly as per the docs, is _undefined_. Because the behavior is undefined, any existing behavior is able to change. To rephrase: anyone using `expect.fail` with less than two arguments is relying on _ondocumented bevhaior_.

meeber:
@keithamus Good point. I agree. Not a breaking change.

s-leroux:
First, thank you @keithamus and @meeber for having reviewed that PR.

I don't think this is a breaking change since currently `fail` is documented to only accepting 2, 3 or 4 arguments. As a matter of fact, I was initially not sure if I should mark that as a "feature" and a "fix"--since, [according to the website](http://chaijs.com/guide/styles/#assert) (emphasis mine):

> The assert style is exposed through assert interface. This provides the **classic assert-dot notation, similar to that packaged with node.js**. This assert module, however, **provides several additional tests** and is browser compatible.


So, I understand `chai.assert` as being a super-set of node.js `.assert`. And  [`assert.fail` with only one argument](https://nodejs.org/api/assert.html#assert_assert_fail_message) is indeed supported there.

Anyway, whatever is your decision, I will follow you.

----

> I'd like the test descriptions to be more descriptive than "fail", even if it's as simple as "fail (1 arg)" and "fail (no args)". Another option would be to create a new `describe` block for fail, and then put descriptive it blocks inside of it.

Of course.

> Rater than using `chai.expect.to.throw`, I think we can use the same `err()` helper function for these assert interface tests that are used in the expect and should interface tests.

I will fix that. I just copied-pasted existing tests. And I assumed there was some obscure reason for not using it in the `assert` test suite.

----

> Anyone know how the website generation script handles listing multiple function signatures in this way? Or the proper way to do this with jsdoc in general?

I have no idea how to do it properly. Any suggestion welcome!

----

> There may also need to be some follow-up actions regarding TypeScript per #1119, but I don't think that needs to be resolved in this PR.

For now, I have absolutly no idea what your talking about ;)
So I suggest to let that out of this PR, but I would be happy to help with that in a second time.

----

> If other maintainers agree that this is the case, then prior to being merged, the commits in this PR should be squashed into a single one

I will do that once I will have your green light for the other changes.

lib/chai/interface/assert.js
s-leroux: There is a possible solution for handling multiple function signature used in the [`assert.throws`](http://chaijs.com/api/assert/#method_throws) documentation. But I'm not sure this is the _proper_ way of doing though: as a user, this is only by looking at the example section I was able to understand the various signatures for that method.

s-leroux: FWIW, I passed the source code through `dox` (which I assume to be the jsdoc parser used for the website). Here is the relevant part of the output:

    "description": {
      "full": "<h3>.fail([message])</h3>\n<h3>.fail(actual, expected, [message], [operator])</h3>\n<p>Throw a failure.</p>",
      "summary": "<h3>.fail([message])</h3>\n<h3>.fail(actual, expected, [message], [operator])</h3>",
      "body": "<p>Throw a failure.</p>"
    },

As you can see, it produces two `<h3>` elements--once for each signature. Which seems quite reasonnable. Maybe we could ping someone at the chai-doc team to check if this is an issue?

meeber: Yup, chai-docs is using `dox`. That behavior looks promising. @astorije might know for sure.

s-leroux:
I fixed the tests, add a little bit of documentation in the method header and squashed the commit.

I left the "double function signature" in the method header since this is definitely the most understandable way to describe the different function signatures and that _seems_ to be supported by `dox`:

```
    * ### .fail([message])
    * ### .fail(actual, expected, [message], [operator])
``` 

But let me know if you want I change that to something else, in case this would cause an issue with the chai-doc tool.

meeber:
Thanks @s-leroux. I like the changes. I'd say let's move forward with the double function signatures unless someone chimes in otherwise.

One thing: [there's a couple of duplicate tests](https://github.com/s-leroux/chai/blob/8a987b4c8b7c45f5280e7e081fb3564e6e0880b7/test/assert.js#L37-L47) that can be deleted.

s-leroux:
Thank you for your review @meeber.
I removed the duplicate tests.

s-leroux:
Thank you @keithamus. It was a pleasure!

PR #1122, opened 2018-01-10
feat: added support for custom object inspectors

DanielSchaffer:
It can be frustrating trying to compare expected/actual objects using the existing string representation, especially for large or complex objects. This feature seeks to remedy that by allowing a developer to control the string representation that gets used. It works by allowing an `Inspector` to be registered using `Utils.addInspector()`. When generating the message, the `inspect()` util function now iterates through any registered inspectors, using the first non-null/undefined representation that gets returned, and falling back to the existing built-in inspector if there are no custom inspectors, or none of them return a value.

- added Utils.addInspector method for registering custom inspectors
- moved existing `inspect()` function to `defaultInspector.js`, and replaced the `inspect()` function with logic for determining the string representation from any available custom inspectors
- added unit tests for above


meeber:
@DanielSchaffer Thanks for the PR. This looks pretty solid and I like the idea. Couple thoughts:

- I wonder if there should be a way to remove inspectors
- Might be worth adding a few "actual usage" tests for each interface (`expect`, `should`, and `assert`) to supplement  the `utilities` test

With that said, I'd advise holding off on making any changes until other maintainers have chimed in.

DanielSchaffer:
Thanks! I don't use a ton of Chai plugins, so I wasn't too sure what all the use cases would be, so I erred on the side of simplistic. My understanding is that there'd basically be two use cases for this - plugin authors adding support for types from 3rd party libraries, likely as part of a broader plugin (I intend to use this as part of a chai-luxon plugin), and developers adding inspectors for their own project-local types.

There could be a case for developers wanting to opt out of inspectors offered by the first case, in which case library authors could offer the inspector as a separate plugin (requiring a second  or alternative `chai.use()` call), though this seems to me like it would be an edge case. The developer could always add their own inspector that would supercede the unwanted one too.

keithamus:
Hey @DanielSchaffer thanks for this PR!

I'm curious what this solves that adding an `inspect` method to your object doesn't? This would be needed for debugging in Node, and also works in Chai.

DanielSchaffer:
To be honest, I hadn't heard of an inspect method until your response, haha. I'm not sure how well known that feature is, and it's probably not terribly necessary today since you can debug in an IDE -
 but regardless, I do think this gives a little added utility:
- Don't need to add a function used only for unit tests/debugging to otherwise production-ready code
- Handle types defined by third parties without messing with their prototype
- Handle plain JS objects
- Can be done per test/spec if needed (e.g. putting the relevant `chai.use` call in a `before` block would work, right?)

meeber:
What I like about this approach is that it doesn't require modifying the object being tested in order to get the custom output in the test.

My thought on being able to remove inspectors is just to have the ability to cleanup after tests. If I'm not mistaken, once a developer registers a custom inspector for one test, that inspector will be active for all tests thereafter. This probably isn't a problem most of the time, but there may be some situations in which its desirable to remove the inspector before proceeding to the next test. Chai's own test suite is such an example.

keithamus:
What I _dislike_ about this proposal is that it affects the entire runtime by mutating internal state within chai. Kind of like having `undefined` as a variable in the pre-es5 days - I could simply call `addInspector(() => '')` and every subsequent chai inspect call is returning useless data. Adding an `inspect` method to your own objects _does not cause this_.

The likelyhood is that any _legitimate_ usecase for an `addInspector` would be to brand check your own types and return an appropriate string. Something like: `addInspector(value => value.constructor === Thingy ? 'Thingy{ ... }' : null)` or `addInspector(value => isThingy(value) ? 'Thingy{ ... }' : null)` or `addInspector(value => value[Symbol.for('thingy')] ? 'Thingy{ ... }' : null)`. The thing is, in an ES6 world there are some several pretty well standardised brand checks you can do - which we can provide hooks for, `constructor` is one, `Symbol.toStringTag` another.


...So, I've been working on refactoring of all the inpsection code (into a lib called `loupe`) - it's been a while coming (too long, but my time is limited) but its a long planned feature. Part of the work which will be done is providing mechanisms like this for custom inspection - there will be a few ways to provide custom inspectors:

 - Adding a method named `inspect`, or a method of either [`[require('util').inspect.custom]`](https://nodejs.org/dist/latest-v9.x/docs/api/util.html#util_util_inspect_custom) or `[Symbol.for('chai/inspect')]`, this is the same mechanism as current, but a little more flexible as you can tuck these methods away in Symbols, which also gives you license to monkey patch them for tests.
 - Passing your constructor, alongside an inspector function, to the registry. If the inspector code lands on an object who's `instanceof` is in the consturctor registry, then that consturctor is run - think `addInspectorForConstructor(MyThingy, value => 'Thingy{ ... }')`. This is nice because - as mentioned above - it doesn't clobber _all inspect calls_, just the ones where the value is a known `instanceof`.
 - If you can't access your instance, but you do have it's `[Symbol.toStringTag]` value, then that - alongside the inspector function - can be added to the registry. Think something like `addInspectorForToStringTag('MyThingy', value => 'Thingy{ ... }')`. Again this has the benefit of not clobbering all inspect calls, only the ones where `value[Symbol.toStringTag] === 'MyThingy'`.

However, in addition to these extension methods, the out-of-the-box behaviour will be much more robust, so classes will naturally look like `ConstructorName { enumerableProp: 1, nonEnumerableProp: 2 }`. More importantly, these representations will be truncated to highlight the _differences_ between two objects, more than just listing every property.

Hopefully this all makes sense and hopefully the new `loupe` features would help cover your use cases @DanielSchaffer

DanielSchaffer:
@keithamus that all sounds AMAZING. Is there anything ready to take a look at yet? Can I help?

keithamus:
@DanielSchaffer right now all the code is a big ugly jumble, I hope to get something up soon though, probably not before the end of this month though.

DanielSchaffer:
Ah, bummer... but totally understandable. Do you have an API spec then? Does it seem like it would be possible to quickly change my implementation in a way that would be forwards-compatible with `loupe`, so that when it's ready, the implementation can just be switched out behind the scenes?

keithamus:
Hi @DanielSchaffer! Thanks for making this PR. We're currently going through and cleaning all currently open PRs. As I mentioned above we're thinking on going in a different direction with `loupe` and I think that will fix the problem you're trying to fix with this PR. Chai@5 will come with a new version of loupe. I suggest checking out Chai 5 when its released and seeing if it addresses your concerns, and if not then please raise an issue and we'll discuss it further!

Please don't take that as a sign that your contribution wasn't worthwhile! We hugely appreciate the time you've put into this and it has definitely caused us to rethink our approach to accomodate your ideals. Please don't let this dissuade you from making more PRs! I really hope we see more contributions from you 😃

Issue #1120, opened 2018-01-07
Ability to show only changes and deletions in diffs, not additions


Mingun:
I use `chai` assertion library and [`chai-like`](https://github.com/zation/chai-like) library for the partial comparing of objects: I am interested in existence and compliance only of those properties that are listed in an `expected` object. Existence of additional properties in an `actual` object permissibly. Unfortunately, when diff is created it is not considered in any way and those properties, existence which for the test it is indifferent get to diff. It would be desirable to have a way to suppress their generation in a diff. Example (runned with mocha):

```js
'use strict';

let chai = require('chai');
let expect = chai.expect;
chai.use(require('chai-like'));

describe("diff test", function() {
  let AST = {
    type: 'grammar',
    rules: [
      { type: 'rule', name: 'rule1' },
      { type: 'rule', name: 'rule2' },
    ],
    // it is a lot more other properties
  };
  it("chai-like", function() {
    expect(AST).like({
      rules: [
        { name: 'rule' },
        { name: 'rule2' }
      ]
    });
  });
});
```

This is current actual output of mocha which is too large in some cases and contains unnecessary details
```console
  1) diff test
       chai-like:

      AssertionError: expected { Object (type, rules) } to be like { Object (rules) }
      + expected - actual

       {
         "rules": [
           {
      -      "name": "rule1"
      -      "type": "rule"
      +      "name": "rule"
           }
           {
             "name": "rule2"
      -      "type": "rule"
           }
         ]
      -  "type": "grammar"
       }

      at Context.<anonymous> (test\test.js:24:17)
```
But I want that diff looked here so:
```diff
 {
   "rules": [
     {
-      "name": "rule1"
+      "name": "rule"
     }
   ]
 }
```
Now there is no opportunity to show diff in that look in what it is required as `AssertionError` has no signs that it is required. Mocha developers advised to create at first the task here before they are able to support this functionality (https://github.com/mochajs/mocha/issues/3193).

I suppose add to an `AssertionError` object new property which could control such behavior of diff.

If the idea is supported, I can make PRs for its implementation in all necessary libraries.

meeber:
Just thinking outloud here: One option would be a modification to [this line](https://github.com/zation/chai-like/blob/5c14498f1c7d4f1f48549b85e3804cfc867d40e1/lib/chai-like.js#L55) in `chai-like` so that instead of passing in the full `object`, pass in a new object that only includes the properties being checked in `object` based on `expected`.

[...]

PR #1124, opened 2018-01-14
fix: remove Chai frames from `.deep.equal` stack

meeber:
I noticed implementation frames in the `.deep.equal` stack trace that @keithamus posted in [this comment](https://github.com/chaijs/chai/issues/1120#issuecomment-357504842). Those frames were present because `.deep.equal` called `.eql` internally without locking the SSFI first. There weren't any tests verifying error messages and stack traces (via `err(fn, msg)`) for `.deep.equal`.


meeber:
Wait don't merge this. There's a problem in the logic: it assumes that flag was previously false. I'll fix in a sec...

meeber:
Fixed rare edge case where the new logic could disable a previously set flag.

Issue #1110, opened 2017-12-18
.lengthOf should work for sets and maps


rijnhard:
sets and maps use a different property `.size` for their length but in the context of chai it should work the same way.

```javascript
expect(new Set([1])).to.have.lengthOf(1);
// currently UncaughtException AssertionError: expected {} to have property 'length'
```

rijnhard:
workaround is:

```javascript
expect(new Set([1])).to.have.property('size', 1);
```

lucasfcosta:
I think this is a very valid suggestion.

If other maintainers agree, I'd love to get this into our codebase.

Btw, this is very similar to #1096.

meeber:
Agreed. May also be worth adding `.size` as an alias to `.lengthOf`.

keithamus:
PRs welcome for this!

sergiorojasa:
I would like to work on this

asbish:
@keithamus I would like to do it by this weekend. Fine with you?

keithamus:
@asbish that would be more than fine. That would be wonderful!

keithamus:
@asbish I have assigned this to myself, and removed the `pull-request-wanted` label - this is so that you have time to make this PR without any one thinking it is free to work on. However, there is no pressure! If you decide you dont or cant work on it any more please let me know 😄

asbish:
@keithamus Thank you for your kinds words! :)

keithamus:
hey @rijnhard thanks for this issue! 

We're adding this in chai 5, we've added this to our Roadmap https://github.com/chaijs/chai/projects/2! We'll be releasing chai 5 soon, but for now I'll close this issue because it is tracked on our roadmap.

PR #1131, opened 2018-01-26
.lengthOf for Maps and Sets

asbish:
As described in [#1110](https://github.com/chaijs/chai/issues/1110), this PR adds support for Map and Set to .lengthOf and chains such as .above, .below, .least, .most and  .within.
I have not added alias 'size' to .lengthOf this time so as not to fix some exist test cases (Levenshtein distance between 'size' and 'pizza' used in the test cases is close). Will you give me some feedback on the above?
Feel free to point out possible mistakes. Thanks!

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -1102,8 +1102,8 @@ module.exports = function (chai, _) {
    *     expect(2).to.equal(2); // Recommended
    *     expect(2).to.be.above(1); // Not recommended
    *
-   * Add `.lengthOf` earlier in the chain to assert that the value of the
-   * target's `length` property is greater than the given number `n`.
+   * Add `.lengthOf` earlier in the chain to assert that the target's `length`
+   * or `size` is greater than the given number `n`.
    *
    *     expect('foo').to.have.lengthOf(3); // Recommended
    *     expect('foo').to.have.lengthOf.above(2); // Not recommended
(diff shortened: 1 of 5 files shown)


meeber:
@asbish Oh nooo we've been foiled again by pizza. Thanks so much for working on this. It's going to be awhile until I can review; working crazy hours this week and then going on vacay next week. I'll try to carve out some time.

lib/chai/core/assertions.js
```diff
@@ -2013,6 +2048,11 @@ module.exports = function (chai, _) {
    *     expect([1, 2, 3]).to.have.lengthOf(3);
    *     expect('foo').to.have.lengthOf(3);
    *
+   * When the target is a map or set, asserts that the target's `size` property.
```
meeber: This reads awkwardly to me. Instead of having this separate line, I suggest updating lines 2045-2046 to read: "Asserts that the target's `length` or `size` is equal to the given number `n`."

And then the new examples could go right beneath the "foo" example.

lib/chai/core/assertions.js
```diff
@@ -2013,6 +2048,11 @@ module.exports = function (chai, _) {
...
+   *
+   *     expect(new Set([1, 2])).to.hove.lengthOf(2);
+   *     expect(new Map([['a', 1], ['b', 2]])).to.hove.lengthOf(2);
```
meeber: There's a typo in these examples: "hove" instead of "have".

lib/chai/core/assertions.js
asbish: Thanks!

asbish:
@meeber Thank you so mush for your review!
I have fixed, including the commit of the indentation in assertions.js at 2124
Following your suggestion have made the documentation much clearer and easier to read:smile:

lib/chai/core/assertions.js
asbish: Thanks!

meeber:
LGTM! Looking for second review from @chaijs/chai

PR #1157, opened 2018-05-10
docs: fix bad `.string` example

### LINKED ISSUE #1156: documentation implies `expect(...).to.have.string(/foo/)` should use RegExp matching
http://www.chaijs.com/api/bdd/#method_string says

```js
expect('foobar').to.have.string(/taco/, 'nooo why fail??');
expect('foobar', 'nooo why fail??').to.have.string(/taco/);
```

This seems to suggest that one can pass a `RegExp` to match against.
It's possible that this was copy/pasted mistakenly from the previous section on `.to.match(/.../)`

`.have.string` does not do `RegExp` matching.

```sh
> grep chai package-lock.json
    "chai": {
      "resolved": "https://registry.npmjs.org/chai/-/chai-4.1.2.tgz",
> node -e "console.log(require('chai').expect('taco').to.have.string(/taco/, 'nooo why fail?'));"
/Users/msamuel/work/jsconf.eu.2018/node_modules/chai/lib/chai/assertion.js:141
      throw new AssertionError(msg, {
      ^

AssertionError: nooo why fail?: expected 'taco' to contain /taco/
    at [eval]:1:52
    ...
```

vs

```sh
> node -e "console.log(require('chai').expect('taco').to.match(/taco/, 'nooo why fail?'));"
Assertion {
  __flags:
   { ssfi: [Function: proxyGetter],
     lockSsfi: undefined,
     object: 'taco',
     message: 'nooo why fail?' } }
```

meeber:
Fixes #1156

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -2189,8 +2189,8 @@ module.exports = function (chai, _) {
    * message to show when the assertion fails. The message can also be given as
    * the second argument to `expect`.
    *
-   *     expect('foobar').to.have.string(/taco/, 'nooo why fail??');
-   *     expect('foobar', 'nooo why fail??').to.have.string(/taco/);
+   *     expect('foobar').to.have.string('taco', 'nooo why fail??');
+   *     expect('foobar', 'nooo why fail??').to.have.string('taco');
    *
    * @name string
    * @param {String} str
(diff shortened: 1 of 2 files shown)


hbetts:
Can this please be published to the website? (This change does not seem to be there - https://www.chaijs.com/api/bdd/#method_string)

PR #1196, opened 2018-09-16
imrove assert message for inherited classes

Yeti-or:
For such cases:
 
`expect(key).to.be.instanceof(ParamedKey);`

before:
```
AssertionError: expected { Object (name, value) } to be an instance of ParamedKey
```

after:
```
AssertionError: expected { Key (name, value) } to be an instance of ParamedKey
```

### DIFF
--- a/lib/chai/utils/objDisplay.js
+++ b/lib/chai/utils/objDisplay.js
@@ -40,7 +40,7 @@ module.exports = function objDisplay(obj) {
         , kstr = keys.length > 2
           ? keys.splice(0, 2).join(', ') + ', ...'
           : keys.join(', ');
-      return '{ Object (' + kstr + ') }';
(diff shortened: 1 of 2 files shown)


Yeti-or:
I don't know is it a fix or feat ?

lib/chai/utils/objDisplay.js
```diff
@@ -40,7 +40,7 @@ module.exports = function objDisplay(obj) {
         , kstr = keys.length > 2
           ? keys.splice(0, 2).join(', ') + ', ...'
           : keys.join(', ');
-      return '{ Object (' + kstr + ') }';
+      return '{ ' + obj.constructor.name + ' (' + kstr + ') }';
```
keithamus: This might want to use `getFunctionName`

PR #1200, opened 2018-09-24
Fix unbalanced quotes in assert.notInclude example

haykam821:
This pull request fixes two lines in the example for `assert.notInclude` using single quotes around the string but also including a single quote in the string:

```js
'this doesn't work'
```

I have changed the erroneous examples to use double quotes around the string instead, so there is no longer a syntax error:

```js
"this doesn't break!"
```


astorije:
@meeber, @keithamus, npmjs has recently invalidated all tokens, so a new one needs to be set up in the `.travis.yml` config file. I hit the same issue very recently at https://github.com/thelounge/thelounge/pull/2716.

PR #1209, opened 2018-11-13
allow addChainableMethod when chainingBehavior = true, it set as method

bluelovers:
> this allow when we wanna get both effect of `addProperty` and `addMethod`/`addChainableMethod`

current

```ts
	let fn = function (v)
	{
		utils.expectTypes(this, ['number']);

		let obj = utils.flag(this, 'object');

		this.assert(
			isInt(obj)
			, 'expected #{this} to be an integer'
			, 'expected #{this} to not be an integer'
			, obj,
		);

		if (typeof v !== 'undefined')
		{
			new chai.Assertion(obj).to.be.deep.equal(v);
		}
	}

	Assertion.addChainableMethod('integer', fn, fn);
```

after

```ts
	let fn = function (v)
	{
		utils.expectTypes(this, ['number']);

		let obj = utils.flag(this, 'object');

		this.assert(
			isInt(obj)
			, 'expected #{this} to be an integer'
			, 'expected #{this} to not be an integer'
			, obj,
		);

		if (typeof v !== 'undefined')
		{
			new chai.Assertion(obj).to.be.deep.equal(v);
		}
	}

	Assertion.addChainableMethod('integer', fn, true);
```

### DIFF
--- a/lib/chai/utils/addChainableMethod.js
+++ b/lib/chai/utils/addChainableMethod.js
@@ -72,8 +72,8 @@ var call  = Function.prototype.call,
 
 module.exports = function addChainableMethod(ctx, name, method, chainingBehavior) {
   if (chainingBehavior === true) {
-    chainingBehavior = method
-  } else ifif (typeof chainingBehavior !== 'function') {
+    chainingBehavior = method;
+  } else if (typeof chainingBehavior !== 'function') {
     chainingBehavior = function () { };
   }
 
--- a/lib/chai/utils/addChainableMethod.js
+++ b/lib/chai/utils/addChainableMethod.js
@@ -71,7 +71,9 @@ var call  = Function.prototype.call,
  */
 
 module.exports = function addChainableMethod(ctx, name, method, chainingBehavior) {
-  if (typeof chainingBehavior !== 'function') {
+  if (chainingBehavior === true) {
+    chainingBehavior = method
+  } else ifif (typeof chainingBehavior !== 'function') {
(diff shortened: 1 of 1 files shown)


meeber:
Hi @bluelovers and thanks for contributing. I don't think the same function should be used for both the `method` and `chainingBehavior` of `addChainableMethod`. In other words, I think your original example is problematic, both the `current` version of passing `fn, fn`, and the `after` version of `fn, true`. In both examples, writing a test like `expect(min).to.be.integer(42)` would cause `fn` to be called twice, and the `this.assert(isInt` part of it would be performed twice.

Instead, I'd recommend:

```js
  let fnChain = function ()
  {
    utils.expectTypes(this, ['number']);

    let obj = utils.flag(this, 'object');

    this.assert(
      isInt(obj)
      , 'expected #{this} to be an integer'
      , 'expected #{this} to not be an integer'
      , obj,
    );
  }

  let fnMethod = function (v)
  {
    let obj = utils.flag(this, 'object');

    if (typeof v !== 'undefined')
    {
      new chai.Assertion(obj).to.be.deep.equal(v);
    }
  }

  chai.Assertion.addChainableMethod('integer', fnMethod, fnChain);
```

bluelovers:
@meeber @vieiralucas 

can i ask here?
how to make it support `.not` chain

`expect(min).to.be.not.integer(42)`

bluelovers:
nvm i know how now

`this` = `Assertion` obj

PR #1216, opened 2018-12-14
Fix build

meeber:
The recently-added `HeadlessChrome` Custom Launcher was being overwritten by the contents of `sauce.browsers.js` when run in a Sauce environment, causing the build to fail.


meeber:
Pinging @chaijs/chai

PR #1220, opened 2018-12-29
WIP: Overhaul `.throw`

meeber:
I've been thinking lately about https://github.com/chaijs/check-error/pull/18. I think some of my decisions in that PR were wrong; in particular, dropping support for checking non-`Error` instances. Although only throwing instances of `Error` (including subclassed errors, like `TypeError`) is a best practice, it's not an important enough issue to justify enforcing that best practice on users in the form of a breaking change. However, there are other issues that are important enough to justify breaking changes:

1. Eliminating the ambiguity of `.throw('waffles')`; currently it asserts that the target function throws either a string that contains "waffles", or an object that has a `.message` property that contains "waffles". Users typically have only one of those scenarios in mind when writing the assertion, and it's an unwelcome surprise for the assertion to pass given the other scenario.

2. Eliminating the ambiguity of `.not.throw(someCriteria)`; currently it asserts that the target function either doesn't throw, or it throws but the thrown value doesn't match one or more of the given criteria. Again, users typically have only one of those scenarios in mind when writing the assertion.

3. Eliminating a reasonable-looking but actually-meaningless assertion from always passing; currently `expect(myFn).to.throw` silently passes, regardless of whether or not `myFn` throws, because `.throw` is only a method-based assertion. This is an unwelcome surprise for users of an assertion library that makes use of property-based assertions (e.g., `.true`). (I'm opposed to jumping to the opposite extreme of removing all property-based assertions from Chai; despite these types of problems, I think property-based assertions are a critical part of Chai's identity and appeal for many users.)

4. Replacing the seldom-used `.throw(ErrorInstance)` behavior, which currently checks if the thrown value is strictly (`===`) equal to `ErrorInstance`, with more useful behavior of checking that the thrown value is deeply equal to `ErrorInstance`. This change was enabled by https://github.com/chaijs/deep-eql/pull/59. 

This PR demonstrates the changes I have in mind in the form of an update to the docs for `.throw`. I believe that implementing these changes would significantly simplify the code (and possibly eliminate the need for a separate `check-error` module altogether), as most of the complexity of the current implementation is caused by handling the ambiguity of the first two issues described above. It'd also lead to the creation of a similar `.error` assertion.

Questions:

1. What are your thoughts on these changes in general?
2. How do you feel about a major release for these changes (and to drop Node v4 and old IE support) prior to the release of anything related to the work being done in the `new-chai` repo?
3. Are the changes suggested here compatible with the work being done in the `new-chai` repo?

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -2447,130 +2447,214 @@ module.exports = function (chai, _) {
   Assertion.addMethod('keys', assertKeys);
   Assertion.addMethod('key', assertKeys);
 
+  // TODO: Document and move to utilities
+  function createErrCriteria(errLike, errMsgMatcher, negate) {
+    var cri = {
+      errLike: errLike,
+      errLikeType: _.type(errLike).toLowerCase(),
+      errMsgMatcher: errMsgMatcher,
+      errMsgMatcherType: _.type(errMsgMatcher).toLowerCase(),
+    };
+
+    // Short-circuit if no criteria is defined.
+    if (cri.errLikeType === 'undefined' &&
+        cri.errMsgMatcherType === 'undefined') {
+      return cri;
+    }
+
+    if (negate) {
+      throw TypeError('errLike and errMsgMatcher must both be undefined when' +
+                      ' negate is true');
+    }
+
+    if (cri.errMsgMatcherType === 'undefined') {
+      // Handle `errMsgMatcher` being passed as the first argument.
+      if (cri.errLikeType === 'string' || cri.errLikeType === 'regexp') {
+        cri.errMsgMatcher = cri.errLike;
+        cri.errMsgMatcherType = cri.errLikeType;
+        cri.errLike = undefined;
+        cri.errLikeType = 'undefined';
+      }
+      return cri;
+    }
+
+    if (cri.errLikeType !== 'function') {
+      throw TypeError('errLike must be a function when errMsgMatcher is' +
+                      ' defined');
+    }
+
+    if (cri.errMsgMatcherType !== 'string' &&
+        cri.errMsgMatcherType !== 'regexp') {
+      throw TypeError('errMsgMatcher must be a string or regexp when' +
+                      ' defined');
+    }
+
+    return cri;
+  }
+
+  // TODO: Document and move to utilities
+  function assertErrCriteria(err, cri, flagMsg, ssfi) {
+    // Short-circuit if no criteria is defined.
+    if (cri.errLikeType === 'undefined' &&
+        cri.errMsgMatcherType === 'undefined') {
+      return;
+    }
+
+    // When `errLike` is the `String` constructor, and `errMsgMatcher` is a
+    // string or regex, assert that the error is a string (as opposed to an
+    // object with a `.message` property) that contains or matches
+    // `errMsgMatcher`.
+    if (cri.errLike === String) {
+      if (cri.errMsgMatcherType === 'string') {
+        new Assertion(err, flagMsg, ssfi, true)
+          .contains.string(cri.errMsgMatcher);
+        return;
+      } else if (cri.errMsgMatcherType === 'regexp') {
+        new Assertion(err, flagMsg, ssfi, true)
+          .is.a('string').that.matches(cri.errMsgMatcher);
+        return;
+      }
+    }
+
+    if (cri.errLikeType === 'function') {
+      new Assertion(err, flagMsg, ssfi, true).is.an.instanceof(cri.errLike);
+    } else if (cri.errLikeType !== 'undefined') {
+      new Assertion(err, flagMsg, ssfi, true).deep.equals(cri.errLike);
+    }
+
+    if (cri.errMsgMatcherType === 'string') {
+      new Assertion(err, flagMsg, ssfi, true)
+        .has.property('message').that.contains.string(cri.errMsgMatcher);
+    } else if (cri.errMsgMatcherType === 'regexp') {
+      new Assertion(err, flagMsg, ssfi, true)
+        .has.property('message').that.matches(cri.errMsgMatcher);
+    }
+  }
+
   /**
-   * ### .throw([errorLike], [errMsgMatcher], [msg])
+   * ### .throw([errLike[, errMsgMatcher[, msg]]])
+   * ### .throw([errMsgMatcher[, msg]])
    *
-   * When no arguments are provided, `.throw` invokes the target function and
-   * asserts that an error is thrown.
+   * When invoked without any arguments, `.throw` invokes the target function
+   * and asserts that it throws.
    *
-   *     var badFn = function () { throw new TypeError('Illegal salmon!'); };
+   *     var badFn = function () { throw TypeError('Illegal salmon!'); };
    *
    *     expect(badFn).to.throw();
    *
-   * When one argument is provided, and it's an error constructor, `.throw`
-   * invokes the target function and asserts that an error is thrown that's an
-   * instance of that error constructor.
+   * When invoked with one argument, and it's a function (typically a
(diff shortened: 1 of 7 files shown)


lib/chai/core/assertions.js
```diff
@@ -2448,135 +2448,138 @@ module.exports = function (chai, _) {
...
    *
-   * When no arguments are provided, `.throw` invokes the target function and
-   * asserts that an error is thrown.
+   * When used as a property assertion or when invoked without any arguments,
+   * `.throw` invokes the target function and asserts that it throws.
    *
-   *     var badFn = function () { throw new TypeError('Illegal salmon!'); };
+   *     var badFn = function () { throw TypeError('Illegal salmon!'); };
    *
+   *     expect(badFn).to.throw;
```
keithamus: This is an interesting addition - but it is also one we tried very much in vain to do back in Chai@2 days. The main issue here is that property assertions that assert and also return a function to assert messes a lot of stuff up. There were some terrible nasty bugs that came from this.

Aside from my concerns of implementation complexity, I also don't think this is actually internally consistent with the property-vs-assertion rule we have: `throw` takes arguments, and so having it as a property sometimes, and other times not, is something that does not have existing precedent.

Further to your point:

> I'm opposed to jumping to the opposite extreme of removing all property-based assertions from Chai; despite these types of problems, I think property-based assertions are a critical part of Chai's identity and appeal for many users.

I'm of the opposite mind set. For one, we see a lot of issues around implicitly passing tests because people often mistake function-assertion for a property-assertion, and get many bug reports about this. For another it makes the learning cliff a lot harder, as we're probably the _only_ library in the users tool-belt that has this behaviour. If it were a common idiom I'd be more behind it, but it is not. Another reason I'd like to see this pattern disappear is that it makes plugin development more complex, both from the API complexity but also from a philosophical perspective where not every plugin strictly adheres to the property-vs-method rubric, which of course in turn decreases usability because Chai's own rules aren't consistent.

Currently, with chai@5 there exists no way to handle property assertions, although I plan to add them in but via an optional interface, and that is forced based on something we can introspect in code (e.g. property assertions are made from any unary assertions - i.e one that only takes only an `actual` and not any additional arguments).

lib/chai/core/assertions.js
```diff
@@ -2448,135 +2448,138 @@ module.exports = function (chai, _) {
...
    *     expect(badFn).to.throw();
    *
-   * When one argument is provided, and it's an error constructor, `.throw`
-   * invokes the target function and asserts that an error is thrown that's an
-   * instance of that error constructor.
+   * When invoked with one argument, and it's a function (presumably a
+   * constructor), `.throw` invokes the target function and asserts that it
```
keithamus: Nit: `s/presumably/typically/`. `presumably` suggests we don't know what it will be, whereas `typically` suggests that is the common pattern.

lib/chai/core/assertions.js
```diff
@@ -2448,135 +2448,138 @@ module.exports = function (chai, _) {
...
+   *     var badFn = function () { throw TypeError('Illegal salmon!'); };
    *
-   *     expect(badFn).to.throw(/salmon/);
+   *     expect(badFn).to.throw(TypeError('Illegal salmon!'));
    *
-   * When two arguments are provided, and the first is an error instance or
-   * constructor, and the second is a string or regular expression, `.throw`
-   * invokes the function and asserts that an error is thrown that fulfills both
-   * conditions as described above.
+   * When invoked with two arguments, and the first is a function (presumably a
```
keithamus: Same nit: `s/presumably/typically/`

lib/chai/core/assertions.js
```diff
@@ -2448,135 +2448,138 @@ module.exports = function (chai, _) {
...
+   * the target function and asserts that it throws a string that contains or
+   * matches the second argument, rather than checking the `.message` property
+   * of the thrown value.
    *
-   *     var goodFn = function () {};
+   *     var badFn = function () { throw 'Illegal salmon!'; };
    *
-   *     expect(goodFn).to.not.throw();
+   *     expect(badFn).to.throw(String, 'salmon');
+   *     expect(badFn).to.throw(String, /salmon/);
```
keithamus: I like this 👍

lib/chai/core/assertions.js
```diff
@@ -2586,24 +2589,33 @@ module.exports = function (chai, _) {
...
+   * any type of thrown value.
+   *
+   *     function NonErrorConstructor(message) {
+   *       this.message = message;
+   *     }
+   *     var nonErrorObject = new NonErrorConstructor('Illegal salmon!');
+   *     var badFn = function () { throw nonErrorObject; };
+   *
+   *     expect(badFn).to.throw(NonErrorConstructor, 'salmon');
+   *     expect(badFn).to.throw(new NonErrorConstructor('Illegal salmon!'));
```
keithamus: If I am to understand the proposed algorithm, the only reason this does this is because the logic falls through to using `deep-eql`. We should mention that explicitly, if it is the case, as there's plenty of edge cases around `deep-eql` which might trip a user up when making this assertion.

lib/chai/core/assertions.js
meeber: My two main arguments against making property assertions "opt-in" in chai@5 are:

1. The history & popularity of Chai undermine the effectiveness of dropping property-based assertions this late in the game. Chai has had 7 years worth of blog posts and tutorials published on the Internet with assertions like `expect(blah).to.be.true`, so if a user unknowingly follows one of those outdated tutorials in chai@5, they now have an assertion that looks legit and has increased legitimacy from being a style that's demonstrated across many tutorials but now in chai@5 passes no matter what.
2. This type of bug report dropped drastically in chai@4 with the introduction of proxy protection. There's now only a couple of remaining assertions in Chai core that could be reasonably written such that they look legit but pass no matter what. `expect(fn).to.throw;` and `expect('meow').to.be.a.string;`. I'm not familiar with the implementation complexity and bugs that popped up when trying to fix this in chai@2; I guess I'd have to experience that pain myself to understand!

Good points about increased complexity for plugin authors, and consistency issues across both the Chai plugin ecosystem as well as JavaScript as a whole. I'm convinced that dropping property-based assertions would've been the right decision at the start of the project, but I'm not so sure it is now given Chai's history & popularity. That's a tough one.

meeber:
@chaijs/chai Here's what I have in mind for solving problems 1, 2, and 4 that I described in my original post, while simultaneously simplifying code. Other assertions (e.g., `rejectedWith` and the yet-to-be-created `.error`) could utilize the `createErrCriteria` and `assertErrCriteria` functions to simplify their logic as well.

This is still a draft, and I haven't updated the `should` or `assert` interface tests.

lib/chai/core/assertions.js
```diff
@@ -2447,130 +2447,214 @@ module.exports = function (chai, _) {
...
+    }
+
+    if (cri.errMsgMatcherType !== 'string' &&
+        cri.errMsgMatcherType !== 'regexp') {
+      throw TypeError('errMsgMatcher must be a string or regexp when' +
+                      ' defined');
+    }
+
+    return cri;
+  }
```
keithamus: I'd be interested to see these functions move into https://github.com/chaijs/check-error where they can be tested for their surface area, and can be reviewed+merged+shipped easier there, thereby making this PR smaller.

meeber: I thought about that. `createErrCriteria` could probably be moved to https://github.com/chaijs/check-error, but it'd be the only function there, without any real non-Chai applicability. `assertErrCriteria` is using a bunch of other Chai assertions internally, so I don't think it can be cleanly moved to a separate module until v5 when assertions are better modularized.

meeber:
I'm still not convinced this is the right implementation. Gonna close for now and revisit post v5 release.

PR #1224, opened 2019-01-15
Update Node versions in Travis config

gin:


meeber:
Thanks for submitting this. Despite Node v4 no longer being supported, and despite the noted date in our Travis config file having long ago passed by, removing support for Node v4 is still a breaking change which I don't think we need to perform until we release a new major version that contains other breaking changes.

PR #1229, opened 2019-01-22
feat: better error message for Map

### LINKED ISSUE #1228: Assertion error does not state problematic parts for Maps
```js
chai.expect(new Map([['a','a']])).to.deep.equal(new Map([['a','b']]))
// AssertionError: expected {} to deeply equal {}
```
As you can see it is not clear why it fails from the error message.

pcorpet:
Closes #1228

### DIFF
--- a/lib/chai/utils/inspect.js
+++ b/lib/chai/utils/inspect.js
@@ -131,6 +131,7 @@ function formatValue(ctx, value, recurseTimes) {
   var base = ''
     , array = false
     , typedArray = false
+    , map = false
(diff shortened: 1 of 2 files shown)


keithamus:
Thanks for this @pcorpet but we're very close to releasing a new version of the inspection engine: https://github.com/chaijs/loupe/pull/14. 

Once we get `loupe@2` out we'll be updating Chai to use it, and it will replace all `inspect` code.

We really appreciate the work you've done here! I hope in closing this PR we don't dissuade you from making further contributions!

PR #1242, opened 2019-03-12
(feat): expect value to contain oneOf

voliva:
Previously, doing an assertion such as

```js
expect('The quick brown fox jumps over the lazy dog').to.contain.oneOf(['cat', 'dog', 'bird']);
```

Would reject with the message `expected [ ... ] to be one of ['cat', 'dog', 'bird']`, as `oneOf` doesn't check for `contains` flag.

This PR adds this additional check to `oneOf`, so now it can be chained with `contain`, `contains`, `include` and `includes`.

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -3117,6 +3117,14 @@ module.exports = function (chai, _) {
    *     expect(1).to.equal(1); // Recommended
    *     expect(1).to.not.be.oneOf([2, 3, 4]); // Not recommended
    *
+   * It can also be chained with `.contain` or `.include`, which will work with
+   * both arrays and strings:
+   * 
+   *     expect('Today is sunny').to.contain.oneOf(['sunny', 'cloudy'])
+   *     expect('Today is rainy').to.not.contain.oneOf(['sunny', 'cloudy'])
+   *     expect([1,2,3]).to.contain.oneOf([3,4,5])
+   *     expect([1,2,3]).to.not.contain.oneOf([4,5,6])
+   * 
    * `.oneOf` accepts an optional `msg` argument which is a custom error message
    * to show when the assertion fails. The message can also be given as the
    * second argument to `expect`.
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -3135,16 +3135,27 @@ module.exports = function (chai, _) {
     if (msg) flag(this, 'message', msg);
     var expected = flag(this, 'object')
       , flagMsg = flag(this, 'message')
-      , ssfi = flag(this, 'ssfi');
+      , ssfi = flag(this, 'ssfi')
+      , contains = flag(this, 'contains');
     new Assertion(list, flagMsg, ssfi, true).to.be.an('array');
 
-    this.assert(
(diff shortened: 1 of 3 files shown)


chai.js
```diff
@@ -3496,16 +3496,27 @@ module.exports = function (chai, _) {
     if (msg) flag(this, 'message', msg);
```
lucasfcosta: I think the docs above need to specify that this assertion supports the `contains` flag since it ultimately changes its behaviour. It would also be good to add an example.

lucasfcosta:
Hi @voliva, this is just a friendly ping to see if you need help with this PR.

I'd love to get it merged in. Thanks again for the work 😊

voliva:
I totally forgot 😱

I'll take a look today, thanks for the reminder!

voliva:
Does that look right? @lucasfcosta

lib/chai/core/assertions.js
```diff
@@ -3135,16 +3143,27 @@ module.exports = function (chai, _) {
...
       , flagMsg = flag(this, 'message')
-      , ssfi = flag(this, 'ssfi');
+      , ssfi = flag(this, 'ssfi')
+      , contains = flag(this, 'contains');
     new Assertion(list, flagMsg, ssfi, true).to.be.an('array');
 
-    this.assert(
+    if(contains) {
+      this.assert(
+        list.some(posibility => expected.indexOf(posibility) > -1)
```
lucasfcosta: There seems to be a typo here. It should be `possibility` instead of `posibility`.

voliva: Good catch, I've just amended the relevant commit.

chai.js
```diff
@@ -3496,16 +3496,27 @@ module.exports = function (chai, _) {
...
     var expected = flag(this, 'object')
       , flagMsg = flag(this, 'message')
-      , ssfi = flag(this, 'ssfi');
+      , ssfi = flag(this, 'ssfi')
+      , contains = flag(this, 'contains');
     new Assertion(list, flagMsg, ssfi, true).to.be.an('array');
 
-    this.assert(
+    if(contains) {
```
vieiralucas: Sorry about this nitpicking but can you add a space between the `if` and the `(contains) {`?
`if (contains) {`
Just to keep style consistency.

lib/chai/core/assertions.js
```diff
@@ -3135,16 +3143,27 @@ module.exports = function (chai, _) {
     if (msg) flag(this, 'message', msg);
     var expected = flag(this, 'object')
       , flagMsg = flag(this, 'message')
-      , ssfi = flag(this, 'ssfi');
+      , ssfi = flag(this, 'ssfi')
+      , contains = flag(this, 'contains');
     new Assertion(list, flagMsg, ssfi, true).to.be.an('array');
 
-    this.assert(
+    if(contains) {
```
vieiralucas: Sorry about this nitpicking but can you add a space between the `if` and the `(contains) {`?
`if (contains) {`
Just to keep style consistency.

voliva:
Hey @vieiralucas thanks for your feedback! I've addressed your comments

PR #1249, opened 2019-04-23
Lerna

keithamus:
I feel like we're in a good position with `new-chai` to start porting stuff over. I don't want to move everything over in one huge commit, I'd rather do a smaller set of commits so people can more easily see whats happening.

This PR just moves chai over to `lerna` so we can have multiple packages in one repo. There will still be one canonical `chai` package - but other packages around it will be developed in an isolated manner, which allows us to move more quickly with various smaller elements.

This is also an opportunity for us to openly discuss if we think this is a good idea, so /cc @meeber @lucasfcosta @shvaikalesh @vieiralucas


meeber:
Unfortunately I'm in the middle of yet another cross-state move, combined with being super busy at work. It feels like it's been a year or more since I've had any time for Chai or even JavaScript in general. I'm not sure when I'll have a chance to look at this. Please don't let my lack of response be a blocker to progress.

lucasfcosta:
In regards to this, I was wondering whether we should get this in before or after the outstanding PRs.

I'd love to get it in but maybe we could get the other pieces of code in so that the show doesn't stop 😄

Btw, again, I really like this idea, thanks @keithamus 💖

Issue #1252, opened 2019-05-15
Chai Assertion error doesn't have 'operator' attribute


rpgeeganage:
Hey everyone,
I'm currently working on a feature in `jest` repo (https://github.com/facebook/jest/pull/8454).
While I came across that output is as follows. (https://github.com/facebook/jest/pull/8454/files#diff-e6f1eef106ffd32cb2e65acae35856b8R7)
```    
assert.(received, expected)

    Expected value   "hello sunshine"
    Received:
      "hello world"

    Message:
      expected 'hello world' to equal 'hello sunshine'

    Difference:

    - Expected
    + Received

    - hello sunshine
    + hello world

      11 | describe('chai.js assertion library test', () => {
      12 |   it('expect', () => {
    > 13 |     chai.expect('hello world').to.equal('hello sunshine');
         |                                   ^
      14 |   });
      15 | 
      16 |   it('should', () => {

      at Object.equal (__tests__/chai_assertion.js:13:35)
```
it is because of https://github.com/facebook/jest/blob/master/packages/jest-circus/src/formatNodeAssertErrors.ts#L127, unable to extract the operator.
In https://nodejs.org/api/assert.html#assert_new_assert_assertionerror_options shows a property of `operator`.

is it this a bug or have I taken a wrong approach?
Thanks a lot.

keithamus:
Hey @rpgeeganage thanks for the issue!

We do not use the `operator` property at all, to my knowledge. I'd happily welcome a PR to add it if you feel so inclined 😄

rpgeeganage:
@keithamus ,
Thank you very much for the clarification.
I would love to spend time on adding this feature to `chai`. I will look into this and try to open a PR.
Thanks again for a wonderful library.

astorije:
Hey @rpgeeganage, thank you so much for helping!

I think I am misunderstanding something though so let me ask: what is the issue with the output you posted above? Is there something missing or incorrect?
I haven't met the `operator` attribute before, so I'm trying to understand what it's used for and what I'm missing :)

Thanks!

SimenB:
`>assert.(received, expected)` at the top, it should say `assert.equal(received, expected)` or some such

rpgeeganage:
@astorije ,
[...]

PR #1253, opened 2019-05-19
Adding operator attribute to assertion error

rpgeeganage:
This is an attempt to add `operator` property for the `AssertionError`.
(As per the issue https://github.com/chaijs/chai/issues/1252)
Operators are as follows.
- https://nodejs.org/api/assert.html#assert_assert_deepstrictequal_actual_expected_message
- https://nodejs.org/api/assert.html#assert_assert_notdeepstrictequal_actual_expected_message
- https://nodejs.org/api/assert.html#assert_assert_strictequal_actual_expected_message
- https://nodejs.org/api/assert.html#assert_assert_notstrictequal_actual_expected_message

Please give me feedback on this draft PR.
Thank you.

### DIFF
--- a/lib/chai/assertion.js
+++ b/lib/chai/assertion.js
@@ -138,11 +138,20 @@ module.exports = function (_chai, util) {
     if (!ok) {
       msg = util.getMessage(this, arguments);
       var actual = util.getActual(this, arguments);
-      throw new AssertionError(msg, {
-          actual: actual
+      var assertionErrorObjectProperties = {
+        actual: actual
         , expected: expected
         , showDiff: showDiff
-      }, (config.includeStack) ? this.assert : flag(this, 'ssfi'));
+      };
+
+      var operator = util.getOperator(this, arguments);
+      if (operator) {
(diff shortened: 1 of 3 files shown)


rpgeeganage:
@keithamus ,
Can you please take a peek and give feedback on this when you have time.
(I haven't added any tests yet, I'll add tests when reviewers are fine with the PR).

keithamus:
The only normative specification of `operator` is that it is a string right? While I'm fine to default most instances to one of `[not][deep][strict]Equal` I also think we should provide a mechanism for allowing the operator to be overridden in the assertion.

To me it makes most sense to use `flag(this, 'operator')` and simply have it default to using `getOperator` if not present. This way if we find mismatches they're easy to fix by just setting the flag.

rpgeeganage:
@keithamus,
Thanks a lot. I'll do the changes.

rpgeeganage:
I'm very sorry that,
I have screwed up the branch. so I'm closing this PR and will open a new PR.

rpgeeganage:
https://github.com/chaijs/chai/pull/1257

PR #1260, opened 2019-05-26
Greenkeeper/karma 4.1.0

vieiralucas:

PR #1259, opened 2019-05-26
Greenkeeper/mocha 6.1.4

vieiralucas:

Issue #860, opened 2016-11-03
Raise IE bar to 11.


jdalton:
Since IE 11 is the [last version](https://www.microsoft.com/en-us/WindowsForBusiness/End-of-IE-support) to continue to receive security updates, compatibility fixes, and technical support on Windows 7, 8.1, and 10 it may be a good time with v4 to raise the bar to IE 11 (previously IE 10).

meeber:
It looks like we have a bit of a mix-up.

Our recently agreed upon stance is that we'd support whatever browsers are still officially under support by their respective vendors. ([source](https://github.com/chaijs/chai/issues/124#issuecomment-244967570))

Microsoft's stance as of early this year is that they'll only support the latest version of IE that's available for each Windows version that's still supported. ([source](https://www.microsoft.com/en-us/WindowsForBusiness/End-of-IE-support))

Windows Vista SP2 is under extended support until 2017-04-11 ([source](https://support.microsoft.com/en-us/help/13853/windows-lifecycle-fact-sheet)), and the latest version of IE that's available on it is IE9. 

Windows Server 2012 (non-R2) is under normal support until 2018-01-09, and under extended support until 2023-01-10 ([source](https://support.microsoft.com/en-us/lifecycle/search?sort=PN&alpha=windows%20server&Filter=FilterNO)), and the latest version of IE that's available on it is IE10.

Therefore, it seems that we either need to continue supporting IE9 through early next year, and IE10 for another 6 years, or we need to modify our stance in regard to IE.

If we continue supporting IE9 through early next year, then we have an immediate problem: IE9 isn't currently tested via Sauce (intentional? unintentional?), and I'm sure the changes made since 3.5.0 have broken support.

Pinging @keithamus @lucasfcosta @vieiralucas @shvaikalesh

jdalton:
Hi, I work on the Edge team at Microsoft. The usage of Windows Vista SP2 is so low that while the technical nit is there our outward messaging and guidance is to treat IE9 as cut to avoid that weird support IE9 but skip IE10 situation.

meeber:
@jdalton What is the team's guidance on IE10 due to Server 2012 non-R2?

jdalton:
IE10 usage is so low it's not even relevant.

jdalton:
IE10 is like IE7. No one uses it.

meeber:
[...]

PR #1261, opened 2019-05-26
chore: drop support for node v4 and v6

vieiralucas:
This should enable us to merge #1259 and #1260

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,8 @@ addons:
   sauce_connect: true
 node_js:
  - 8 # to be removed 2019-12-31
+ - 10 # to be removed 2021-04-31
+ - 12 # to be removed 2022-04-31
  - lts/* # safety net; don't remove
  - node # safety net; don't remove
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,7 @@ language: node_js
 addons:
   sauce_connect: true
 node_js:
- - 4 # to be removed 2018-04-30
- - 6 # to be removed 2019-04-01
  - 8 # to be removed 2019-12-31
- - 9 # to be removed 2018-06-30
  - lts/* # safety net; don't remove
  - node # safety net; don't remove
 
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "test": "make test"
   },
   "engines": {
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -3,8 +3,6 @@ language: node_js
 addons:
   sauce_connect: true
 node_js:
- - 4 # to be removed 2018-04-30
- - 6 # to be removed 2019-04-01
  - 8 # to be removed 2019-12-31
  - 9 # to be removed 2018-06-30
```
astorije: Probably worth removing v9 as well (end of life since 2018-06-30), and add v10 and v12 to the list, no?

vieiralucas:
@lucasfcosta v9 is actually over, expired in 2018.
I added v10 and v12 as @austince suggested. Did not added v11 since it will expire in 3 days from now 😄

austince:
Believe you meant to tag @astorije 😄

astorije:
Thank you for this!

vieiralucas:
I'll try to fix the `codecov` issue. I believe that we were relying on the fact that we used to run some code in node 4 for coverage (Checking the existance of Proxy).
Just need to add another test.

vieiralucas:
Now I'm very confused on why this is failing :(

astorije:
I can't explain it either but according to [the logs](https://travis-ci.org/chaijs/chai/jobs/538333358#L3601), it's a syntax error coming from [this line](https://github.com/chaijs/chai/blob/master/test/utilities.js#L599). I don't see anything wrong with it though 😕

vieiralucas:
Hey @lucasfcosta and @astorije 
Just fixed the build, problem was that IE 10 didn't support arrow functions (obviously right? 😄 )

I need @chaijs/chai approval again so we can merge this.

vieiralucas:
Also, I just remembered that we did agree with raising IE bar to 11 (see this https://github.com/chaijs/chai/issues/860#issuecomment-395977926).

We just need to move forward with https://github.com/chaijs/guidelines/pull/3

ljharb:
Heads up that this is a breaking change in a minor version - chai v4.0.0 supported node 4, so all versions of chai 4.x need to continue supporting it.

PR #1277, opened 2019-07-15
fix: ISSUE 1276: assert.doesNotDecreaseBy not work as expected

benben77:
fix the issue that `assert.doesNotDecreaseBy` does not works as expected / as described in docs

### DIFF
--- a/lib/chai/interface/assert.js
+++ b/lib/chai/interface/assert.js
@@ -2853,7 +2853,7 @@ module.exports = function (chai, util) {
     }
 
     return new Assertion(fn, msg, assert.doesNotDecreaseBy, true)
-      .to.not.decrease(obj, prop).by(delta);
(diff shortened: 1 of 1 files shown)


rpgeeganage:
Hi,
Nice work!
But I'm wondering, why any of the tests are not breaking?
I guess it is better to have tests for this.

benben77:
> Hi,
> Nice work!
> But I'm wondering, why any of the tests are not breaking?
> I guess it is better to have tests for this.

Oh you are right, it's weird that `doesNotDecreaseBy` is not covered in tests and there's no `doesNotIncreaseBy` 😂

is2ei:
I think it doesn't need `doesNotDecreaseBy` since it has`decreasesButNotBy`.

keithamus:
Can we please add some tests to assert this behaviour?

PR #1306, opened 2019-11-10
Node version >=8 in README

flaambe:


brutalcrozt:
Nice catch as the readme should same as [here](https://github.com/chaijs/chai/blob/master/package.json#L33). It also promotes wider user to use >=8

flaambe:
Will you merged it? Or support for node v8 will be dropped soon? @keithamus @lucasfcosta

brutalcrozt:
skip v8 ? :thinking:

PR #1317, opened 2020-01-28
Export chai as ES module

sebamarynissen:
Since Node 13.2 `--experimental-modules` was unflagged. Since then mocha [has enabled support](https://github.com/mochajs/mocha/pull/4038) for tests written as an ES module in [v7.0.0-esm1](https://github.com/mochajs/mocha/releases/tag/v7.0.0-esm1).

This means that from now on it is possible to write your tests as 
```js
import chai from 'chai';

describe('foo', function() {
  it('bar', function() {
    chai.expect(true).to.be.true;
  });
});
```

Unfortunately it is not possible to write
```js
import { expect } from 'chai';
```
because Node doesn't support named exports from CommonJS modules.

This PR changes this by exporting chai as an ES module and by making use of conditional exports, which was unflagged - but is still experimental - in Node 13.7.

I am aware that conditional exports is still experimental, and hence you'll still get a warning, both for
```js
import { expect } from 'chai';
const { expect } = require('chai');
```
I therefore propose to release the conditional exports as an experimental version - similar to mocha `v7.0.0-esm1` - but to *include* the `chai.mjs` file in a semver minor so that people can do
```js
import { expect } from 'chai/lib/chai.mjs';
```
if they want.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
   "main": "./index",
   "exports": {
     "import": "./lib/chai.mjs",
-    "default": "./index.js",
+    "default": "./index.js"
   },
   "scripts": {
     "test": "make test"
--- a/package.json
+++ b/package.json
@@ -27,8 +27,8 @@
   },
   "main": "./index",
   "exports": {
-    "require": "./index.js",
-    "import": "./lib/chai.mjs"
+    "import": "./lib/chai.mjs",
+    "default": "./index.js",
   },
   "scripts": {
     "test": "make test"
--- a/package.json
+++ b/package.json
@@ -26,6 +26,10 @@
     "url": "https://github.com/chaijs/chai/issues"
   },
   "main": "./index",
+  "exports": {
+    "require": "./index.js",
+    "import": "./lib/chai.mjs"
+  },
   "scripts": {
     "test": "make test"
   },
--- a/lib/chai.mjs
+++ b/lib/chai.mjs
@@ -1,0 +1,30 @@
+import chai from './chai.js';
+
+// Destructure everything we want to export from chai.
+const {
+	version,
+	AssertionError,
+	use,
+	util,
+	config,
+	Assertion,
+	expect,
+	should,
+	Should,
+	assert,
+} = chai;
+
+// Re-export.
+export {
+	chai as default,
+	version,
+	AssertionError,
+	use,
+	util,
+	config,
+	Assertion,
+	expect,
+	should,
+	Should,
+	assert,
+};


sebamarynissen:
I think it should indeed be considered as an interim measure until chai has been fully migrated to ES modules - which I suppose will eventually be done once ES modules are no longer experimental in Node.

MylesBorins:
This is exactly the pattern we recommend.

https://nodejs.org/dist/latest-v13.x/docs/api/esm.html#esm_approach_1_use_an_es_module_wrapper

package.json
```diff
@@ -26,6 +26,10 @@
     "url": "https://github.com/chaijs/chai/issues"
   },
   "main": "./index",
+  "exports": {
+    "require": "./index.js",
```
MylesBorins: if you rename this `default` and put it below `import` then there will be no warning when requiring the package, only on import

sebamarynissen: Thanks, I didn't know that. I will push a new version tomorrow. Is this documented somewhere?

MylesBorins: Honestly it isn't. We only discovered that this was an issue for consumers of modules due to threads like this... and only realized this pattern worked live in the meeting on wednesday while we debating how to handle it 😅

I'm going to not have time to get any copy together in the near future but if you have time and want to send a PR I'd be happy to review and help get it landed.

sebamarynissen: Do you mean a PR for chai or for the Node documentation or ... ? Don't really know what you mean here, because I think this PR for chai is pretty much done.

MylesBorins: I meant to node documentation, sorry for not being clearer. Also not at all intending this as "PRs welcome"... we'll definitely follow up with this. Just wanted to offer to help you if you had time + interest.

chrisveness:
Thanks for this work – it would  be good to see it merged & released.

brettz9:
I would expect this could also potentially allow browsers to use chai via ESM without transpiling (using a direct `node_modules` path or an import map)--if necessary using "browser" or "browser-esm" within the conditional exports (the exact naming hasn't been standardized).

giltayar:
DO NOT MERGE THIS PR, as you will disable deep-linking into chai. See https://github.com/chaijs/chai/pull/1340#issuecomment-634862587 for an explanation.

sebamarynissen:
I'm closing this following #1340. I verified that indeed it does not work if you want to deep import chai:
```js
const chai = require('chai/lib/chai');
```

@MylesBorins: Is this something that the modules team is aware of? I fear that a lot of modules will suffer from this problem when trying to migrate to ESM with conditional exports.

GeoffreyBooth:
> Is this something that the modules team is aware of? I fear that a lot of modules will suffer from this problem when trying to migrate to ESM with conditional exports.

Yes, this is by design. With `"exports"` you need to explicitly define what deep imports you want to be exported: https://nodejs.org/api/esm.html#esm_package_entry_points:

> Warning: Introducing the "exports" field prevents consumers of a package from using any entry points that are not defined, including the package.json (e.g. `require('your-package/package.json')`. **This will likely be a breaking change.**

The docs go into greater detail, but basically you're encouraged to define what the package's actual supported entry points are; or if you truly want to allow access to any file, you can add `"./": "./"`.

sebamarynissen:
Allright, so if I understand correctly the best way to go forward with this is to only publish it in a semver major, perhaps with some explicit additional entry points. Correct?

MylesBorins:
@sebamarynissen if you included `"./": "./"` mapping the root to the root it will be a semver-minor change

package.json
```diff
@@ -26,6 +26,10 @@
     "url": "https://github.com/chaijs/chai/issues"
   },
   "main": "./index",
+  "exports": {
+    "import": "./lib/chai.mjs",
+    "default": "./index.js"
+  },
```
MylesBorins, line 32: "exports": {
    ".": {
      "import": "./lib/chai.mjs",
      "default": "./index.js"
    },
    "./": "./"
  },

sebamarynissen:
Ok, but perhaps that we should go with #1340 then as it adds the same functionality, handles the deep imports problem *and* contains tests for it.

PR #1322, opened 2020-03-11
Add non-nullable modifier to return type of functions never returning null

vrana:


keithamus:
Could you please explain the motivation for this change?

vrana:
In Closure type system, `Array` actually means `Array` or `null`, it's equivalent to `?Array`. This removes the `null` from the return type declaration because the function never returns null. It means that you will be able to assign to a variable declared as `!Array`.

keithamus:
Are you aware if these changes have ramifications to other consumers of JSDoc, such as TypeScript? Is this valid JSDoc syntax?

vrana:
Yes, this is valid JSDoc syntax. https://jsdoc.app/tags-type.html

I believe TS would consume this just fine as it is used in many other projects.

keithamus (approved):

PR #1324, opened 2020-03-19
Add `also` language chain

sportshead:
This would be helpful, as the property doesn't clash with other methods.


keithamus:
Thanks for the PR @sportshead! I'm not sure how this will clash with other plugins and the like, so we might have to hold off on merging this just yet before doing more research.

keithamus (approved):
Agreed. Let's add it!

PR #1329, opened 2020-04-03
docs: Fix simple typo, adher -> adhere

timgates42:
There is a small typo in chai.js.

Should read `adhere` rather than `adher`.


keithamus:
Hey @timgates42 thanks for the PR!

This is actually a compiled file which is replaced with each build, the comments you're editing belong to https://github.com/chaijs/type-detect, so you might want to raise the PR there.

I'll close this because there's no point in merging - it'll only be blown away with the next release.

PR #1334, opened 2020-04-18
Deep equality check in oneOf

### LINKED ISSUE #1241: Obj is oneOf array of objects
Hi ! :)

I would to test if an object if one of entries from an array containing objects.
It's something like: `expect({}).to.deep.oneOf([{}, {}, {}])`

But I got this error: `AssertionError: expected { Object (type, sender, ...) } to be one of [ Array(3) ]`

With more details, this is the test with the object and the array:

```
expect({
  "type": "text",
  "sender": "bot",
  "user": "USER_TEST",
  "payload": {
    "value": "Lorem ipsum"
  }
}).to.deep.oneOf([
  {
    "type": "text",
    "sender": "bot",
    "user": "USER_TEST",
    "payload": {
      "value": "Dolor sit amet"
    }
  },
  {
    "type": "text",
    "sender": "bot",
    "user": "USER_TEST",
    "payload": {
      "value": "Lorem ipsum"
    }
  },
  {
    "type": "text",
    "sender": "bot",
    "user": "USER_TEST",
    "payload": {
      "value": "Consectetur adipiscing elit"
    }
  }
])
```

How can I do this kind of test ?

rgroothuijsen:
In issue #1241 it was noted that `.oneOf` gives an unexpected result when attempting deep comparisons, and it turns out that `.oneOf` does not check for the `deep` flag. Since this is already an existing option for `.equal`, it would make sense to extend this behavior to its counterpart in `.oneOf`. 

Fixes #1241.

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -3150,7 +3150,8 @@ module.exports = function (chai, _) {
     var expected = flag(this, 'object')
       , flagMsg = flag(this, 'message')
       , ssfi = flag(this, 'ssfi')
-      , contains = flag(this, 'contains');
+      , contains = flag(this, 'contains')
+      , isDeep = flag(this, 'deep');
     new Assertion(list, flagMsg, ssfi, true).to.be.an('array');
 
     if (contains) {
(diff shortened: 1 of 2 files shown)


keithamus (approved):
This looks like a useful change! Thanks!

Issue #1321, opened 2020-02-25
Support for BigInt


fubar:
Chai still doesn't support BigInts.

```js
expect(BigInt(1)).to.equal(BigInt(2));
```

Results in `TypeError: Do not know how to serialize a BigInt` (triggered by JSON.stringify).

BigInt is now a standard primitive and it would be great if Chai supported it out of the box.

Stale/closed issue: https://github.com/chaijs/chai/issues/1195. The roadmap has that issue in todo but it has not been updated since Apr 2019: https://github.com/chaijs/chai/projects/2

dwelle:
Serialization can be solved by doing something like:

```
BigInt.prototype.toJSON = val => `${val}n`;
```

But assertions will also fail because chai boxes bigints into some kind of object (when storing it to `__flags.object`.

keithamus:
I'm going to close this, as it is a dupe of #1195. The roadmap is the preferred way to track progress for us. The current in flight project is `loupe` - which was deemed in most need of attention, and the last commit was 22nd Apr 2020. Once we're happy with loupe we'll be moving onto the next todo item.

If you uncover any bugs please feel free to file more issues, but rest assured we're aware of a lack of support for ES6+ types and are working hard to fix that.

dwelle:
@keithamus if it's not too much trouble, can you give me a hint where chai boxes BigInts (and I reckon any types it doesn't know about) into `[BigInt: 122054031626471353n]` when I log `utils.flag(this, 'object')`? I'm getting lost in the source, and could use some help for a fork before chai v2 is ready. Thanks!

keithamus:
Just FYI we support BigInts for assertions like `.to.be.equal`. If you run `expect(0n).to.equal(0n)` you'll get a passing assertion. I don't know where the OP saw the error `TypeError: Do not know how to serialize a BigInt` (perhaps from a plugin or some kind of monkey-patch on `BigInt`?), but if I run `expect(0n).to.equal(1n)` I get `AssertionError: expected {} to equal {}` 

Chai doesn't box the values in any way. Expect just wraps the given argument in a new `Assertion` type: https://github.com/chaijs/chai/blob/e54d834be75430fce2e624063ca7abde9bf6bcfc/lib/chai/interface/expect.js#L7-L9

[...]

PR #1349, opened 2020-06-17
unbox BigInt primitives in shouldGetter

dwelle:
Partially addresses https://github.com/chaijs/chai/issues/1321 so that global `should` getter unboxes primitive BigInt values.

Lemme know if I should write any tests.

### DIFF
--- a/lib/chai/interface/should.js
+++ b/lib/chai/interface/should.js
@@ -13,7 +13,8 @@ module.exports = function (chai, util) {
       if (this instanceof String
           || this instanceof Number
           || this instanceof Boolean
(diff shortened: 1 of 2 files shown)


keithamus:
Thanks for doing this @dwelle :+1:

Please could you write a couple of tests just so we don't end up regressing this behaviour? Thanks!

dwelle:
I've written tests for the unboxing behavior, and when I was at it wrote for other types as well. Hopefully I didn't duplicate it from somewhere else. Used a custom assertion -- not sure if I should have use chai assertions for this.

PR #1351, opened 2020-06-29
Use object.constructor.name for objDisplay.

tomalec:
To make failure messages more descriptive, like:
```log
 AssertionError: expected { MyComponent (foo, bar) } ...
```
instead of 
```log
 AssertionError: expected { Object (foo, bar) } ...
```

I was not sure where to add a specific test for this change. `objDisplay` does not have a dedicated test file. Could you give me any hint?

### DIFF
--- a/lib/chai/utils/objDisplay.js
+++ b/lib/chai/utils/objDisplay.js
@@ -40,7 +40,7 @@ module.exports = function objDisplay(obj) {
         , kstr = keys.length > 2
           ? keys.splice(0, 2).join(', ') + ', ...'
           : keys.join(', ');
-      return '{ Object (' + kstr + ') }';
+        return '{ ' + (obj.constructor.name || 'Object') + ' (' + kstr + ') }';
     } else {
       return str;
     }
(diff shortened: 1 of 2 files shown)


keithamus:
This looks like a great change, but we're working on integrating [`loupe`](https://github.com/chaijs/loupe) into Chai 4.4.0 which will handle this behaviour. 

I'm going to close this PR as it'll only be overwritten when we integrate loupe. We really value your contribution here though - and it would be amazing if you could add some tests to loupe to verify class names work as you'd expect! Would you be able to do that please?

Thanks!

tomalec:
Do I get it right, that the plan is to replace current [`objDisplay`](https://github.com/tomalec/chai/blob/master/lib/chai/utils/objDisplay.js) with `loupe` in chaijs? Or would I have to integrate it into my project by myself?

I didn't run it against the case that was the source of the original issue, but the tests at https://github.com/chaijs/loupe/blob/master/test/class.js#L4-L22 seem to cover my scenario.

keithamus:
Yeah Chai's `inspect` function internally will be replaced with `loupe`. You'll have to do nothing with your code to integrate it. `loupe` a project is just in a better place to make improvements with.

If `loupe` currently works as you hope it does, then there's nothing to do! We'll get it integrated in Chai 4.4.0 and when it's released this bug will be fixed :tada:

PR #1355, opened 2020-07-17
Fix JSDoc identation

DanielVenturini:
This indentation will fix the code area in [api/assert](https://www.chaijs.com/api/assert/#method_notsamedeeporderedmembers) for the method `.sameDeepOrderedMembers` and `.notSameDeepOrderedMembers`.


keithamus (approved):
:+1: thanks for this!

PR #1365, opened 2020-10-19
fix: support inspecting bigints (#1321)

vapier:


vapier:
running `npm test` seems to have regenerated `chai.js` with changes before mine.  not sure if i should include them all here.

keithamus:
This looks like a great change, but we're working on integrating [`loupe`](https://github.com/chaijs/loupe) into Chai 4.4.0 which will handle this behaviour. 

I'm going to close this PR as it'll only be overwritten when we integrate loupe. We really value your contribution here though - and it would be amazing if you could add some tests to loupe to verify BigInt works! Would you be able to do that please?

Thanks!

PR #1380, opened 2021-03-03
Fixed a regression that caused SyntaxErrors on IE 11

### LINKED ISSUE #1379: Chai 4.3.x or higher doesn't run on IE 11
Chai 4.3 introduced a regression via chaijs/chai#1334 that causes chai to throw a syntax error on IE 11. This is because it's using an arrow function [here](https://github.com/chaijs/chai/blob/main/lib/chai/core/assertions.js#L3181) which isn't supported on IE 11.

image

This is also causing the automated saucelab tests setup for Chai to fail: https://app.saucelabs.com/tests/2e7d344d051d40a9b1524d6df4b86a1b

lnewson:
The changes made in #1334 used an arrow function and as this isn't supported on IE 11 it causes a SyntaxError to be thrown when loading chai. This just corrects that to use the older function syntax instead so it'll still be able to run on IE 11.

Fixes #1379


keithamus (approved):

keithamus:
Merged in 4.x.x branch

lnewson:
Ahh, sorry I missed that 4.x had moved off to a side branch. Thanks for getting that fixed 👍

PR #1401, opened 2021-06-07
feat: use chaijs/loupe for inspection

### LINKED ISSUE #1228: Assertion error does not state problematic parts for Maps

pcorpet:
Fix #1228

Note that for now this is just a quick attempt to explore which diffs are OK in the error messages and which ones we'd like to fix in `loupe` before merging.


pcorpet:
```
+ expected - actual

-expected +0 to be truthy
+expected 0 to be truthy
```
I think we'd like https://github.com/chaijs/loupe to display `0` instead of `+0`. Or do you feel this is a good change?

pcorpet:
```
      + expected - actual

      -blah: expected Foo{} to not be an instance of Foo
      +blah: expected {} to not be an instance of Foo
```
I think this one is a good change, please comment if you disagree

keithamus:
In both of these instances the changes are welcome I think.

pcorpet:
Alright, thanks for the confirmation. For the next ones, I suggest you just add an emoji on the github comment to say that you're OK with the change or add a new comment if you're not.

pcorpet:
Using ISO format for printing dates:
```
      + expected - actual
      -blah: expected 2021-06-09T11:40:15.786Z to be above 2021-06-09T11:40:16.786Z
      +blah: expected Wed, 09 Jun 2021 11:40:15 GMT to be above Wed, 09 Jun 2021 11:40:16 GMT
```

pcorpet:
Unquoted complex key in objects:
```
      + expected - actual

      -expected { foo.bar: 'baz' } to have nested property 'foo.bar'
      +expected { 'foo.bar': 'baz' } to have nested property 'foo.bar'
```

pcorpet:
We'll need to fix https://github.com/chaijs/loupe/issues/44 as well

pcorpet:
A bit more complex: when using prototype properties,

```js
Object.create({tea: 'chai'})
```

We use to show it as regular properties:
```
{tea: 'chai'}
```
We now exclude them which makes it less readable:
```
[object Object]
```

What do you think?

flaambe:
> Using ISO format for printing dates:
> 
> ```
>       + expected - actual
>       -blah: expected 2021-06-09T11:40:15.786Z to be above 2021-06-09T11:40:16.786Z
>       +blah: expected Wed, 09 Jun 2021 11:40:15 GMT to be above Wed, 09 Jun 2021 11:40:16 GMT
> ```

https://github.com/chaijs/loupe/issues/34

pcorpet:
Typed array:
```
      + expected - actual

      -Uint8Array[ 1, 2, 3 ]
      +[ 1, 2, 3 ]
```

keithamus:
> Unquoted complex key in objects:
> 
> ```
> 
>       + expected - actual
> 
> 
> 
>       -expected { foo.bar: 'baz' } to have nested property 'foo.bar'
> 
>       +expected { 'foo.bar': 'baz' } to have nested property 'foo.bar'
> 
> ```

I think this is a regression. Ideally loupe would give us inspections which are syntactically valid and can be pasted into a REPL with minimal editing. I think loupe should detect if it needs quotes and add them.

keithamus:
> A bit more complex: when using prototype properties,
> 
> 
> 
> ```js
> 
> Object.create({tea: 'chai'})
> 
> ```
> 
> 
> 
> We use to show it as regular properties:
> 
> ```
> 
> {tea: 'chai'}
> 
> ```
> 
> We now exclude them which makes it less readable:
> 
> ```
> 
> [object Object]
> 
> ```
> 
> 
> 
> What do you think?

This looks like a regression. I'd expect `{}` considering it's a prototype property, but we could also special case empty objects to show more info if needed.

keithamus:
> Typed array:
> 
> ```
> 
>       + expected - actual
> 
> 
> 
>       -Uint8Array[ 1, 2, 3 ]
> 
>       +[ 1, 2, 3 ]
> 
> ```

This is intentional. It is useful to know what typed array class is being inspected. We should update chai here.

pcorpet:
Another one:
```js
  it('inspect with custom stylize-calling inspect()s', function () {
    chai.use(function (_chai, _) {
      var obj = {
        outer: {
          inspect: function (depth, options) {
            return options.stylize('Object content', 'string');
          }
        }
      };
      expect(_.inspect(obj)).to.equal('{ outer: Object content }');
    });
  });
```

```
      + expected - actual

      -{ outer: 'Object content' }
      +{ outer: Object content }
```

keithamus:
> Another one:
> 
> ```js
>   it('inspect with custom stylize-calling inspect()s', function () {
>     chai.use(function (_chai, _) {
>       var obj = {
>         outer: {
>           inspect: function (depth, options) {
>             return options.stylize('Object content', 'string');
>           }
>         }
>       };
>       expect(_.inspect(obj)).to.equal('{ outer: Object content }');
>     });
>   });
> ```
> 
>  ```
>       + expected - actual
> 
>       -{ outer: 'Object content' }
>       +{ outer: Object content }
> ```

What does nodejs do here? To me the loupe example seems correct but I’d be interested in Nodes output as it has the same API

pcorpet:
> What does nodejs do here? To me the loupe example seems correct but I’d be interested in Nodes output as it has the same API

For reference [this is Node's API](https://nodejs.org/api/util.html#util_custom_inspection_functions_on_objects).

So the test is correct and I should fix the code.

pcorpet:
To make progress on this, I would now need some reviews on loupe pull requests. @keithamus can you take a look please? https://github.com/chaijs/loupe/pulls

keithamus:
All fixes on loupe and been merged and published to npm. This should be able to progress now.

keithamus:
@pcorpet I've also invited you to the chai org as a maintainer for loupe!

pcorpet:
Thanks for the review, this PR should now be good for review as the tests pass.

test/assert.js
```diff
@@ -365,7 +365,7 @@ describe('assert', function () {
 
     err(function () {
       assert.deepEqual(obj1, obj2);
-    }, "expected { tea: \'chai\' } to deeply equal { tea: \'black\' }");
+    }, "expected {} to deeply equal {}");
```
keithamus, line 368: I feel like we're losing data here. Perhaps this should be a feature of loupe that if the object has no own-properties then it can inspect the prototype to make it more useful for instances like this?

test/globalErr.js
```diff
@@ -136,7 +136,7 @@ describe('globalErr', function () {
     err(function () {
       expect(val1).to.equal(val2);
     }, {
-        message: 'expected [ Array(4) ] to equal [ Array(4) ]'
+        message: "expected [ 'string1', 'string2', …(2) ] to equal [ 'string5', 'string6', …(2) ]"
```
keithamus, line 139: This is so much nicer IMO 👌

keithamus (approved):
This LGTM but perhaps a second set of eyes would be good. @flaambe as you worked on this before, do you think this looks good?

flaambe (approved):
LGTM, but I was thinking we are going to replace ```lib/chai/utils/inspect.js``` completely?

keithamus:
I agree that's definitely the ultimate goal here. Merging this PR gets us a lot closer to that though so I'm happy to ship this.

keithamus:
@pcorpet did you want to try porting this PR to the `4.x` branch and we can cut a prerelease to make sure it doesn’t break folks workflow.

PR #1411, opened 2021-08-18
chore: add "type":"module" to package.json

jrobinson01:
Adds "type":"module" to package.json, so chai can be used in browsers as an es module. 

related to: https://github.com/open-wc/open-wc/pull/2267

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "url": "https://github.com/chaijs/chai/issues"
   },
   "main": "./index",
+  "type":"module",
   "exports": {
     ".": {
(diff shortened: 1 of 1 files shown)


jrobinson01:
bump! Anything I can do to help get this pushed forward?

keithamus (approved):
This will also need to be back ported to the `4.x.x` branch for a release there.

keithamus:
I'm going to close this as it won't work in its current state. There needs to be more work done for this to work properly. Perhaps if this is important to folks, it could be followed up with a PR which works?

PR #1427, opened 2021-11-04
Add explicit export for `package.json`

ntkme:
#### Purpose

Add an explicit export entry for `package.json` to fix the following error when running karma:

```
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './package.json' is not defined by "exports" in node_modules/chai/package.json
```

 #### Background

https://github.com/nodejs/node/issues/33460

There is a lot of discussion around whether declaring `package.json` as "public API" via `exports` is a good or bad idea. My take is simply that we need to expose this as a workaround before node team decide what to do, otherwise it breaks many tools that relies on `package.json`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,8 @@
       "require": "./index.js",
       "import": "./index.mjs"
     },
-    "./": "./"
+    "./": "./",
+    "./package.json": "./package.json"
   },
   "scripts": {
     "prebuild": "npm run clean",


ntkme:
Did a little bit more digging and the issue is caused by this line in karma-sinon-chai:

https://github.com/kmees/karma-sinon-chai/blob/c0663bcf150a4f28b0225f91d878a33d2ce8992d/index.js#L34

ntkme:
Opened PR against karma-sinon-chai: https://github.com/kmees/karma-sinon-chai/pull/64

If it get fixed in karma-sinon-chai, I think this PR might not be necessary. Let's see.

keithamus:
Thanks for the PR, I've merged a different fix which should resolve this.

PR #1438, opened 2021-12-17
Update the domain address while it's being renewed.

chuckrincon:
While chaijs/chaijs.github.io#200 is being resolved, the `[.com]` address was updated to `[.github.io]` on the `README.md`

Once chaijs/chaijs.github.io#200 is resolved, this changes will be reverted.


keithamus:
Thanks for this PR @chuckrincon, much appreciated! I’ll close this though as we’ve managed to get the domain pointing back at the docs site!

Issue #1358, opened 2020-08-24
assert.deepEqual / expect().to.deep.equal incorrect with Symbols as keys


derMart:
consider this:
```
var sym = Symbol('foo');
expect({[sym] : 1}).to.deep.equal({[sym] : 0});
assert.deepEqual({[sym] : 1}, {[sym] : 0});
```
I would expect both variants to fail, but they pass and do not throw an Error.
Using chai 4.2.0

The [documentation](https://www.chaijs.com/api/assert/#method_deepequal) does not state anything about this behaviour.
Is this working as intended or a bug?

wertlex:
Well, the problem could be resolved by updating `deep-eql` version.
[`eql`](https://www.chaijs.com/api/bdd/#method_eql) underneath uses `deep-eql` which already supports Symbol comparing. However, it could be a breaking change

snewcomer:
This PR has been merged - https://github.com/chaijs/deep-eql/pull/83

Awaiting a bump on deep-eql!

PR #1483, opened 2022-11-07
fix: deep-eql bump package to support symbols comparison

snewcomer:
ref https://github.com/chaijs/deep-eql/pull/81

close https://github.com/chaijs/chai/issues/1358

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
       "dependencies": {
         "assertion-error": "^1.1.0",
         "check-error": "^1.0.2",
-        "deep-eql": "^3.0.1",
+        "deep-eql": "^4.1.2",
         "get-func-name": "^2.0.0",
         "loupe": "^2.3.1",
         "pathval": "^1.1.1",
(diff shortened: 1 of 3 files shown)


keithamus (approved):

bugwelle:
It seems that https://www.chaijs.com/releases/ does not include the latest releases. Could it be added? This change is rather important because it broke our tests (or rather: showed we had issues).   :-)

snewcomer:
@bugwelle What exact error are you seeing?  Chai.js seems to be published on npm - https://www.npmjs.com/package/chai

bugwelle:
@snewcomer Sorry for the late reply, I must have marked the notification as read...

----------------

The chai release log on https://www.chaijs.com/releases/  does not include 4.3.7. The latest version mentioned there is 4.3.4.  That made it hard to debug.

Our tests failed because we only want to compare enumerable properties (using deep equal). With this change, non-enumerable Symbols are now checked as well.  In my case, we only had to adapt one assertion.  Colleagues, however, will need to adapt way more.

What made this even more bothersome to debug: The default assertion messages we got (using mocha) said something like "`{ a: 1 }` is no equal to `{ a: 1}`". I had to debug it to see that one of the objects had a Symbol on it. 

IMHO, this is a breaking change, which is fine for 5.x, but rather annoying for 4.3.7. In my case, though, my tests are fixed now, so I don't mind it too much.

keithamus:
Sorry about the breaking change :disappointed:, it's not ideal.

simlu:
Man... a breaking change like this with a minor version increment :disappointed:

Thanks for all the hard work, but this definitely just caused me some headaches. Time to roll back!

PR #1498, opened 2023-01-20
ES module conversion PoC

43081j:
I'm sure there's plenty of other things to take into account, tests to add (such that we test both CJS/ESM equally) and maybe some problems in this I missed - however, here's a PoC of converting the codebase to ESM.

As the commit says, it uses esbuild under the hood to create two bundled entrypoints:

- `chai.js` - ESM
- `chai.cjs` - CommonJS

Two because there are still dependencies written in CommonJS, which invalidates the ESM-only tree (meaning you wouldn't be able to use it natively in a browser). Bundling them means we can leave it up to esbuild to figure the dependencies out and ship pure ESM.

If you ever want this or feel like picking it up, let me know, i can probably find a little time to pair on it some way.

The diff's a little funky because I first caught up from 4.x.x into my 5.x.x-based branch, while the current 5.x.x branch is actually out of date (and I fixed the conflicts there).

Also, it'd _probably_ break the `use` function externally (haven't tried it). has no use anymore internally but who knows what third party packages are depending on it. tried to imitate it with some mock exports object but haven't really tested it.

i have tested it with a few chai plugins to see if `use` works, and all seems fine.


43081j:
@keithamus did you ever get chance to take a peek at this? would be good to get some thoughts

giving chai a real es module would be so useful

keithamus:
Thanks for the reminder! I’ll give this a closer look tomorrow ❤️

keithamus:
@43081j I've published this as `chai@5.0.0-alpha.0`. Thanks for the hard work!

43081j:
@keithamus what a guy :D thanks so much for spending some time to review it.

this'll be such a huge step forward for the library IMO.

many of us have moved to ES packages in recent times, but there's no equally mature/stable assertion library yet. so we've ended up with bundled copies of chai, or very new (and experimental) libraries. it'll be great to be able to stick with chai.

43081j:
@keithamus just saw your post about this. we haven't actually dropped commonjs yet. you can still require() the alpha.

if the plan is to not support it, we can halve the npm package size by dropping the commonjs bundle i left in here!

do you want me to do that in a separate PR?

keithamus:
@43081j I understand you created the cjs files in this PR, but yes, I think let's make a clean break from CommonJS to ease releasing. If you're happy to make the separate PR I'm happy to accept it :smile:

43081j:
Sure happy to do that. 

Right now it means there's two bundles basically, double the package size because of needing esm and cjs

Would love to drop the cjs one. We still need to bundle because of cjs dependencies but at least it'll get rid of one of the two bundles

keithamus:
Right. I'd also like to convert the cjs deps we have, like loupe.

43081j:
Yup I agree

End goal should be a purely esm dependency tree. So we don't have to bundle at all and can ship the sources as is

Maybe an issue to track any incompatible dependencies would be good. There's possibly a few

PR #1503, opened 2023-02-14
chore: drop commonjs support

43081j:
@keithamus this is what we talked about before - removing the commonjs bundle.

i wasn't too sure about removing the `exports` so a second opinion would be great.

my thinking was that we have no conditional exports anymore, just the one entrypoint (`chai.js`). someone could try import the source files and they'd probably work ok in node (since it'd handle cjs dependencies). but really everyone should use the main entrypoint.


keithamus (approved):
Yep this LGTM. 

Here's what I'd like to see as an eventuality:

- `./src/` houses a set of TypeScript files.
- `./mod.ts` imports all the TypeScript files - this is what Deno users will use.
- `./lib/` is like `./src/` but with TS types compiled out.
- `./index.js` is like `./mod.ts` but imports from `./lib/` not `./src/` - this is what Node users will use.
- `./chai.js` is an ESM bundle that's a single file that can be dropped into a `<script>` tag in the browser.

keithamus:
@43081j please check your emails/notifications for an invite to the chai team where you can collect your maintainer badge :smile:

43081j:
> 

i agree.

one thing we should consider too is what to use for a test runner going forward.

right now, karma works still because we create a bundle. however, if we try use the source JS (which is perfectly valid now in node), the bare specifiers in imports will cause it to fall over since we don't have an import map specified.

so either we'd need some karma plugin which understands import maps, or we use a test runner like web-test-runner which transforms the imports for you behind the scenes (other similar test runners do the same).

PR #1527, opened 2023-07-21
Update pathval

koddsson:
Since it's now an ES Module! :)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,7 +13,7 @@
         "check-error": "^1.0.2",
         "deep-eql": "^4.1.2",
         "loupe": "^2.3.1",
-        "pathval": "^1.1.1",
+        "pathval": "^2.0.0",
         "type-detect": "^4.0.5"
       },
       "devDependencies": {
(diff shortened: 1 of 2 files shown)


keithamus (approved):

43081j (approved):

43081j:
Good catch! One step closer to pure esm 👀

voxpelli (changes requested):
Would have to change from `require` to `import` here?

https://github.com/chaijs/chai/blob/1a8247f30dbe0f54268a9748ae673caec75d6bfe/lib/chai/utils/index.js#L11

koddsson:
> Would have to change from `require` to `import` here?

Nope, the file has already got `import` calls in the `5.x.x` branch which this PR is targeting.

voxpelli (approved):
Missed that this PR was targeting a non-main branch as it wasn't mentioned anywhere and I didn't think of looking at the target branch as its so rare for projects to merge towards a non-main branch.

I would suggest creating a milestone and/or tracking issue that can convey this context to new contributors.

koddsson:
@keithamus; I don't have merge rights, so I can't land this myself. 

It also might be good to remove node12 and node14 as required builds in the repo settings, since they are never going to be reported.

keithamus:
I've given you merge rights, so feel free to merge 👍

PR #1528, opened 2023-07-21
Update check-error

koddsson:
Since it's now an ES Module! :)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "license": "MIT",
       "dependencies": {
         "assertion-error": "^1.1.0",
-        "check-error": "^1.0.2",
+        "check-error": "^2.0.0",
         "deep-eql": "^4.1.2",
         "loupe": "^2.3.1",
         "pathval": "^1.1.1",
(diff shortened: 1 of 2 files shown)


keithamus (approved):

voxpelli:
@koddsson I see a new release for `check-error` on npm, but nothing in https://github.com/chaijs/check-error – no new tag or commits that refers to `2.0.0`

As for chai itself, you need to alter how its loaded here:

https://github.com/chaijs/chai/blob/1a8247f30dbe0f54268a9748ae673caec75d6bfe/lib/chai/utils/index.js#L148

`require()` can't load ESM-modules, you have to use `(await import()).default` instead.

voxpelli (changes requested):
See my previous comment

koddsson:
> As for chai itself, you need to alter how its loaded here:

I don't know what branch this file is in, but I'm targetting the `5.x.x` branch, which doesn't have the `require` function call.

voxpelli:
> I don't know what branch this file is in, but I'm targetting the `5.x.x` branch, which doesn't have the `require` function call.

Right, I missed that part as it wasn't mentioned in the PR itself and neither any link to any milestone or such :)

voxpelli (approved):

koddsson:
@keithamus; I don't have merge rights, so I can't land this myself. 

It also might be good to remove node12 and node14 as required builds in the repo settings, since they are never going to be reported.

Issue #1525, opened 2023-07-03
chai.version field is incorrect


burnall:
The runtime Chai version is set to '4.3.3', though the current version is 4.3.7.
Please see https://github.com/chaijs/chai/blob/4.x.x/chai.js#L17C18-L17C18
It seems it should be updated with every release.

voxpelli:
Same goes for the `5.x.x` branch, it also references `4.3.3`: https://github.com/chaijs/chai/blob/ad82f6229ee1073a9534cf0d8d9d39418ee9de12/lib/chai.js#L22

keithamus:
PRs welcome 😄

voxpelli:
@keithamus How is it meant to work? Normally when it’s needed I read it from the package.json rather than hard code it, would that work?

keithamus:
We need to support browser environments which cannot read from the disk. Frankly I think we can just drop it from v5 as it serves little value.

peanutenthusiast:
Hi @keithamus! Is this issue still open to PR's?

keithamus:
Yes, please go right ahead.

peanutenthusiast:
Great. Would you mind clarifying the expected change? Just to update to '4.3.7'? Or drop given the previous comment.

koddsson:
We've removed chai version in #1550.

PR #1534, opened 2023-08-23
Fix: update exports.version to current version

peanutenthusiast:
This pull request changes [this line](https://github.com/chaijs/chai/blob/4.x.x/lib/chai.js#L13)
to 4.3.7

See issue: #1525

### DIFF
--- a/lib/chai.js
+++ b/lib/chai.js
@@ -10,7 +10,7 @@ var used = [];
  * Chai version
  */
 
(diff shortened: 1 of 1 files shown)


keithamus:
This will have to be `4.3.8` as we'll need to ship it in a bugfix version, rather than publish over the old `4.3.7` version. 

If you could please update this to `4.3.8` I'll merge and publish!

PR #1537, opened 2023-09-26
fix: Removes the get-fun-name dependency, which is not used in the pr…

xuchen-cx:
Since the dependency get-fun-name is not used in chai, it can be removed

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,6 @@
     "assertion-error": "^1.1.0",
     "check-error": "^1.0.2",
     "deep-eql": "^4.1.2",
-    "get-func-name": "^2.0.0",
     "loupe": "^2.3.1",
     "pathval": "^1.1.1",
     "type-detect": "^4.0.5"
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -43,7 +43,6 @@
     "assertion-error": "^1.1.0",
     "check-error": "^1.0.2",
     "deep-eql": "^4.1.2",
-    "get-func-name": "^2.0.0",
```
Ep1s0de, line 46: "get-func-name": "2.0.0",

i think

samuelsolis:
Furthermore... This dependency are creating some errors.

This line:
`https://github.com/chaijs/chai/blob/4.x.x/lib/chai/utils/inspect.js#L4`

is creating an error in our tests, when are executed in gh (I don't know the reason because in the local instance it's working). The error is this one:

`ERROR: Error [ERR_REQUIRE_ESM]: require() of ES Module /home/runner/work/[PATH]/node_modules/get-func-name/index.js from /home/runner/work/[PATH]/node_modules/chai/lib/chai/utils/inspect.js not supported.`

BTW, it seems a thing that can be improved and even better if we can remove it!

keithamus:
It's not used in Chai but it's used in Chai's dependants, namely `loupe`.

The underlying issue has been fixed though, so we'll close this one for now.

PR #1544, opened 2023-10-20
Inline `type-detect` as a simple function

koddsson:
I think the `type-detect` function can be deprecated as all there is much more spec conformance than was when `type-detect` was created.

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,7 +145,6 @@ Chai offers a robust Plugin architecture for extending Chai's assertions and int
 - [chaijs / chai-docs](https://github.com/chaijs/chai-docs): The chaijs.com website source code.
 - [chaijs / assertion-error](https://github.com/chaijs/assertion-error): Custom `Error` constructor thrown upon an assertion failing.
 - [chaijs / deep-eql](https://github.com/chaijs/deep-eql): Improved deep equality testing for Node.js and the browser.
-- [chaijs / type-detect](https://github.com/chaijs/type-detect): Improved typeof detection for Node.js and the browser.
 - [chaijs / check-error](https://github.com/chaijs/check-error): Error comparison and information related utility for Node.js and the browser.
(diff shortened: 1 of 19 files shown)


lib/chai/utils/type-detect.js
```diff
@@ -0,0 +1,18 @@
...
+  "String",
+  "Boolean",
+  "Null",
+  "Undefined",
+  "Function",
+  "Symbol",
+];
+
+export function type(obj) {
+  const type = Object.prototype.toString.call(obj).slice(8, -1);
```
keithamus, line 12: This could probably be simplified by checking `[Symbol.toStringTag]` on the object.

README.md
```diff
@@ -145,7 +145,6 @@ Chai offers a robust Plugin architecture for extending Chai's assertions and int
 - [chaijs / chai-docs](https://github.com/chaijs/chai-docs): The chaijs.com website source code.
 - [chaijs / assertion-error](https://github.com/chaijs/assertion-error): Custom `Error` constructor thrown upon an assertion failing.
 - [chaijs / deep-eql](https://github.com/chaijs/deep-eql): Improved deep equality testing for Node.js and the browser.
-- [chaijs / type-detect](https://github.com/chaijs/type-detect): Improved typeof detection for Node.js and the browser.
```
keithamus, line 148: We might consider porting all of the tests from `type-detect` to make sure it's stable. This will also help us highlight what a migration path would look like for the `.as()` assertions to just check `[Symbol.toStringTag]`.

keithamus (approved):

43081j:
you're probably right but it does mean we'd be missing the special results like `global` right?

i.e. it looks like the old one `type(globalThis) === 'global'`

README.md
koddsson: Just did, let me know what you think!

keithamus (approved):
This LGTM :+1:

PR #1546, opened 2023-10-21
Replacing Karma with Web Test Runner

koddsson:
Web Test Runner is a pretty good test runner that runs tests in browsers and a good replacement for Karma.

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,11 +58,6 @@
     "codecov": "^3.8.1",
     "esbuild": "^0.17.3",
     "istanbul": "^0.4.3",
-    "karma": "^6.1.1",
-    "karma-chrome-launcher": "^3.1.0",
-    "karma-firefox-launcher": "^2.1.0",
-    "karma-mocha": "^2.0.1",
-    "karma-sauce-launcher": "^4.3.5",
     "mocha": "^8.3.0"
   }
 }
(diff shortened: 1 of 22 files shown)


package.json
```diff
@@ -34,7 +34,7 @@
     "pretest": "npm run build",
     "test": "npm run test-node && npm run test-chrome",
     "test-node": "mocha --require ./test/bootstrap/index.js --reporter dot test/*.js",
-    "test-chrome": "karma start karma.conf.cjs --single-run --browsers HeadlessChrome",
+    "test-chrome": "web-test-runner",
     "test-firefox": "karma start karma.conf.cjs --browsers Firefox",
```
43081j, line 38: should this one be updated too and remove karma entirely? 👀 maybe im just reading the pr before its ready :D

koddsson: hehe yeah the plan is the remove all of this. I just wanted to push this branch to keep it somewhere as I tend to forget random branches on my machine.

koddsson:
Looks like Safari has some nonstandard `sourceURL` property that's showing up in the `AssertionError`. Looking into this.

koddsson:
I actually just dropped checking in Safari. These issues also exist on branch 5.x.x but it looks like we've never tested in Safari before. Since WebKit uses a different standard than V8 we'd need to normalise the stack trace somehow but I'd say that it's out of scope for this work.

Reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/stack

keithamus (approved):

PR #1548, opened 2023-11-02
remove codecov

koddsson:
I don't think this is really used any more and as I understand it `codecov` is deprecated.

https://about.codecov.io/blog/codecov-uploader-deprecation-plan/


keithamus (approved):

koddsson:
> Codecov the module is deprecated, one can use GitHub Actions instead though

Sure! I just don’t think we’re currently using this but if we are and someone is relying on this data we can work on getting that up and running again.

PR #1549, opened 2023-11-02
Remove istanbul

koddsson:
Istanbul is deprecated and I don't think we are doing a lot with code coverage these days.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,7 +22,6 @@
         "@web/test-runner-playwright": "^0.10.2",
         "bump-cli": "^1.1.3",
         "esbuild": "^0.17.3",
-        "istanbul": "^0.4.3",
         "mocha": "^8.3.0"
       },
       "engines": {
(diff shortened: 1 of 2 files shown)


voxpelli:
Replace it with nyc or c8? Modern successors to Istanbul

keithamus (approved):

43081j:
if we can still have coverage, that would be nice. sometimes it is useful just as a quick guide to obvious uncovered blocks, etc locally.

@voxpelli is right in that we should use c8 or nyc if we do that (i prefer c8)

maybe another PR though

koddsson:
> if we can still have coverage, that would be nice. sometimes it is useful just as a quick guide to obvious uncovered blocks, etc locally.
> 
> @voxpelli is right in that we should use c8 or nyc if we do that (i prefer c8)
> 
> maybe another PR though

As far as I can tell the coverage doesn't seem to work at all[^1] and I don't see any complaints about that so I assumed that it's unused. With it not working currently I figure we remove it as non-functional and someone can add coverage back in a different PR.

If it's just used as a developer tool it might be as simple as running `npx nyc mocha` instead of adding it as a dev dependency of the project (that we then have to maintain and update as part of the project).

[^1]: Running `npm run test-cov` ends with the message: "No coverage information was collected, exit without writing coverage information"

Issue #1520, opened 2023-05-30
Allow deepEqual to use a custom comparator


forty:
deep-eql, the underlying library doing deep equality supports passing a custom comparator https://github.com/chaijs/deep-eql/blob/main/index.js#L87

Exposing it in chai would be super useful. This idea has been mentioned in several comments (especially on issue about loose equality, which is one use case for it) for example https://github.com/chaijs/chai/issues/644#issuecomment-220047196 (2016) but I did not find a proper ticket for it.

Syntax:

```
expect({a: 1}).to.eql({b: 2}, 'nooo why fail??', comparator);
expect({a: 1}, 'nooo why fail??').to.eql({b: 2}, comparator);
// maybe
expect({a: 1}, 'nooo why fail??').usingComparator(comparator).to.eql({b: 2});

// assert style
 assert.deepEqual({ tea: 'green' }, { tea: 'green' }, comparator);
```

Example use case: interval

```
const comparator = (left, right) => {
   if(right && typeof right == 'object' && typeof right.__interval' === 'number' && typeof right.__value === number){
      // right is of the form {__value: 100, __interval: 5}, use custom comparison
      if(typeof left !== 'number){
         return false;
      } else {
         return Math.abs(right.__value - left) < right.__interval
      } 
   } else {
      return null; // use default comparison
   }
}
assert.deepEqual({ amount: 103}, { amount: {__value: 100, __interval: 5} }, comparator);
```
(ugly example to make it clear how that would work, but it'd certainly be more elegant with an IntervalComparison class for example which would work the same)

forty:
I had a quick look at the code, and unless I'm mistaken, that doesn't seem horribly hard to do (but I only looked very quickly tbh), at least for the simple expect and assert syntax (which to be frank is all that matters to me, I hate.expect.syntax.with.a.passion but from my understanding of the code, the easiest is to implement it for both interfaces anyway)

forty:
One step further (or maybe it'd be the first step as it doesn't involve changing the API) could be to be able to change the default comparator

```
chai.config.defaultComparator = comparator;
```

which would be even more useful

koddsson:
Hey @forty; I'm not sure if we're able to land this in 5.0 but I think we can do this in a way that it's not a breaking change so it could be done in a minor version after 5.0.

A pull request with this feature would help tremendously but otherwise it's just a matter of time until someone has the time to pick this up to work on.

forty:
[...]

PR #1551, opened 2023-11-10
Allow deepEqual fonction to be configured globally

forty:
Tentative MR to allow configuring the deepEqual function globally, as discussed https://github.com/chaijs/chai/issues/1520

I have done things using a "flag", planning for the possibility of customizing the deepEqual function on a per assertion basis (but this is not implemented in this MR).

Please let me know what you think.

### DIFF
--- a/lib/chai/assertion.js
+++ b/lib/chai/assertion.js
@@ -40,6 +40,8 @@ import * as util from './utils/index.js';
  *   from within another assertion. It's also temporarily set to `true` before
  *   an overwritten assertion gets called by the overwriting assertion.
  *
+ * - `eql`: This flag contains the deepEqual function to be used by the assertion.
+ *
  * @param {Mixed} obj target of the assertion
  * @param {String} msg (optional) custom error message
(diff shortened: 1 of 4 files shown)


forty:
Can I/Should I do the same MR on the 4.x branch? I think it would be useful for users (like me :grimacing: ) which are going to be stuck on this release because of the CommonJS drop

koddsson (approved):
Looks good to me :)

koddsson:
> Can I/Should I do the same MR on the 4.x branch? I think it would be useful for users (like me 😬 ) which are going to be stuck on this release because of the CommonJS drop

Yeah that could be good if you'd like it in the 4.x branch. I'm not sure how much support we can give to the 4.x branch once we release 5.x but we can at least merge PRs and publish the package.

forty:
Awesome, thanks for the review & merge! I created the same MR for 4.x.x then https://github.com/chaijs/chai/pull/1553 (same diff as far I can can tell, except chai.js)

PR #1556, opened 2023-12-21
Fix register-should exports for node 18

### LINKED ISSUE #1555: Package subpath './register-should' is not defined by "exports" in node_modules/chai/package.json for node 18.14.0
"exports": {
      

> "./register-should": "./register-should.js",

      ".": {
        "require": "./index.js",
        "import": "./index.mjs"
      },
      "./": "./"
    },

piyushsi:
Fix #1555

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
   },
   "main": "./index",
   "exports": {
+    "./register-should": "./register-should.js",
     ".": {
       "require": "./index.js",
       "import": "./index.mjs"


43081j:
i suspect you had an error because you were trying to import `chai/register-should` rather than `chai/register-should.js` (with a file extension).

the existing `./*` export map should already handle the one with an extension

see what @keithamus thinks but my preference is you should be using file extensions these days

piyushsi:
@43081j  In both cases i am getting same error :
Error: Cannot find module '/node_modules/chai/register-should'
    at createEsmNotFoundErr (node:internal/modules/cjs/loader:1068:15)
    at finalizeEsmResolution (node:internal/modules/cjs/loader:1061:15)
    at resolveExports (node:internal/modules/cjs/loader:538:14)
    at Function.Module._findPath (node:internal/modules/cjs/loader:607:31)
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:1033:27)
    at Function.Module._resolveFilename.sharedData.moduleResolveFilenameHook.installedValue [as _resolveFilename]

43081j:
strange, that error suggests you're importing it without an extension

```ts
import * as foo from 'chai/register-should.js'; // i would expect this to work right now
import * as bar from 'chai/register-should'; // this should error right now
```

locally, that's the behaviour im seeing

piyushsi:
I am getting error for both cases: 
import 'chai/register-should';
import 'chai/register-should.js';

koddsson:
@piyushsi; Do you have a reproduction for us? Is it just doing `import 'chai/register-should.js';` in a file and then running it with node? What version if so? Is there any other code needed to reproduce? This would be helpful in understanding your issue.

piyushsi:
Screenshot 2023-12-22 at 7 24 39 PM
Screenshot 2023-12-22 at 7 25 21 PM
Screenshot 2023-12-22 at 7 29 45 PM

43081j:
the correct import should be `chai/register-should.js` these days fwiw. i understand adding an extra export mapping seems to make the extensionless import work, but really we should figure out why the correct import doesn't work

we shouldn't need to add that change to the export map imo

something in your build must be changing it somehow (yarn, ts-node, etc).

i would try without yarn (npm), and also try build the ts via tsc then run the resulting js instead. just remove some of the hoops you're jumping through to narrow the problem down

piyushsi:
In the second screenshot, I attempted using the .js format, yet I still encountered the same error.

43081j:
Yes that is true, but the code it executed in the end did not have the extension, hence the error

Which suggests something in the build/test pipeline is transforming the import 

That's why I think it would be good for you to try the standard import with the extension, no changes to the package JSON. Directly in node, or just use TSC directly to output JS, and run that. 

I'm not saying change your repo to do that also, just saying do it temporarily to help narrow the problem down

piyushsi:
Still i am getting this:
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './register-should.js' is not defined by "exports" in /Users/piyushsinha/Desktop/TB/tbi-aic-cloudsync/node_modules/chai/package.json

43081j:
if you can make a small reproduction of that @piyushsi , it would be much appreciated

i can't cause it locally.

locally, i have:

```ts
import 'chai/register-should.js';

console.log('abc'.should);
```

which works fine on chai 4 and 5

piyushsi:
Could you please verify using Node version 18.14.0?

43081j:
> Could you please verify using Node version 18.14.0?

just gave it a try and it still works, so we must be doing something different.

`package.json`:

```json
{
  "name": "temp-chai",
  "type": "module",
  "version": "1.0.0",
  "main": "index.js",
  "devDependencies": {
    "chai": "^4.4.0"
  }
}
```

`main.js`:

```ts
import 'chai/register-should.js';

console.log('abc'.should);
```

then `node main.js` results in logging the `should` object

keithamus:
Given this works as is, I'm going to close this PR. 

We can continue the discussion, and continue to diagnose the issues here in #1555. Perhaps a viable PR to better serve this issue would be a PR to the docs, explaining what the issue could be and how to resolve it?

piyushsi:
Ah, okay, I understand. I was using Chai version 4.3.4, and now it's working for me in version 4.4.0.

Thanks

PR #1566, opened 2023-12-29
Fix 1564

### LINKED ISSUE #1564: [v5] `assert.isFunction()` fails with async function
The function `assert.isFunction` is failing on AsyncFunction on v5.0.0, while on v4.3.10 it was working

A simple scenario is the one below:
```javascript
describe("Scenario", ()=>{
    it("Should be a function", ()=>{
        async function foo() {
            console.log("Hello World");
        }
        assert.isFunction(foo); // fails on v5.0.0
    });
});
```

For now I changed my tests to use the JS builtin `typeof`, like:
```javascript
assert.strictEqual(typeof foo, "function");`
```

koddsson:
Fix #1564 by introducing new `isAsyncFunction` and `isNotAsyncFunction` assertions. 

The thinking is that even though async functions are functions this change will allow for more granular assertions.

I'm still not sure if this is what we'd like though. Async functions are still functions so I could see a argument being made for `isFunction` asserting that a async function is indeed a function. Maybe we should add a `isSynchronousFunction` to specifically check if a function is not async and then leave `isFunction` to assert for both sync and async functions?

Interested in hearing thoughts.

@43081j @keithamus @ReDemoNBR

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -673,7 +673,19 @@ Assertion.addProperty('true', function () {
 });
 
 /**
- * TODO
+ * ### .callable
+ *
+ * Asserts that the target a callable function.
+ *
+ *     expect(console.log).to.be.callable;
+ *
+ * A custom error message can be given as the second argument to `expect`.
+ *
+ *     expect('not a function', 'nooo why fail??').to.be.callable;
+ *
+ * @name callable
+ * @namespace BDD
+ * @api public
  */
 Assertion.addProperty('callable', function () {
   const val = flag(this, 'object')
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -673,6 +673,54 @@ Assertion.addProperty('true', function () {
 });
 
 /**
+ * TODO
+ */
+Assertion.addProperty('callable', function () {
+  const val = flag(this, 'object')
+  const ssfi = flag(this, 'ssfi')
+  const msg = flag(this, 'message') || ''
+
+  if (!['Function', 'AsyncFunction', 'GeneratorFunction', 'AsyncGeneratorFunction'].includes(_.type(val))) {
+    throw new AssertionError(
+      msg + 'expected ' + _.inspect(val) + ' to be callable',
+      undefined,
+      ssfi
+    );
+  }
+});
+
+Assertion.addProperty('asyncFunction', function () {
+  const val = flag(this, 'object')
+  const ssfi = flag(this, 'ssfi')
+  const msg = flag(this, 'message') || ''
+
+  if (!['AsyncFunction', 'AsyncGeneratorFunction'].includes(_.type(val))) {
+    throw new AssertionError(
+      msg + 'expected ' + _.inspect(val) + ' to be a AsyncFunction',
+      undefined,
+      ssfi
+    );
+  }
+});
+
+/**
+ * TODO
+ */
+Assertion.addProperty('generatorFunction', function() {
+  const val = flag(this, 'object')
+  const ssfi = flag(this, 'ssfi')
+  const msg = flag(this, 'message') || ''
+
+  if (!['GeneratorFunction', 'AsyncGeneratorFunction'].includes(_.type(val))) {
+    throw new AssertionError(
+      msg + 'expected ' + inspect(val) + ' to be a GeneratorFunction',
+      undefined,
+      ssfi
+    );
+  }
+});
+
+/**
  * ### .false
  *
  * Asserts that the target is strictly (`===`) equal to `false`.
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -709,14 +709,25 @@ Assertion.addProperty('callable', function () {
   }
 });
 
+/**
+ * @todo
+ */
 Assertion.addProperty('asyncFunction', function () {
   const val = flag(this, 'object')
   const ssfi = flag(this, 'ssfi')
-  const msg = `${flag(this, 'message')}: ` || ''
+  const message = flag(this, 'message')
+  const msg = message ? `${message}: ` : ''
+  const negate = flag(this, 'negate');
 
-  if (!['AsyncFunction', 'AsyncGeneratorFunction'].includes(_.type(val))) {
+  const assertionMessage = negate ?
+    `${msg}expected ${_.inspect(val)} not to be a AsyncFunction` : 
+    `${msg}expected ${_.inspect(val)} to be a AsyncFunction`;
+
+  const isAsyncCallable = ['AsyncFunction', 'AsyncGeneratorFunction'].includes(_.type(val));
+
+  if ((isAsyncCallable && negate) || (!isAsyncCallable && !negate)) {
     throw new AssertionError(
-      msg + 'expected ' + _.inspect(val) + ' to be a AsyncFunction',
+      assertionMessage,
       undefined,
       ssfi
     );
@@ -724,16 +735,24 @@ Assertion.addProperty('asyncFunction', function () {
 });
 
 /**
- * TODO
+ * @todo
  */
 Assertion.addProperty('generatorFunction', function() {
   const val = flag(this, 'object')
   const ssfi = flag(this, 'ssfi')
-  const msg = `${flag(this, 'message')}: ` || ''
+  const message = flag(this, 'message')
+  const msg = message ? `${message}: ` : ''
+  const negate = flag(this, 'negate');
 
-  if (!['GeneratorFunction', 'AsyncGeneratorFunction'].includes(_.type(val))) {
+  const assertionMessage = negate ?
+    `${msg}expected ${_.inspect(val)} not to be a GeneratorFunction` : 
+    `${msg}expected ${_.inspect(val)} to be a GeneratorFunction`;
+
+  const isGeneratorCallable = ['GeneratorFunction', 'AsyncGeneratorFunction'].includes(_.type(val));
+
+  if ((isGeneratorCallable && negate) || (!isGeneratorCallable && !negate)) {
     throw new AssertionError(
-      msg + 'expected ' + _.inspect(val) + ' to be a GeneratorFunction',
+      assertionMessage,
       undefined,
       ssfi
     );
@@ -741,16 +760,24 @@ Assertion.addProperty('generatorFunction', function() {
 });
 
 /**
- * TODO
+ * @todo
  */
 Assertion.addProperty('asyncGeneratorFunction', function() {
   const val = flag(this, 'object')
   const ssfi = flag(this, 'ssfi')
-  const msg = `${flag(this, 'message')}: ` || ''
+  const message = flag(this, 'message')
+  const msg = message ? `${message}: ` : ''
+  const negate = flag(this, 'negate');
 
-  if (_.type(val) !== 'AsyncGeneratorFunction') {
+  const assertionMessage = negate ?
+    `${msg}expected ${_.inspect(val)} not to be a AsyncGeneratorFunction` : 
+    `${msg}expected ${_.inspect(val)} to be a AsyncGeneratorFunction`;
+
+  const isAsyncGeneratorCallable = _.type(val) === 'AsyncGeneratorFunction';
+
+  if ((isAsyncGeneratorCallable && negate) || (!isAsyncGeneratorCallable && !negate)) {
     throw new AssertionError(
-      msg + 'expected ' + _.inspect(val) + ' to be a AsyncGeneratorFunction',
+      assertionMessage,
       undefined,
       ssfi
     );
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -721,6 +721,23 @@ Assertion.addProperty('generatorFunction', function() {
 });
 
 /**
+ * TODO
+ */
+Assertion.addProperty('asyncGeneratorFunction', function() {
+  const val = flag(this, 'object')
+  const ssfi = flag(this, 'ssfi')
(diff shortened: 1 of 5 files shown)


43081j:
my preference would be that `isFunction` matches all types of function, which means either asserting against a union of types or reintroducing the `typeof` code for functions.

an `isAsyncFunction` and `isSyncFunction` seems to make sense then, like you suggested. though things could get a bit sketchy since it doesn't mean any async function, it means any _syntactically_ async function (i.e. via the `async` keyword). A regular function which returns a promise is async, but isn't an `AsyncFunction`.

not sure how best to avoid confusion there

keithamus (commented):
I agree but I'd add that we should aim for consistency between interfaces.

IMO:

- `expect(f).to.be.a.function()`/`f.should.be.a.function()`/`assert.isFunction(f)` should alias a `callable()` assert, and so likewise `not.a.function`/`isNotFunction` should be an alias for `.not.callable()`
- `expect(f).to.be.callable()`/`f.should.be.callbable()`/`assert.isCallable(f)` should check to see if the object can be called (i.e. `typeof f == 'function'`). `not.callable` would be `!= 'function`.
- `expect(f).to.be.an.async.function()`/`f.should.be.an.async.function()`/`assert.isAsyncFunction(f)` can check `be.a('AsyncFunction')`. This means it should be possible to chain `.a.callable.async.function()` even if that's redundant. Because `not` is chainable, `assert.isNotAsyncFunction` will only test that its not _async_, not that it's not callable. So `.a.callable.not.async.function()` would pass for regular functions & generator functions, and fail for async functions (generator or otherwise). 
- `expect(f).to.be.a.generator.function()`/`f.should.be.a.generator.function()`/`assert.isGeneratorFunction(f)` can check `be.a('GeneratorFunction')`, although arguably it should also pass for async generators, as they're still generator functions?
- This also means `expect(f).to.be.an.async.generator.function()`/`f.should.be.an.async.generator.function()`/`assert.isAsyncGeneratorFunction(f)` should work for async generators.

Perhaps the async asserts should check if the stringtag matches `/^Async.*Function$/` and the generator asserts should check it matches `/GeneratorFunction$/`?

ReDemoNBR:
Recently in NodeJS's new [`util.promisify` deprecation on promisified functions](https://github.com/nodejs/node/pull/49609) there was a discussion which involves how to detect if function is already promisified.

One of the points is that a regular Function can return a Promise and thus be used like an AsyncFunction, even though it is not.
On top of that, I had an issue before in a project where a transpiler was converting an AsyncFunction to a Function. So I imagine tools like Babel, Typescript, Webpack, and friends may do similar transpilations and cause some confusion.

So, my preference is that `assert.isFunction(f)` shjould assert for any type of function, which would also be good for compatibility with Chai v4. And then add new precise assertions to test for AsyncFunction, GeneratorFunction, and _SyncFunction_.

koddsson:
Yooo! I pushed some changes based off of https://github.com/chaijs/chai/pull/1566#pullrequestreview-1798856715 for early feedback. Please comment any thoughts (and prayers) if something looks off otherwise I'll continue down this road 😸

lib/chai/interface/assert.js
```diff
@@ -590,6 +591,56 @@ assert.isNotFunction = function (val, msg) {
...
+ * TODO
+ */
+assert.isCallable = function (val, msg) {
+  if (!['Function', 'AsyncFunction', 'GeneratorFunction', 'AsyncGeneratorFunction'].includes(type(val))) {
+    throw new AssertionError(
+      msg + ': expected ' + inspect(val) + ' to be callable',
+      undefined,
+      assert.isCallable
+    );
+  }
```
koddsson, line 604: I think this might be wrong and should just live in `core/assertions.js` and we can call that from here. I'll clean that up later.

ReDemoNBR:
I like the way it

test/assert.js
```diff
@@ -530,6 +530,48 @@ describe('assert', function () {
     }, "blah: expected {} to be a function");
   });
 
+  it('isCallable', function() {
```
ReDemoNBR, line 533: Very nitpicky from me, and I am not the maintainer of this lib.

Right now these tests are placed right in between `isFunction` and `isNotFunction`. I would prefer the more generic `isCallable` to be placed before the `isFunction` and the other ones after `isNotFunction`. This would make it easier for non-regular contributors to find specifics part of the code. Not sure if you guys agree.

ReDemoNBR:
I like the way the code is going and I appreciate your work on this.

From my understanding of the code right now, it looks like this in the `assert` API:
- `assert.isFunction`: changed in v5.0.0 to assert it is a _Function_ (read it as _SyncFunction_)
- `assert.isCallable`: new assertion to replace the old Chai v4 behavior of `assert.isFunction`, which checks if it can be called regardless of being _Function_, _AsyncFunction_, _GeneratorFunction_ or _AsyncGeneratorFunction_
- `assert.isAsyncFunction`: new assertion to check if it is an _AsyncFunction_
- `assert.isGeneratorFunction`: new assertion to check if it is a _GeneratorFunction_

So far, to check if an object is an AsyncGeneratorFunction we could test for AsyncFunction and GeneratorFunction.

Is this right? If so, I'm in love with it :+1: 

----

I believe some tests to ensure the behavior of `assert.isNotFunction` with Async/Generators functions would be nice. What do you guys think?

Also, I am not familiar with the internals of Chai, but does it automatically create the negation of the new assertions, i.e. `assert.isNotGeneratorFunction` and the other ones as well?

lib/chai/interface/assert.js
```diff
@@ -590,6 +591,56 @@ assert.isNotFunction = function (val, msg) {
...
+}
+
+/**
+ * ### .isAsyncFunction(value, [message])
+ *
+ * Asserts that `value` is a async function.
```
ReDemoNBR, line 610: I am not a native english speaker

```suggestion
 * Asserts that `value` is an async function.
```

koddsson: Heh, yeah me neither and I always get a/an mixed up. Nice catch :)

koddsson:
> So far, to check if an object is an AsyncGeneratorFunction we could test for AsyncFunction and GeneratorFunction.

We could add a `isAsyncGeneratorFunction` as well for completeness.

test/assert.js
```diff
@@ -530,6 +530,65 @@ describe('assert', function () {
...
+    var func = async function* () {}
+    assert.isAsyncGeneratorFunction(func) 
+
+    err(function () {
+      assert.isAsyncGeneratorFunction(function() {}, 'blah');
+    }, "blah: expected [Function] to be a AsyncGeneratorFunction");
+    
+    err(function () {
+      assert.isAsyncGeneratorFunction(async function() {}, 'blah');
+    }, "blah: expected [Function] to be a AsyncGeneratorFunction");
```
koddsson, line 585: This should probably be:

```suggestion
    }, "blah: expected [AsyncFunction] to be a AsyncGeneratorFunction");
```

but needs a change in `loupe`. See https://github.com/chaijs/loupe/pull/77

43081j:
> * `expect(f).to.be.a.function()`/`f.should.be.a.function()`/`assert.isFunction(f)` should alias a `callable()` assert, and so likewise `not.a.function`/`isNotFunction` should be an alias for `.not.callable()`

i think it is important we do this part Keith mentioned

otherwise, the various places people use `function` now will have different behaviour than in v4.

e.g. `expect(f).to.be.a('function')`

then you would just negate `isAsyncFunction` or whatever we call it, to check if something is a non-async func

koddsson:
> > * `expect(f).to.be.a.function()`/`f.should.be.a.function()`/`assert.isFunction(f)` should alias a `callable()` assert, and so likewise `not.a.function`/`isNotFunction` should be an alias for `.not.callable()`
> 
> i think it is important we do this part Keith mentioned

Y'all know of a good way to do this that allows us to set the error message to say `expected ${val} to be Function`? I keep getting `expected ${val} to be callable` when aliasing. I wanted to ask before I try rigging up something custom.

lib/chai/core/assertions.js
```diff
@@ -672,6 +672,79 @@ Assertion.addProperty('true', function () {
...
+  const isCallable = !['Function', 'AsyncFunction', 'GeneratorFunction', 'AsyncGeneratorFunction'].includes(_.type(val));
+
+  if ((!isCallable && negate) || (isCallable && !negate)) {
+    throw new AssertionError(
+      assertionMessage,
+      undefined,
+      ssfi
+    );
+  }
+});
```
koddsson, line 698: If this is correct then I should do the same thing for `AsyncFunction`, `GenerationFunction` and `AsyncGeneratorFunction`.

lib/chai/core/assertions.js
```diff
@@ -672,6 +672,79 @@ Assertion.addProperty('true', function () {
...
+  const ssfi = flag(this, 'ssfi')
+  const message = flag(this, 'message')
+  const msg = message ? `${message}: ` : ''
+  const negate = flag(this, 'negate');
+
+  const assertionMessage = negate ?
+    `${msg}expected ${_.inspect(val)} not to be a callable function` : 
+    `${msg}expected ${_.inspect(val)} to be a callable function`;
+
+  const isCallable = !['Function', 'AsyncFunction', 'GeneratorFunction', 'AsyncGeneratorFunction'].includes(_.type(val));
```
43081j, line 689: is this meant to be negated? shouldn't it be true when it includes the type?

koddsson: Yess! Good catch.

43081j:
in its current state, looks like you have this:

* `expect(foo).to.be.callable`
* `expect(foo).to.be.asyncFunction`
* `expect(foo).to.be.generatorFunction`
* `expect(foo).to.be.asyncGeneratorFunction`

i do wonder if we are over complicating it. i think i would have:

* `expect(foo).to.be.callable` (exactly the same as the one below)
* `expect(foo).to.be.a('function')`
  * this one used to work but we broke it as part of our change of how we do `type(val)`
  * it should now be truthy for _all_ types of function (async, generator, async generator, function)
* `expect(foo).to.be.an('AsyncFunction')`
  * should already work (remember `an` and `a` are the same function)

i think we should just fix up `a('function')` and introduce `callable`. i don't think we need explicit property chains for the various types of function.

you could argue then we don't need `callable` either (since its an alias of `a('function')` by then), but maybe people prefer that DX.

what do you think? sorry for any chaos im causing by picking at this 😂

koddsson:
> what do you think? sorry for any chaos im causing by picking at this 😂

No this is great! I think getting this right is important.

>i do wonder if we are over complicating it.

Yeah I think we might be. When in doubt go with the most simple solution so I'm all for it. I'll take a crack at that next.

koddsson:
> `expect(foo).to.be.an('AsyncFunction')`
> * should already work (remember `an` and `a` are the same function)

I was looking at this and this the tests just fail because we're using `deep-eql` in the `expect` assertions which results in these test failures:

```
     AssertionError: expected [AsyncGeneratorFunction] to be an asyncfunction
```

Now we could carve out a exception for functions instead of using `deep-eql` or make a change to `deep-eql` but that seems like it would go against that libraries mantra.

43081j:
strange, that isn't what i expected. what leads to it using deep-eql?

i thought `an` just used a regular old strict equality check on the result of `type(expr)` (`===`)

koddsson:
> strange, that isn't what i expected. what leads to it using deep-eql?
> 
> i thought `an` just used a regular old strict equality check on the result of `type(expr)` (`===`)

Ah! The chai code can be a bit terse to read sometimes 😅  You're correct! I'll make a push in a minute.

lib/chai/core/assertions.js
```diff
@@ -299,17 +299,34 @@ Assertion.addProperty('all', function () {
  * @api public
  */
 
+const functionTypes = {
+  'function': ['function', 'asyncfunction', 'generatorfunction', 'asyncgeneratorfunction'],
+  'asyncfunction': ['asyncfunction', 'asyncgeneratorfunction'],
+  'generatorfunction': ['generatorfunction', 'asyncgeneratorfunction'],
+  'asyncgeneratorfunction': ['asyncgeneratorfunction']
+}
+
```
koddsson, line 308: I kind of hate this and it needs to be moved above the JSDoc comment any ways.

koddsson (commented):
This is quite a bit smaller PR now I think but might need more tests for the different interfaces.

43081j:
as far as the interface goes, looks good to me i think. much simpler than introducing brand new chains for each function type IMO

be good to get keith's opinion some time though too

test/expect.js
```diff
@@ -385,6 +385,55 @@ describe('expect', function () {
...
+  })
+
+  it('asyncGeneratorFunction', function() {
+    expect(async function*() {}).to.be.a('asyncGeneratorFunction');
+    
+    err(function(){
+      expect(async function*() {}, 'blah').to.not.be.a('asyncgeneratorfunction');
+    }, "blah: expected [AsyncGeneratorFunction] not to be an asyncgeneratorfunction");
+  })
+
```
keithamus, line 436: Looks like some more tests could be added here for testing the "sloppyness", for e.g. `expect(async function*() {}).to.be.an('asyncfunction')`.

keithamus (dismissed):
Generally LGTM, but could add a few more tests to prevent future regressions of this type.

koddsson:
I think I've added enough test coverage but let me know if you think I'm missing something and I'll write more :)

keithamus (approved):
This LGTM :+1:

ReDemoNBR (approved):
This LGTM

43081j (dismissed):
🚢 

you have some conflicts too by the looks of it, may want to catch the branch up first

PR #1580, opened 2024-01-24
Re-enable some webkit tests

koddsson:
These should be all fixed in webkit now. We disabled safari in CI because of some stacktrace issues but I ran the tests locally on Safari since the tests aren't running in CI and there wasn't a increase in failures.

Before: Finished running tests in 1.4s with 24 failed tests.
After: Finished running tests in 5.3s with 24 failed tests.


keithamus (approved):
Awesome

PR #1581, opened 2024-01-24
Remove a bunch of if statements in test/should.js

koddsson:
I don't think these are needed anymore as we've dropped support for older browsers.


koddsson:
I'd review the code changes with whitespace hidden: https://github.com/chaijs/chai/pull/1581/files?diff=split&w=1

keithamus (approved):
Great stuff.

PR #1582, opened 2024-01-24
Remove a bunch of unused files

koddsson:
We aren't using bower, karma or saucelabs anymore so we can remove these.


keithamus (approved):
:fire: :fire:

Issue #1468, opened 2022-05-29
assert.sameMembers should work for sets


djmattyg007:
In the mathematical definition of a "set", the contents of a set are known as "members". It therefore stands that an assertion named "same members" should be able to handle sets. I believe this would be significantly more intuitive than using `hasAllKeys`, because a set does not actually have keys.

PR #1583, opened 2024-01-26
Set support in same members

koddsson:
Fixes https://github.com/chaijs/chai/issues/1468

Add `Set` support to `members`/`sameMembers`. While doing that I want to introduce a `iterable` assertion that I think could be useful.

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -319,6 +319,12 @@ function an (type, msg) {
       , 'expected #{this} to be ' + article + type
       , 'expected #{this} not to be ' + article + type
     );
+  } else if (type === 'iterable') {
+    this.assert(
+        typeof obj !== 'string' && obj != undefined && obj[Symbol.iterator]
+      , 'expected #{this} to be ' + article + type
+      , 'expected #{this} not to be ' + article + type
+    );
   } else {
     this.assert(
       type === detectedType
@@ -3037,7 +3043,9 @@ Assertion.addMethod('closeTo', closeTo);
 Assertion.addMethod('approximately', closeTo);
 
 // Note: Duplicates are ignored if testing for inclusion instead of sameness.
-function isSubsetOf(subset, superset, cmp, contains, ordered) {
+function isSubsetOf(_subset, _superset, cmp, contains, ordered) {
+  let superset = Array.from(_superset);
+  let subset = Array.from(_subset);
   if (!contains) {
     if (subset.length !== superset.length) return false;
     superset = superset.slice();
@@ -3175,6 +3183,7 @@ Assertion.addMethod('members', function (subset, msg) {
  * Asserts that the target is an iterable, which means that it has a iterator.
  *
  *     expect([1, 2]).to.be.iterable;
+ *     expect("foobar").to.be.iterable;
  *
  * Add `.not` earlier in the chain to negate `.iterable`.
  *
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -3147,8 +3147,20 @@ Assertion.addMethod('members', function (subset, msg) {
     , flagMsg = flag(this, 'message')
     , ssfi = flag(this, 'ssfi');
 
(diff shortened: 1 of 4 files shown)


keithamus:
I think supporting iterators generally is a good idea. Otherwise though this is a good step forward.

koddsson:
> should we try support all iterables? may be possible by looking for `Symbol.iterator`? 🤔
> 
> just wondering if we want it to support something which isn't a set or array but is iterable



> I think supporting iterators generally is a good idea. Otherwise though this is a good step forward.

Yes! I wanted to change this:

https://github.com/chaijs/chai/pull/1583/files#diff-e3056660026d4431cbb909136ea4290116f5f8be089125a67d6e9ba8f364402dR3145-R3151

to `new Assertion(obj, flagMsg, ssfi, true).to.be.an('iterable');` but ran into issues so I'm coming back to it.

koddsson:
Something that I found is that `string` has `Symbol.iterator` so there might be some edge-cases.

lib/chai/core/assertions.js
```diff
@@ -3037,7 +3043,9 @@ Assertion.addMethod('closeTo', closeTo);
 Assertion.addMethod('approximately', closeTo);
 
 // Note: Duplicates are ignored if testing for inclusion instead of sameness.
-function isSubsetOf(subset, superset, cmp, contains, ordered) {
+function isSubsetOf(_subset, _superset, cmp, contains, ordered) {
+  let superset = Array.from(_superset);
+  let subset = Array.from(_subset);
```
koddsson, line 3048: This might not be the most robust way of doing this. Maybe it's fine.

lib/chai/core/assertions.js
```diff
@@ -3134,14 +3142,29 @@ function isSubsetOf(subset, superset, cmp, contains, ordered) {
...
+    , ssfi = flag(this, 'ssfi');
+
+  if (typeof obj === 'string' || !obj[Symbol.iterator]) {
+    throw new AssertionError(
+      `${flagMsg}: expected ${_.inspect(obj)} to be an iterable`,
+      undefined,
+      ssfi
+    );
+  }
+});
```
koddsson, line 3158: This needs to have a JSDoc comment and needs to move so `members` has the correct JSDoc comment associated with it.

lib/chai/core/assertions.js
```diff
@@ -3134,14 +3142,29 @@ function isSubsetOf(subset, superset, cmp, contains, ordered) {
  * @api public
  */
 
+Assertion.addMethod('iterable', function(msg) {
```
koddsson, line 3145: This needs tests.

lib/chai/core/assertions.js
```diff
@@ -319,6 +319,12 @@ function an (type, msg) {
       , 'expected #{this} to be ' + article + type
       , 'expected #{this} not to be ' + article + type
     );
+  } else if (type === 'iterable') {
```
koddsson, line 322: Another special case for the concept of "iterable" like we did with "functions". Might be good to identify other supersets like this if they already exists in the code.

keithamus: I'd like us to be cautious around adding too many super-sets and making it hard to explain how `an()` works, without extensive docs. Ideally assertions should map to observable expressions in the language, e.g. `x[Symbol.toStringTag]`, `typeof x`.

lib/chai/core/assertions.js
```diff
@@ -3136,26 +3142,29 @@ function isSubsetOf(_subset, _superset, cmp, contains, ordered) {
  * @api public
  */
 
-Assertion.addMethod('members', function (subset, msg) {
+Assertion.addMethod('iterable', function(msg) {
```
keithamus: 👍 this is good

koddsson:
I was gonna point this PR to #1592 but I guess I can't because it's on a different remote. In any case, I split out the `iterable` assertion into it's own PR which I think we should merge first and then land this fix.

koddsson:
I think this is good but might benefit from more tests.

keithamus (approved):

Issue #1425, opened 2021-10-18
Bad comments in `chai.js` (and elsewhere)


elibarzilay:
Looks like #451 switched many docstring comments from `/**` to `/*!` -- where the latter are comments that are not dropped by minifies / bundlers / etc, and are mainly intended for copyright notices.  As a result, having chai included in a bundle that esbuild generates has a whole bunch of useless comments at the end (see below).  Shouldn't these non-copyright comments use `/**`?

```js
... all code ...
...
/*!
 * Module dependencies.
 */
/*!
 * Module export.
 */
...
```

PR #1598, opened 2024-02-01
Convert comments in "legal comments" format to jsdoc or normal comments

koddsson:
These are preserved by default in some bundlers (not sure why you'd bundle chai but 🤷🏻 ). Figured this was more correct than not.

See https://esbuild.github.io/api/#legal-comments

Fixes https://github.com/chaijs/chai/issues/1425

### DIFF
--- a/lib/chai.js
+++ b/lib/chai.js
@@ -15,10 +15,7 @@ import {assert} from './chai/interface/assert.js';
 
 const used = [];
 
-/*!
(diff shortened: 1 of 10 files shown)


43081j (approved):
Makes sense to me, may be interesting to know if there was originally any reason behind this though

keithamus (approved):

PR #1601, opened 2024-02-12
Assert interface fix 

developer-bandi:
- notOwnInclude, notDeepOwnInclude description modify include negative expression
- ")" is added to the example code where there is a syntax error because it does not have ")"


43081j (approved):
good catch!

PR #1605, opened 2024-02-19
Set up ESLint for JSDoc comments

koddsson:
I really want to explore both better typing and possible automatic generation of API docs since the JSDoc comments that exist in the codebase are of such high quality.

This PR introduces ESLint and `eslint-plugin-jsdoc` to start linting JSDoc comment in effort to make sure that all of the comments stay high quality as we refactor the code

This PR can be followed up by more changes to the rules we apply as well as some explorations into API documentation and stricter type-checking.


eslint.config.js
```diff
@@ -0,0 +1,12 @@
+import jsdoc from "eslint-plugin-jsdoc";
+
+export default [
+  jsdoc.configs["flat/recommended"],
```
43081j, line 4: while you're in there, can you maybe also enable eslint's recommended rule set? im surprised we didn't already have eslint

if it ends up with too big of a change, you could do a PR to add eslint + recommended, then a separate one for jsdoc on top

lib/chai/core/assertions.js
```diff
@@ -2116,6 +2110,13 @@ function assertOwnPropertyDescriptor (name, descriptor, msg) {
 Assertion.addMethod('ownPropertyDescriptor', assertOwnPropertyDescriptor);
 Assertion.addMethod('haveOwnPropertyDescriptor', assertOwnPropertyDescriptor);
 
+/**
+ *
```
43081j, line 2114: maybe add a description? "Flags that the `length` property should be asserted in this chain"

43081j (approved):
looks good, though it may be worth marking optional params as such or just wait for my typescript conversion to land

e.g. `@param {string=} iAmAnOptionalParam`

koddsson:
lol, maybe I shouldn't have had it on auto-merge :D 

I'll make a follow up PR to apply some of your suggestions @43081j

PR #1606, opened 2024-02-20
Support big int in approximently

### LINKED ISSUE #1465: assert.approximately does not work with bigints
Basically what the title says.

`assert.approximately(1n, 2n, 1n, "Should not throw");` results in `AssertionError: Should not throw: expected 1n to be a number`

koddsson:
Added a new assertion property `numeric` which checks if given variable is of type `Number` or `BigInt`. Then used that new assertion to replace a `is.a('number')` check in the `closeTo`/`approximately` assertion.

Also replaced a `Math.abs` function call which doesn't support `BigInt` which all adds up to supporting `BigInt` in `closeTo`/`approximately`.

## TODO

- [x] Write more tests for `numeric` assertion.

## Follow up

- Look for more places where we should support BigInt.

Fixes #1465


lib/chai/core/assertions.js
```diff
@@ -3013,19 +3025,14 @@ function closeTo(expected, delta, msg) {
     , flagMsg = flag(this, 'message')
     , ssfi = flag(this, 'ssfi');
 
-  new Assertion(obj, flagMsg, ssfi, true).is.a('number');
-  if (typeof expected !== 'number' || typeof delta !== 'number') {
-    flagMsg = flagMsg ? flagMsg + ': ' : '';
-    var deltaMessage = delta === undefined ? ", and a delta is required" : "";
```
43081j, line 3019: By removing this logic, we have lost some nice messages right? 

Where we used to specify that the delta was wrong, for example, we would now just complain that some value was meant to be numeric

Could we maybe pass custom messages to the new assertions to help with that? (On L3028)

koddsson: Yup, we can change the message back to match what we had. You can see in the diff where I updated all the messages for this message.

In short for tests like this:

```js
expect(1.5).to.be.approximately(1.0);
```

The message changed thusly:

```diff
< "the arguments to closeTo or approximately must be numbers, and a delta is required"
> "expected undefined to be numeric"
```

I thought this message would be more clear as it identifies which arguments are wrong and what is expected. But reading it back now it's not clear that there's a argument missing.

I'll try to see if I can add this back in a way that's more clear than what was here previously.

koddsson:
@43081j; let me know what you think of 45f8825

lib/chai/core/assertions.js
```diff
@@ -3013,19 +3025,18 @@ function closeTo(expected, delta, msg) {
...
-        undefined,
-        ssfi
-    );
-  }
+  new Assertion(obj, flagMsg, ssfi, true).is.numeric;
+  let message =  'A `delta` value is required for `closeTo`';
+  if (!delta) throw new AssertionError(flagMsg ? `${flagMsg}: ${message}` : message, undefined, ssfi);
+  new Assertion(delta, flagMsg, ssfi, true).is.numeric;
+  message =  'A `expected` value is required for `closeTo`';
+  if (!delta) throw new AssertionError(flagMsg ? `${flagMsg}: ${message}` : message, undefined, ssfi);
```
43081j, line 3033: Can you have a delta of `0`? 👀 Do we need to compare explicitly against undefined rather than a falsy check?

koddsson: Could be but then we wouldn't be comparing to see if the two values are close since they'd need to be the same.

I can't think of a reason why someone would do that but adding a test and making this a specific `undefined` check is good in case someone does want to do that in future for whatever reason.

lib/chai/core/assertions.js
```diff
@@ -1633,12 +1643,14 @@ Assertion.addMethod('within', function (start, finish, msg) {
   if (doLength && objType !== 'map' && objType !== 'set') {
     new Assertion(obj, flagMsg, ssfi, true).to.have.property('length');
   }
+  
+  const isNumeric = x => ['Number', 'BigInt'].includes(_.type(x))
```
43081j, line 1647: Can we extract this to somewhere reusable? Seems to be repeated a few times

koddsson: Sure! I didn't want to add this to the global scope via `utils` because it adds another way to check if a value is numeric.

PR #1609, opened 2024-02-26
fix: support some virtual contexts in `toThrow`

43081j:
This adds support for VM situations where we pass a `RegExp` from another process.

Note that we don't have a full fix for this stuff until `check-error` also supports `Error` being from another origin.

**There's also a tiny fix in here for supporting `throw ''`.** I have left a PR comment by the relevant line of code

### DIFF
--- a/lib/chai/utils/index.js
+++ b/lib/chai/utils/index.js
@@ -95,6 +95,12 @@ export {isNaN} from './isNaN.js';
 // getOperator method
 export {getOperator} from './getOperator.js';
 
+/**
+ * Determines if an object is a `RegExp`
+ * This is used since `instanceof` will not work in virtual contexts
+ * @param {*} obj Object to test
+ * @return {boolean}
+ */
 export function isRegExp(obj) {
   return Object.prototype.toString.call(obj) === '[object RegExp]';
 }
--- a/lib/chai/utils/index.js
+++ b/lib/chai/utils/index.js
@@ -94,3 +94,7 @@ export {isNaN} from './isNaN.js';
 
 // getOperator method
 export {getOperator} from './getOperator.js';
+
+export function isRegExp(obj) {
+  return Object.prototype.toString.call(obj) === '[object RegExp]';
+}
--- a/lib/chai/utils/index.js
+++ b/lib/chai/utils/index.js
@@ -98,8 +98,9 @@ export {getOperator} from './getOperator.js';
 /**
  * Determines if an object is a `RegExp`
  * This is used since `instanceof` will not work in virtual contexts
+ *
  * @param {*} obj Object to test
- * @return {boolean}
+ * @returns {boolean}
  */
 export function isRegExp(obj) {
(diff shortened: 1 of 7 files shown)


web-test-runner.config.js
```diff
@@ -5,7 +5,10 @@ const commonjs = fromRollup(rollupCommonjs);
 
 export default {
   nodeResolve: true,
-  files: ["test/*.js"],
+  files: [
+    "test/*.js",
+    "!test/virtual-machines.js"
```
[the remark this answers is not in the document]
43081j: good catch! i forgot to `git add` it

keithamus (approved):
LGTM

lib/chai/utils/index.js
```diff
@@ -94,3 +94,7 @@ export {isNaN} from './isNaN.js';
 
 // getOperator method
 export {getOperator} from './getOperator.js';
+
+export function isRegExp(obj) {
+  return Object.prototype.toString.call(obj) === '[object RegExp]';
+}
```
koddsson, line 100: Why is this needed?

koddsson (approved):
Looks good :)

lib/chai/utils/index.js
43081j: A RegExp from a virtual context isn't the same RegExp as in the owning context, so instanceof won't work

If you pass chai into the virtual context, it'll be comparing against it's own RegExp rather than that of the virtualised code

43081j:
@koddsson could you re-review this? i updated slightly to support throwing of _anything_ (so even `throw undefined` will work)

and i wrapped `getConstructorName` in a try/catch since it may be possible we pass a non-constructable function to it

test/virtual-machines.js
```diff
@@ -0,0 +1,31 @@
...
+}
+
+describe('node virtual machines', function () {
+  it('throws', function() {
+    const shouldNotThrow = [
+      `assert.throws(function() { throw ''; }, /^$/);`,
+      `assert.throws(function() { throw new Error('bleepbloop'); });`,
+      `assert.throws(function() { throw new Error(''); });`,
+      // TODO (43081j): enable this test once check-error supports
+      // cross-vm `Error` objects
```
koddsson, line 19: Could we link to a issue on `chaijs/check-error` so it's easy to follow this work. In case this doesn't get done in the near future we then have captured the relevant context. If you got a PR practically ready then no worries of course.

koddsson (approved):
Looks good to me! I have one question but happy to merge whatever the answer to that is.

test/virtual-machines.js
43081j: this is the one you published today 🎉 

i just need to update the branch i think and remove this comment

koddsson: Oh I see haha 👯

koddsson (approved):
Looks good!

PR #1615, opened 2024-05-09
Publish npm packages with provenance

koddsson:
This will allow us to have a cool badge on npm.

Example:
image

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -1,7 +1,7 @@
 # This workflow will run tests using node and then publish a package to GitHub Packages when a release is created
 # For more information see: https://help.github.com/actions/language-and-framework-guides/publishing-nodejs-packages
 
-name: Node.js Package
+name: Publish to npm
 
 on:
   release:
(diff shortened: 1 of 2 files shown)


.github/workflows/npm-publish.yml
```diff
@@ -23,13 +23,20 @@ jobs:
...
         with:
-          node-version: 20
-          registry-url: https://registry.npmjs.org/
+          node-version: 22.x
+          registry-url: "https://registry.npmjs.org"
+          cache: "npm"
       - run: npm ci
-      - run: npm publish
+      - run: npm run build --if-present
+      - run: npm version ${TAG_NAME} --git-tag-version=false
```
43081j, line 37: Does this mean we can create releases without having to push a version bump to main? 

Will we lose the tags and version in our package file by doing this? Sounds nice being able to release without the extra push but also I like having tags and the right version in the JSON 😅

keithamus: > Does this mean we can create releases without having to push a version bump to main?

Yes.

> Will we lose the tags and version in our package file by doing this? Sounds nice being able to release without the extra push but also I like having tags and the right version in the JSON 😅

Also yes. Typically I set package.json version to `0.0.0-development` so it's clear that the version is generated.

keithamus (dismissed):
:+1: LGTM

.github/workflows/npm-publish.yml
43081j: No worries I think that makes sense. @koddsson can you update the package version so that is clear?

43081j (approved):

43081j (approved):

Issue #1603, opened 2024-02-13
use() function only returns plugin first time


ramicohen303:
Related to #1569 

I have multiple test files, each loading its own copy of chai and chai-http plugin. The first file works fine, all others fail. If I comment out the first file, the second file works fine, all others fail.

Looking at the code, I believe the issue cause is in chai.js, in the "use" function. The "used" array is global, and once a plugin is added to it, it cannot be added again. However, the exports object is instantiated on every call, but only the first instance gets the plugin.

For now, I created a factory function that returns a singleton chai and chai-http, and I use it in every test file. However, I believe this should be fixed properly.

koddsson:
@ramicohen303; could you write up a quick failing test in pseudo code for this? It would be helpful so we can make sure we are fixing the correct thing.

jdmarshall:
I’m willing to work on this but I still don’t understand the problem.


   ```
  if (!~used.indexOf(fn)) {
     fn(exports, util);
     used.push(fn);
   }

return exports;
```

Seems like that returns the correct results every time. I don’t understand why your multiple use scenario wouldn’t work.

I haven’t seen this problem with sinon-chai or chai-as-promised. Something tells me perhaps your http addon is the issue?

**Comment by 43081j** [comment]
because `use` returns a namespace-like object, and that is the object which plugins can mutate.

so if you have two test files which both do this:

```ts
import * as chai from 'chai';
import {fn} from './shared-lib.js';

chai.use(fn);
```

the `fn` is the same function in both files, so chai will not call it again, and will not push it into `used` (since its already there)

but the `exports` object chai mutates is unique to each test file, so every file after the first one will not call `fn` again.

in a well designed chai plugin, this is ok, because the plugin should mutate prototypes and such (i.e. stuff that exists across all files).

the only place this will fall over is if a plugin tries to mutate `exports` itself (tracked in #1569 )

PR #1617, opened 2024-05-14
Enable chaining of chai plugins

koddsson:
This changes moves the responsibility of checking if a plugin has already been registered to the plugin itself.

I _think_ this fixes https://github.com/chaijs/chai/issues/1603, if not I probably need a better test to make sure.

### DIFF
--- a/test/plugins.js
+++ b/test/plugins.js
@@ -3,6 +3,8 @@ import {use, expect, Should} from '../index.js';
 function plugin(chai) {
   if (chai.Assertion.prototype.testing) return;
 
+  chai.assert.testing = 'successful';
+
   Object.defineProperty(chai.Assertion.prototype, 'testing', {
     get: function () {
       return 'successful';
@@ -12,6 +14,8 @@ function plugin(chai) {
 
 function anotherPlugin(chai) {
   if (chai.Assertion.prototype.moreTesting) return;
+
+  chai.assert.moreTesting = 'more success';
 
   Object.defineProperty(chai.Assertion.prototype, 'moreTesting', {
     get: function () {
@@ -77,6 +81,26 @@ describe('plugins', function () {
       const {expect} = use(anotherPlugin);
 
       expect(expect('').moreTesting).to.equal('more success');
+    });
+  });
+
+  describe('assert', () => {
+    it('basic usage', function () {
+      const {assert} = use(plugin);
+      expect(assert.testing).to.equal('successful');
+    });
+
+    it('multiple plugins apply all changes', function () {
+      const chai = use(plugin).use(anotherPlugin);
+
+      expect(chai.assert.testing).to.equal('successful');
+      expect(chai.assert.moreTesting).to.equal('more success');
+    });
+
+    it('.use detached from chai object', function () {
+      const {assert} = use(anotherPlugin);
+
+      expect(assert.moreTesting).to.equal('more success');
     });
   });
 });
--- a/test/plugins.js
+++ b/test/plugins.js
@@ -1,4 +1,4 @@
-import {use, expect} from '../index.js';
+import {use, expect, Should} from '../index.js';
 
 function plugin(chai) {
   if (chai.Assertion.prototype.testing) return;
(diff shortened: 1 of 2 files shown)


43081j:
this would mean chai extensions (properties, methods, etc) get overwritten/added each time right?

do we know what our suggestion would be to plugin authors for ensuring they don't accidentally duplicate things?

just thinking of a situation where they add a method which wraps the original. so if you call it many times, you end up with `customMethod(customMethod(originalMethod(...args)))` basically (which would probably work in many cases but be not great)

koddsson:
> this would mean chai extensions (properties, methods, etc) get overwritten/added each time right?

I think so. Isn't that how it worked in `chai@4` as well?

> do we know what our suggestion would be to plugin authors for ensuring they don't accidentally duplicate things?
> 
> just thinking of a situation where they add a method which wraps the original. so if you call it many times, you end up with `customMethod(customMethod(originalMethod(...args)))` basically (which would probably work in many cases but be not great)

I'm probably missing something so maybe we can express this as a test? Something like:

```js
function somePlugin(chai) {
  if (chai.Assertion.prototype.testing) return;

  Object.defineProperty(chai.Assertion.prototype, 'testing', {
    get: function () {
      return 'successful';
    },
  });
}

function someOtherPlugin(chai) {
  if (chai.Assertion.prototype.testing) return;

  Object.defineProperty(chai.Assertion.prototype, 'testing', {
    get: function () {
      return 'failed';
    },
  });
}

let chai = use(somePlugin);
chai = chai.use(someOtherPlugin);

expect(chai.myProperty).to.equal('successful');
```

43081j:
in chai 4 it would've only called it once afaik (thats what `used` was for)

i may be talking nonsense but maybe someone does something like this:

```ts
chai.overwriteProperty('equal', function (_super) {
  if (something) {
    return _super.call(this);
  }
  return someOtherThing();
});
```

which presumably, when called many times in some cases, would result in `_super(_super(_super()))` etc

so it may be worth us documenting some recommended way of avoiding that, i.e. checking you already overwrote the property, added the method, etc

43081j:
basically im not sure if `overwriteProperty` multiple times results in the original being `super` or the last overwritten one

given the example before:

```ts
chai.overwriteProperty('equal', function (_super) {
  if (something) {
    return _super.call(this);
  }
  return someOtherThing();
});
```

i wonder if `_super` is the _current_ property, not the original

so if you did this:

```ts
const handler = function (_super) {
  console.log('Called!'); // log something out

  if (something) {
    return _super.call(this);
  }
  return someOtherThing();
};

chai.overwriteProperty('equal', handler);
chai.overwriteProperty('equal', handler); // equivalent to you calling `use` twice

chai.expect(123).to.equal(123);
```

does it log `Called!` twice or once?

im just saying, if it calls it twice, we should probably document a recommended way of avoiding registering the same thing twice

does it make sense now?

koddsson:
> does it make sense now?

I think I'm getting it. I was hoping to run some of the code so I could poke and prod and compare chai@4 and chai@5 but I can't get anything to work. Running this in chai@4 results in a error:

```js
const chai = require("chai");

function somePlugin(chai) {
  chai.util.overwriteProperty(chai.Assertion.prototype, "equal", function (_super) {
    console.log("Called!"); // log something out

    console.log(_super);
    return _super.call(this);
  });
}

chai.use(somePlugin);
chai.use(somePlugin);

chai.expect(123).to.equal(123);
```

>im just saying, if it calls it twice, we should probably document a recommended way of avoiding registering the same thing twice


So document things like https://github.com/chaijs/chai/pull/1617/files#diff-919c27039f4865fa252acfbd90d91c8b6c33b42dbf3d593fac7648cc6aa7a3f5R4 but maybe a bit more reliable? Otherwise we could also document that people should be vary of registering the same plugin twice as that can result in undefined behaviour?

43081j:
i just tried it and it seems to call it once in both 5 and 4

its because its a method (`equal`), not a property:

```ts
import * as chai from 'chai';

function somePlugin(base) {
  base.util.overwriteMethod(base.Assertion.prototype, "equal", function (_super) {
    return function(...args) {
      console.log("Called!"); // log something out

      return _super.call(this, ...args);
    };
  });
}

chai.use(somePlugin);
chai.use(somePlugin);

chai.expect(123).to.equal(123);
```

so i think you can ignore me and all is fine

test/plugins.js
```diff
@@ -1,42 +1,106 @@
...
+    }).to.not.throw;
+  });
+
+  describe('should', () => {
+    before(() => {
+      Should();
+    });
+
+    it('basic usage', function () {
+      use(plugin);
```
43081j, line 49: do these tests actually pollute each other? in that `use` in this test will mean the next one already has `plugin` setup?

not sure how you'd solve that though since we can't reset the prototype 🤔  we'd have to use a unique plugin for each test which is also a bit of a pain

koddsson: Aha! Yes they do!

What about I set `configurable: true` in the plugins and then `delete` the properties in the `after`? I think that should work.

43081j: That could work. Maybe give it a try, hopefully the tests still pass 🙏

koddsson: I ended up leaning into the global nature of plugins and just apply them once in the beginning of tests.

test/plugins.js
```diff
@@ -1,42 +1,110 @@
...
+function plugin(chai) {
+  if (chai.Assertion.prototype.testing) return;
+
+  chai.assert.testing = 'successful';
+
+  Object.defineProperty(chai.Assertion.prototype, 'testing', {
+    get: function () {
+      return 'successful';
+    },
+  });
```
koddsson, line 17: I thought it was surprising I had to do this. Unless I'm misunderstanding how this is supposed to work.

If it's correct that we need to define a property on both `chai.assert` and `chai.Assertion.prototype` I was thinking we could think about a more user friendly way to write plugins for the next breaking version.

43081j: you would normally use `util.addProperty` at least, like:

```ts
util.addProperty(Assertion.prototype, 'testing', function () {
  this.assert(this._obj === 'foo');
});
```

so expect and should will inherit that

but you're right you'd still have to manually extend the assert interface i think, something like:

```ts
assert.isTesting = function() {
  new Assertion(val).testing;
}
```

koddsson:
I'm cleaning up some old PRs and am gonna close this. I'll re-open a different one if I end up working on this.

Issue #1616, opened 2024-05-13
chai-subset is unmaitained


GustavoOS:
Hi

As in https://github.com/debitoor/chai-subset/issues/84 , chai-subset is unmaitained. The biggest contributor no longer has owner rights, but proposed to fork and maintain it. Is chaijs interest in owning the fork?

keithamus:
Sure. Happy to add it to the org.

43081j:
it may also be worth just building this functionality into chai itself, as some variant of `includes`

much of the logic in chai-subset exists in chai already, just not for this particular usage (e.g. recursively checking equality of nested types)

maybe this could just be achieved by some modifier of `includes` ("partial" or something, or `includes.subset`)

just a thought

keithamus:
I am happy to do that, but we should only do that _after_ transferring or at least discussing with the owners/maintainers.

eagleeye:
Hey guys, I'm the main contributor to the chai-subset while I was working in Debitoor. I couldn't convince them to transfer ownership. I think it should be fine to add it into core functionality of chai of fork the repo and support the fork. The question is what is preferrable for the chaijs org?

keithamus:
I think it would be great to clarify our intent to debitoor before moving forward. There's a strong chance if we fork the library into this org, or otherwise integrate changes into core, then the debitoor library will be rendered obsolete. I don't want that to come as a surprise to them. So I think a good next step is perhaps contacting them again, if you don't mind?

In terms of next steps beyond that, I think we can look at what it'll take to integrate directly into core (PRs welcome :wink:) but also I think having a fork (or ideally the source repo) under the chaijs org for archival purposes would be a good step also.

zetxek:
Thanks for the convo and for letting us know! Both here and in https://github.com/debitoor/chai-subset/issues/84 :)

While `debitoor` as an org is not active, it's now part of [@sumup](https://github.com/sumup) :) - and while we do maintain some open source software (https://github.com/sumup-oss), as you have seen, chai-subset (https://github.com/debitoor/chai-subset/issues) is not among the well-kept ones.

Feel free to fork and integrate the codebase into chaijs :) 

For the ownership transfer I'd have to look into the process to transfer it (might take longer). With the matter of the original repo going obsolete - maybe we could add a notice to the readme and mark it as deprecated? (given that the last commit is from 2017, 7 years ago)

eagleeye:
[...]

PR #1621, opened 2024-05-20
Integrate `chai-subset` into chai core

koddsson:
I tried to make sure to preserve the `chai-subset` history when doing this.

Fixes https://github.com/chaijs/chai/issues/1616


43081j:
it looks good to me but i have a couple of thoughts:

1 - i know you put effort into keeping the git history but i'd actually squash this anyway if we merge it as i don't think the subset history makes sense in our tree. it may be better to just note somewhere this was originally built by andrii

2 - something's bugging me about the overlap here of `includes`. we have basically the same algorithm in `includes` (written differently) than we do here, but in that one we don't do partial equality on the deep properties.

i wonder if we can at least extract the traversal logic so they both share it, or we make `includes` have a flag that achieves what subset does. im cautious of putting too much logic in `includes` though as its already a pretty big, multi-use assertion.

keithamus:
> 2 - something's bugging me about the overlap here of `includes`.

We'll likely have similar logic in `deep-eql` too. I'll add that `deep-eql` is also very well optimised. I wonder if we can leverage it and it's custom comparators to effectively traverse the first layer?

koddsson:
I'll start looking at squaring this up according to the comments.

koddsson:
I'm cleaning up some old PRs and am gonna close this. I'll re-open a different one if I end up working on this.

PR #1644, opened 2024-10-04
feat: add assert deepMatch methods

### LINKED ISSUE #1643: [Feature Request] - Add a full regex match method on assert
Unfortunately the `.match` follows the standard Javascript Regex engine behavior of `string.test` returning true if there is a partial match. This behavior works great except for a few examples. 

Example) 

```js
const freeRegex = /fr[e3]|[e3]|[e3] money/i; 
```

This regular expression will register as match  for the phrase `fr33 mon3y` despite that logic tells us there shouldn't be.  I want a method  that runs `String.match` from the input ( `fr33 mon3y`) on `freeRegex`. From there I want to compare if the the first match (index 0 ) from the return value) is equal to original input. 

It would be the equivalent of running something to the following as a one line assertion. 

```js
const freeRegex =  /fr[e3][e3] mon[e3]y/ig;
const matches = "fr33 mon3y".match(freeRegex); 
chai.assert.equal(matches[0],"fr33 mon3y") 
``

Thank you.

a2937:
This adds and test two brand new assert methods. The methods are `assert.deepMatch` and `assert.notDeepMatch`. The purpose of these two methods are to ensure that full regex matches are caught like in the case of the regex "/fr[e3]|[e3]|[e3] money/i"  and the test string "fr33 mon3y". While there are partial methods, there wasn't a way to assert there weren't.

Appropriate tests have been added. 

Closes #1643.

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -2247,6 +2247,49 @@ Assertion.addMethod('match', assertMatch);
 Assertion.addMethod('matches', assertMatch);
 
 /**
+ * ### .match(re[, msg])
+ *
+ * Asserts that the target matches the given regular expression `re`.
+ *
+ *     expect('foobar').to.match(/^foo/);
+ *
+ * Add `.not` earlier in the chain to negate `.match`.
+ *
+ *     expect('foobar').to.not.match(/taco/);
+ *
+ * `.match` accepts an optional `msg` argument which is a custom error message
+ * to show when the assertion fails. The message can also be given as the
+ * second argument to `expect`.
+ *
+ *     expect('foobar').to.match(/taco/, 'nooo why fail??');
+ *     expect('foobar', 'nooo why fail??').to.match(/taco/);
+ *
+ * The alias `.matches` can be used interchangeably with `.match`.
+ *
+ * @name match
+ * @alias matches
+ * @param {RegExp} re
+ * @param {string} msg _optional_
+ * @namespace BDD
+ * @public
+ */
+function assertDeepMatch(re, msg) {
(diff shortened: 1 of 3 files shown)


keithamus (changes requested):
Thanks for working on this! I have some comments.

The `deep` keyword already has a defined semantic in our API, so I don't think the naming is correct here.

I'm also not _fully_ convinced we need this. Nothing in https://github.com/chaijs/chai/issues/1643 nor https://github.com/freeCodeCamp/freeCodeCamp/issues/56471 is especially compelling to the use case here, in my opinion. Code camps having false positives is common enough and not really the intended use case for a testing library. I would expect to see more developers noting this as a problem when testing their applications. I _doubt_ many people are using `.exec()` for regexps, and if - during testing - a false positive occurred I would imagine a developer would tweak the regex rather than looking for alternate APIs to do the same thing.

In order for this to land here's what I'd like to see:

1. Strong justifiable use cases for this. Why does this need to be added? What problems does it solve for people testing their applications? Why would they reach for this over `.match`?
2. An alternate name for the method. The word `deep` must not be used, perhaps `full` can be used in place. I'm also not enamoured with using the word `match` when it uses `.exec()` but that's not as important.
3. More compelling unit tests that verify this functionality and how it differs from `.match` - the current tests don't really achieve this IMO.
4. Likewise, some more detailed docs which describe the differences.

I would expect all 4 of these to be resolved before we merge this, but I also want to stress that time shouldn't be spent on 2-4 before resolving 1; as it stands this change isn't sufficiently justified, IMO, even after resolving 2-4.

PR #1659, opened 2025-01-07
chore: enable eslint recommended config

43081j:
Enables the recommended lint rules from `eslint`.

Also deletes `getEnumerableProperties.js` which is no longer referenced.

technically someone could've been importing this file directly, but it seems extremely unlikely. i searched github for it and the only meaningful result is cypress, but it uses chai 4.x still.


43081j:
@keithamus any chance you can give this a quick review? 🙏

eslint.config.js
```diff
@@ -1,12 +1,26 @@
 import jsdoc from "eslint-plugin-jsdoc";
+import eslintjs from "@eslint/js";
+
+const {configs: eslintConfigs} = eslintjs;
```
koddsson, line 4:
```suggestion
import {configs: eslintConfigs} from "@eslint/js";
```

Not possible?

lib/chai/interface/assert.js
```diff
@@ -3164,16 +3164,22 @@ assert.isNotEmpty = function (val, msg) {
...
-  'isExtensible',
-  'extensible'
-)('isNotExtensible', 'notExtensible')('isSealed', 'sealed')(
-  'isNotSealed',
-  'notSealed'
-)('isFrozen', 'frozen')('isNotFrozen', 'notFrozen')('isEmpty', 'empty')(
-  'isNotEmpty',
-  'notEmpty'
-)('isCallable', 'isFunction')('isNotCallable', 'isNotFunction');
+}
```
koddsson, line 3185: This reads so much better (to me at least)

koddsson (approved):
Looks good to me!

eslint.config.js
43081j: Not possible, it ships as a commonjs module and node tells me to do it this way if I try use a named import 

I could try again but doubt it'll have changed. Node would throw and told me it's because it uses a default cjs export

koddsson: Makes sense!

PR #1660, opened 2025-01-07
chore: introduce prettier and run it

43081j:
Adds prettier and a `format` script, and includes the results of the initial run.

i've manually reviewed the results and all seems good, but a second pair of eyes would be good to scan through it

cc @keithamus @koddsson


keithamus (approved):
LGTM

PR #1662, opened 2025-01-15
chore: drop old CJS entrypoints

43081j:
Removes the old `register-*.cjs` entrypoints as they have been broken since we shipped ESM-only (since `chai.cjs` doesn't exist anymore).


koddsson (approved):
Looks good!

PR #1664, opened 2025-01-17
Integrated `chai-subset` and added assert-based negation to containSubset

### LINKED ISSUE #1616: chai-subset is unmaitained

BreadInvasion:
A revival of @koddsson 's previous attempt at merging the abandoned `chai-subset` functionality, with some added comments and an assert-style negation of containSubset which was not present in the original package. I know there was some concern about code duplication in the original PR, so I'm happy to respond to feedback! Thanks all.

Fixes #1616


test/subset.js
```diff
@@ -0,0 +1,206 @@
+import {assert, expect} from '../index.js';
+
+describe('plain object', function () {
```
43081j, line 3: this file should have a root `describe('containSubset', ...)` block which these all live below

lib/chai/core/assertions.js
```diff
@@ -4061,3 +4062,89 @@ Assertion.addProperty('finite', function (msg) {
...
+  if (expected === actual) {
+    return true;
+  }
+  if (typeof actual !== typeof expected) {
+    return false;
+  }
+  if (typeof expected !== 'object' || expected === null) {
+    return expected === actual;
+  }
+  if (!!expected && !actual) {
```
43081j, line 4083: at this point, we already know `expected` is an `object`, so you can avoid the `!!`

```suggestion
  if(!actual) {
    return false;
  }
```

lib/chai/core/assertions.js
```diff
@@ -4061,3 +4062,89 @@ Assertion.addProperty('finite', function (msg) {
...
+    return false;
+  }
+
+  if (Array.isArray(expected)) {
+    if (typeof actual.length !== 'number') {
+      return false;
+    }
+    var aa = Array.prototype.slice.call(actual);
```
43081j, line 4091: is there a reason we clone the array? and why we don't use `Array.isArray(actual)` above?

43081j:
we should also add tests for using this with should and assert style assertions

otherwise looks good so far though 👍

test/subset.js
BreadInvasion: Fixed in latest push (force-pushed for rebase to 1 commit).

lib/chai/core/assertions.js
BreadInvasion: Fixed in latest push (force-pushed for rebase to 1 commit).

lib/chai/core/assertions.js
BreadInvasion: Fixed in latest push (force-pushed for rebase to 1 commit).

BreadInvasion:
@43081j I'm not sure if it's necessary to run the full battery of tests in each assertion style - is it not enough to ensure that they actually call the right function, since all three styles ultimately just call the same `containsSubset` function? We do this at 147-168 of the test file.

BreadInvasion:
Fixed merge conflict with the change in method of declaring aliases. Ready for re-review.

koddsson (approved):
This is looking good to me but I'll let @43081j have the final approval

lib/chai/core/assertions.js
```diff
@@ -4061,3 +4062,88 @@ Assertion.addProperty('finite', function (_msg) {
...
+  return Object.keys(expected).every(function (key) {
+    var eo = expected[key];
+    var ao = actual[key];
+    if (typeof eo === 'object' && eo !== null && ao !== null) {
+      return compareSubset(eo, ao);
+    }
+    if (typeof eo === 'function') {
+      return eo(ao);
+    }
+    return ao === eo;
```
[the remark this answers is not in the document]
43081j: I agree, let's keep the names readable to help other contributors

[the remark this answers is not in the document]
BreadInvasion: Fixed in latest commit!

lib/chai/core/assertions.js
```diff
@@ -4061,3 +4062,88 @@ Assertion.addProperty('finite', function (_msg) {
     'expected #{this} to not be a finite number'
   );
 });
+
+/**
+ * A subset-aware compare function
+ *
+ * @param {unknown} expected
+ * @param {unknown} actual
+ * @returns {void}
```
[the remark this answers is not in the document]
BreadInvasion: Fixed in latest commit!

43081j (approved):

43081j:
looks good to me 👍 

big thanks for implementing this 🙏

koddsson:
Thanks @BreadInvasion ❤️

Issue #1481, opened 2022-10-03
Incorrect closeTo result


Tb0:
Hello,
I've an issue with the method `closeTo`.

`expect(1682.6).to.be.closeTo(1682.7, 0.1)` returns an error ❌  while I was expecting the test to pass.

It seems to me that when you have a number with more than 3 digits this fails ❌ .

`expect(168.6).to.be.closeTo(168.7, 0.1)` is passing ✅ , while `expect(1682.6).to.be.closeTo(1682.7, 0.1)` doesn't ❌.

In the meantime, `expect(1682.6).to.be.closeTo(1682.7, 0.2)` is passing ✅.

Can you please tell me what's going on or if I misunderstand something?

Tb0:
Any update on this one?

PR #1667, opened 2025-01-22
fix floating point precision in closeTo assertion

koddsson:
Fixes https://github.com/chaijs/chai/issues/1481


keithamus (approved):
Looks reasonable. I'm a little sceptical of the code but OTOH we've got enough tests that I'm confident to give this a :heavy_check_mark:

PR #1675, opened 2025-02-23
Enable `no-var` rule and fix violations

koddsson:
Lets enforce more modern JavaScript.

See https://eslint.org/docs/latest/rules/no-var


43081j (approved):
this looks good but i suspect a whole bunch of these could be `const`. is that true?

maybe we should also enable `prefer-const` (i don't quite remember if that's the rule name) and auto-fix them

PR #1676, opened 2025-02-23
Run prettier as part of ESLint

koddsson:
This way we get a failure in CI if the code isn't formatted according to prettier.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,8 +37,8 @@
     "test-node": "c8 --99 --check-coverage mocha --require ./test/bootstrap/index.js test/*.js",
     "test-chrome": "web-test-runner --playwright",
     "lint": "npm run lint:js && npm run lint:format",
-    "lint:js": "eslint lib/",
-    "lint:format": "prettier --check lib",
+    "lint:js": "eslint lib/ && prettier --check lib/",
+    "lint:format": "eslint lib/ --fix && prettier lib/",
     "clean": "rm -rf chai.js coverage/"
   },
   "engines": {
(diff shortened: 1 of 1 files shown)


43081j:
i actually wouldn't do this

i'd just `prettier --check` in an npm script

one day really, prettier's eslint plugin/config should go away. since it should just be an automated thing run at the end

package.json
```diff
@@ -37,8 +37,8 @@
     "test-node": "c8 --99 --check-coverage mocha --require ./test/bootstrap/index.js test/*.js",
     "test-chrome": "web-test-runner --playwright",
     "lint": "npm run lint:js && npm run lint:format",
-    "lint:js": "eslint lib/",
-    "lint:format": "prettier --check lib",
+    "lint:js": "eslint lib/ && prettier --check lib/",
+    "lint:format": "eslint lib/ --fix && prettier lib/",
```
koddsson, line 41:
```suggestion
    "lint:format": "eslint lib/ --fix && prettier --write lib/",
```

package.json
```diff
@@ -37,8 +37,8 @@
     "test-node": "c8 --99 --check-coverage mocha --require ./test/bootstrap/index.js test/*.js",
     "test-chrome": "web-test-runner --playwright",
     "lint": "npm run lint:js && npm run lint:format",
-    "lint:js": "eslint lib/",
-    "lint:format": "prettier --check lib",
+    "lint:js": "eslint lib/ && prettier --check lib/",
+    "lint:format": "eslint lib/ --fix && prettier --write lib/",
```
43081j, line 41: This means the lint script will run both of them twice. I guess you meant to also update the lint script to only call lint:js?

Maybe we can make it clearer by calling the other lint:fix instead

PR #1677, opened 2025-02-23
Convert Assertion function to a class

koddsson:
I want to chip away at https://github.com/chaijs/chai/pull/1567 to make it easier to land.

This PR converts the `Assertion` function (which was being used as a class) into a real class.

### DIFF
--- a/lib/chai/assertion.js
+++ b/lib/chai/assertion.js
@@ -10,10 +10,7 @@ import {AssertionError} from 'assertion-error';
 import * as util from './utils/index.js';
 
 /**
- * Assertion Constructor
- *
  * Creates object for chaining.
- *
  * `Assertion` objects contain metadata in the form of flags. Three flags can
  * be assigned during instantiation by passing arguments to this constructor:
  *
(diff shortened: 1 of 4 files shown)


lib/chai/assertion.js
```diff
@@ -42,139 +39,135 @@ import * as util from './utils/index.js';
  *
  * - `eql`: This flag contains the deepEqual function to be used by the assertion.
  *
- * @param {unknown} obj target of the assertion
- * @param {string} msg (optional) custom error message
- * @param {Function} ssfi (optional) starting point for removing stack frames
- * @param {boolean} lockSsfi (optional) whether or not the ssfi flag is locked
+ * @param {unknown} ?obj target of the assertion
```
43081j, line 42: we should use typescript's jsdoc format

```ts
/**
 * @param {unknown=} obj some optional object
 */
```

koddsson: Yeah, I realised that this is wrong and fix it in #1679

keithamus (approved):

PR #1679, opened 2025-02-23
More typing

koddsson:
Adds more types hand-picked from #1567

A stacked PR waiting for #1678


koddsson:
I was trying to do fancy stacked PRs but I think the tool I was using was mis-configured. Should be correct going forward.

keithamus (approved):

PR #1742, opened 2025-11-12
assertion-error should be in dependencies, not devDependencies

Gautam-Bharadwaj:
### **Summary**
The `assertion-error` package is being used at runtime throughout the Chai library (e.g., in `chai.js`, `assertion.js`, `assertions.js`, etc.), but it is currently listed under `devDependencies` in `package.json`.

---

### **Problem**
Because `assertion-error` is listed as a development dependency, end users who install Chai via npm will not automatically receive it as a transitive dependency.  
This can cause **runtime import errors** when the library tries to require or import `assertion-error`.

---

### **Root Cause**
`assertion-error` is a **runtime dependency**, not a development-only one.  
It was incorrectly categorised under `devDependencies` instead of `dependencies`.

---

### **Proposed Solution**
Move the following entry:

----------------------------------------------------------------------------------------------------------------------------
Previous
----------------------------------------------------------------------------------------------------------------------------
Screenshot 2025-11-12 at 11 16 11 PM

----------------------------------------------------------------------------------------------------------------------------
Updated
----------------------------------------------------------------------------------------------------------------------------
Screenshot 2025-11-12 at 11 16 57 PM

----------------------------------------------------------------------------------------------------------------------------
```json
"assertion-error": "^2.0.1"

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,23 +48,25 @@
   "engines": {
     "node": ">=18"
   },
+  "dependencies": {
+    "assertion-error": "^2.0.1",
+    "check-error": "^2.1.1",
+    "deep-eql": "^5.0.1",
+    "loupe": "^3.1.0",
+    "pathval": "^2.0.0"
+  },
   "devDependencies": {
     "@eslint/js": "^9.17.0",
     "@rollup/plugin-commonjs": "^29.0.0",
     "@web/dev-server-rollup": "^0.6.1",
     "@web/test-runner": "^0.20.0",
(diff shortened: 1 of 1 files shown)


Gautam-Bharadwaj:
@43081j Got it, thanks for clarifying. Just to double-check, are all the published entry points fully bundled, with no external require('assertion-error') or import statements left in the npm package?

If the bundling is completely reliable for everything that gets published, then keeping it in devDependencies sounds good. Otherwise, it might be safer to move it to dependencies.

Issue #1743, opened 2025-11-12
Bug: Incorrect `showDiff` Default Value Handling in `Assertion.assert()`


Gautam-Bharadwaj:
# Bug: Incorrect `showDiff` Default Value Handling in `Assertion.assert()`

## Description
The `showDiff` parameter handling in the `Assertion.assert()` method contains a logic flaw that incorrectly forces **all non-`false` values** (including `null`, `0`, `''`, etc.) to `true`.  
This happens because the current condition defaults `showDiff` to `true` for any value not strictly equal to `false`, rather than only when `showDiff` is `undefined`.

---

## Location
- **File:** `lib/chai/assertion.js`  
- **Line:** 159  
- **Method:** `Assertion.prototype.assert()`

---

## Current Behavior
```javascript
assert(_expr, msg, _negateMsg, expected, _actual, showDiff) {
  const ok = util.test(this, arguments);
  if (false !== showDiff) showDiff = true;  // ❌ Bug: converts null, 0, '' → true
  if (undefined === expected && undefined === _actual) showDiff = false;
  if (true !== config.showDiff) showDiff = false;
  // ...
}

PR #1744, opened 2025-11-12
Bug: Incorrect `showDiff` Default Value Handling in `Assertion.assert()`

Gautam-Bharadwaj:
#1743 Solved Ready to merge
-------------------------------------------------------------------------------------------------------------------------
# Bug: Incorrect `showDiff` Default Value Handling in `Assertion.assert()`

## Description
The `showDiff` parameter handling in the `Assertion.assert()` method contains a logic flaw that incorrectly forces **all non-`false` values** (including `null`, `0`, `''`, etc.) to `true`.  
This happens because the current condition defaults `showDiff` to `true` for any value not strictly equal to `false`, rather than only when `showDiff` is `undefined`.

---

## Location
- **File:** `lib/chai/assertion.js`  
- **Line:** 159  
- **Method:** `Assertion.prototype.assert()`

---

## Current Behavior
```javascript
assert(_expr, msg, _negateMsg, expected, _actual, showDiff) {
  const ok = util.test(this, arguments);
  if (false !== showDiff) showDiff = true;  // ❌ Bug: converts null, 0, '' → true
  if (undefined === expected && undefined === _actual) showDiff = false;
  if (true !== config.showDiff) showDiff = false;
// ...
}

### DIFF
--- a/BUG_FIX_ISSUE.md
+++ b/BUG_FIX_ISSUE.md
@@ -1,0 +1,78 @@
+# Bug: Incorrect showDiff default value handling in Assertion.assert()
+
+## Description
+
+The `showDiff` parameter handling in `Assertion.assert()` method has a logic bug that incorrectly converts non-`false` values (including `null`, `0`, empty strings, etc.) to `true` instead of only defaulting when the parameter is `undefined`.
+
+## Location
+
+**File:** `lib/chai/assertion.js`  
+**Line:** 159  
+**Method:** `Assertion.prototype.assert()`
+
+## Current Behavior
+
+```javascript
+assert(_expr, msg, _negateMsg, expected, _actual, showDiff) {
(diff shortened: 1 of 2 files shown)


Gautam-Bharadwaj:
@jeffbski,@sophiebits,@duncanbeevers,@pwnall Hi! I’ve already fixed this issue and submitted a proposal for the change.
When you get a chance, could you please review my PR?

43081j:
this is on purpose and not a bug

if `showDiff` isn't explicitly `false`, it is the default: `true`. 

as per the types, it should only ever be `true | false | undefined` but could be anything at runtime. so we default to `true` unless you explicitly set it to `false`

for future reference i would wait for the issue to get a response before opening the PR in case there is a reason behind it like this one

PR #1748, opened 2025-11-20
fix: avoid BigInt literal in closeTo for runtime compat

bheemreddy-samsara:
## Summary
  - Avoid emitting a BigInt literal in `Assertion.closeTo`, which causes Hermes (React Native) to fail bundle
  evaluation with `No identifiers allowed directly after numeric literal`.
  - Use regular number math for `abs` so the code path stays compatible with engines lacking BigInt literal parsing.

  ## Context
  - RN 0.79.x, Hermes on Android.
  - Chai 5.3.3.
  - Hermes rejects the `0n` literal during bundle load; app crashes before tests run.

  ## Change
  - `lib/chai/core/assertions.js`: change `const abs = (x) => (x < 0n ? -x : x);` to `const abs = (x < 0 ? -x : x);`
  and document the rationale.

  ## Testing
  - Verified on a React Native app with Hermes: harness tests now load and pass; the prior syntax error is gone.

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -3147,9 +3147,6 @@ function closeTo(expected, delta, msg) {
   }
   new Assertion(expected, flagMsg, ssfi, true).is.numeric;
 
-  // Avoid emitting BigInt literals in the runtime path; some JS engines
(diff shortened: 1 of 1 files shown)


bheemreddy-samsara:
@snewcomer can i get a review on this ?

Issue #1708, opened 2025-10-02
Disable loupe's `customInspect` option to avoid collisions.


OldStarchy:
One of my objects has a `.inspect` function that collides with `chai/loupe`'s `customInspect` setting.

The options provided to loupe's `inspect` are hardcoded (which is why I'm opening the issue here rather than in loupe):
https://github.com/chaijs/chai/blob/0b0353cb18a9dcf8b6c3124802238283a210e174/lib/chai/utils/inspect.js#L24-L29

Setting [`customInspect: false`](https://github.com/chaijs/loupe/blob/b5b084b8429a3ad2ed4e8abfab2e3cdef92ba6e2/src/types.ts#L7) below would disable the check.
https://github.com/chaijs/loupe/blob/b5b084b8429a3ad2ed4e8abfab2e3cdef92ba6e2/src/index.ts#L99-L101

There's already an issue over there questioning the validity of using a string-keyd `'inspect'` function given that there is support for Symbols widely available https://github.com/chaijs/loupe/issues/74, but it seems to have been missed by the maintainers of the repo.
https://github.com/chaijs/loupe/blob/b5b084b8429a3ad2ed4e8abfab2e3cdef92ba6e2/src/index.ts#L28-L30

keithamus:
Let's resolve this in https://github.com/chaijs/loupe/issues/74. PRs welcome for that issue.

PR #1750, opened 2025-11-22
fix(inspect): disable string-based customInspect to avoid .inspect collisions

Himanshu197200:
### Summary (#1708)

This PR fixes a long-standing issue where Chai's `inspect` wrapper
accidentally enables loupe's string-based custom inspection behavior,
causing collisions with user-defined `.inspect()` methods.

Because Chai hardcodes the options passed to loupe's `inspect` and does
not currently expose a way to configure `customInspect`, loupe ends up
calling any object's enumerable `.inspect()` method—even if it is not
intended as a Node.js inspection hook.

This results in objects being printed incorrectly or throwing errors
when their `.inspect()` method expects different parameters.

### Root Cause

loupe still interprets a plain string-keyed `.inspect` property as a
custom inspector when `customInspect` is true. Since Chai does not set
this option, the default behavior triggers unintentionally.

### Fix

This PR explicitly sets:

```js
customInspect: false

### DIFF
--- a/lib/chai/utils/inspect.js
+++ b/lib/chai/utils/inspect.js
@@ -25,7 +25,8 @@ export function inspect(obj, showHidden, depth, colors) {
     colors: colors,
     depth: typeof depth === 'undefined' ? 2 : depth,
     showHidden: showHidden,
-    truncate: config.truncateThreshold ? config.truncateThreshold : Infinity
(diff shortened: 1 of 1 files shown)


43081j:
is this not already fixed by chaijs/loupe#114?

we removed `inspect` from loupe itself, and now require users to use a `Symbol` to set it instead

PR #1753, opened 2025-11-29
fix: allow empty keys assertion when target object is empty

milanchahar:
### Description
This PR fixes the issue where expecting empty keys on an empty object would throw "keys required".

It allows: `expect({}).to.have.all.keys([])` to pass.

### History
This is a manual revival of PR #1384 by @Zirak. I have reapplied the logic to the latest codebase and added a regression test.

Closes #1384

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -2535,6 +2535,7 @@ function assertKeys(keys) {
     all = true;
   }
 
+  const uniqueExpected = [...new Set(expected)];
   const canHaveEmptyKeys = all && !contains && keysType !== 'undefined';
 
   if (!keys.length && !canHaveEmptyKeys) {
@@ -2543,7 +2544,7 @@ function assertKeys(keys) {
 
   // Has any
   if (any) {
-    ok = expected.some(function (expectedKey) {
+    ok = uniqueExpected.some(function (expectedKey) {
       return actual.some(function (actualKey) {
         return isEql(expectedKey, actualKey);
       });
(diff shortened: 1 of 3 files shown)


test/issue-1384.js
```diff
@@ -0,0 +1,15 @@
+import {expect} from '../index.js';
+
+describe('Fix #1384: Empty keys support', function () {
+  it('should accept empty keys when the object is empty', function () {
```
43081j: these should be added to the existing `expect` tests, either as part of the existing test case or just move these two new ones and reword them

milanchahar:
@43081j  Thanks for the review! I have updated the PR with the requested changes:

1.  Reverted `package-lock.json` to discard unintended changes.
2.  Removed the separate test file and moved the regression tests into `test/expect.js`.
3.  Updated the logic in `assertions.js`. Now it strictly checks for the `all` flag.
    * `.all.keys([])` -> **Passes** (on empty target)
    * `.any.keys([])` -> **Throws Error** (keys required)

Ready for another look!

43081j:
did you forget to push? i can see you moved the tests but can't see the other changes there

milanchahar:
@43081j  Sir please check.

lib/chai/core/assertions.js
```diff
@@ -2538,6 +2534,12 @@ function assertKeys(keys) {
     all = true;
   }
 
+  // Fix for PR #1384: Allow empty keys only if we are checking 'all'.
+  // 'any' keys requires at least one key to be checked against.
```
43081j:
```suggestion
  // Allow empty keys if we're checking `all` since you can assert it has _no keys_
  // For `any`, we need at least one key.
```

since the PR isn't relevant info anymore at this point (its a feature rather than a bug)

43081j:
did you run a formatter? the diff is huge now 😬 

if you can narrow it back down to the tests you added, we're good to go

43081j:
i've updated this to be the correct logic now but im unsure how safe of a change it is

@keithamus can i get your opinion here? if we change this, is it really a fix or is it a breaking change?

we're basically saying `keys([])` means "have no keys" now, instead of erroring (what main does now).

i suppose it is a validation error rather than an assertion error we're changing, so nobody in the wild should be relying on it. but still unsure

PR #1756, opened 2025-12-14
chore: add --legal-comments=none option

hyperz111:
Before: `140.7Kb`
After: `133.8Kb`
Reduce **5%** of size.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
   },
   "main": "./index.js",
   "scripts": {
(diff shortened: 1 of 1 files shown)


43081j:
@keithamus what do you think of this? all our dependencies are our own, so the only license comments in there are chai ones. probably should be ok in that case

hyperz111:
Maybe close it?

keithamus:
Right, and we ship `LICENSE`, so this seems fine.

PR #1798, opened 2026-04-27
fix: support deep include oneOf assertions

### LINKED ISSUE #1493: deep.include.oneOf not working correctly
I would expect the following to pass, but it doesn't:
```
chai.expect([ [1], [2] ]).to.deep.include.oneOf([ [1], [2] ]);
```

I think this is related to #1241, but was under the impression that it had been fixed by #1334? I am using chai 4.3.7 which is the current latest.

lawrence3699:
Fixes #1493.

`deep.include.oneOf` currently takes the same strict containment path as `include.oneOf`, so nested arrays or objects are compared by identity even when the `deep` flag is present. For example, `expect([[1], [2]]).to.deep.include.oneOf([[1], [2]])` fails on current `main`.

This keeps the existing strict `indexOf` path, and falls back to Chai's configured deep equality comparison when `.deep` is set and the target supports array-style iteration.

Validation:
- `npm run build && npx mocha --require ./test/bootstrap/index.js test/expect.js --grep "oneOf"`
- `npm run lint`
- `npm run test-node`
- `npm run test-chrome`

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -3394,14 +3394,18 @@ function oneOf(list, msg) {
         return (
           expected.indexOf(possibility) > -1 ||
           (isDeep &&
-            typeof expected.some === 'function' &&
+            Array.isArray(expected) &&
             expected.some(function (item) {
               return eql(item, possibility);
             }))
         );
       }),
-      'expected #{this} to contain one of #{exp}',
-      'expected #{this} to not contain one of #{exp}',
+      'expected #{this} to ' +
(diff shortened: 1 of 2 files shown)


43081j (approved):

43081j:
@lawrence3699 FYI i made a couple of minor changes:

- use `Array.isArray` instead of checking the existence of `some` (we already assume it is an array so arguably don't need either, but this is fine)
- added error case tests
- changed `include` to `contain` in the test cases for consistency

PR #1835, opened 2026-07-22
docs: fix nestedInclude path and notMatch example in assert JSDoc

### LINKED ISSUE #1299: Documentation error api/assert nestedInclude
Error [page](https://www.chaijs.com/api/assert/)

> assert.nestedInclude({'.a': {'b': 'x'}}, {'\\.a.[b]': 'x'});
to
> assert.nestedInclude({'.a': {'b': 'x'}}, {'\\.a.b': 'x'});

Hashim1999164:
## Summary
- Correct the `assert.nestedInclude` JSDoc example so the escaped path is `\\.a.b` (matching `notNestedInclude` and the object shape `{'.a': {'b': 'x'}}`).
- Change the `assert.notMatch` example from `'foobar'` to `'barfoo'` so the sample actually fails `/^foo/` as the comment describes.

Fixes #1299

## Test plan
- [x] Confirm examples match the patterns already used in neighboring JSDoc (`notNestedInclude`)
- [ ] Docs site regenerates from assert.js JSDoc on next publish

### DIFF
--- a/lib/chai/interface/assert.js
+++ b/lib/chai/interface/assert.js
@@ -1020,7 +1020,7 @@ assert.notDeepInclude = function (exp, inc, msg) {
  * properties.
  * '[]' and '.' in property names can be escaped using double backslashes.
  *
- *     assert.nestedInclude({'.a': {'b': 'x'}}, {'\\.a.[b]': 'x'});
+ *     assert.nestedInclude({'.a': {'b': 'x'}}, {'\\.a.b': 'x'});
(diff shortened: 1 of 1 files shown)


lbesecker195 (approved):
Verified: the `notNestedInclude` example a few lines down already uses `'\\.a.b'` for the same object shape, and `'foobar'` does match `/^foo/`, so the old `notMatch` example would fail. Both fixes are correct. LGTM.

PR #1857, opened 2026-09-12
fix: reject ordered member sequences longer than the target

emme1t:
Ordered member inclusion currently accepts expected sequences that extend past the target with `undefined` values. For example, `expect([1]).to.include.ordered.members([1, undefined])` passes because reading past the target's end produces `undefined`.

Check the sequence lengths before ordered comparison. Regression coverage exercises strict and deep comparisons through `expect`, `should`, and `assert`, including empty targets, trailing `undefined` values, valid equal-length matches, and negation.

Validation on Node.js v24.16.0 (Windows):

- Before the fix: all 8 affected test cases fail with the added regressions.
- After the fix: all 8 affected test cases pass.
- `npm run test-node -- --reporter dot`: 511 passing.
- `npm run lint`: passes with the existing JSDoc warnings.
- Browser suite: 510 passing on Chromium 151.0.7922.34, using the repository's browser configuration with a local executable path override.

Developed with assistance from OpenAI Codex; reproduction and tests were run locally.

### DIFF
--- a/lib/chai/core/assertions.js
+++ b/lib/chai/core/assertions.js
@@ -3176,6 +3176,7 @@ Assertion.addMethod('approximately', closeTo);
 function isSubsetOf(_subset, _superset, cmp, contains, ordered) {
   let superset = Array.from(_superset);
   let subset = Array.from(_subset);
+  if (ordered && subset.length > superset.length) return false;
   if (!contains) {
     if (subset.length !== superset.length) return false;
     superset = superset.slice();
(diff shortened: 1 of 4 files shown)


lbesecker195 (approved):
Confirmed the bug on main: `isSubsetOf` compares `elem === superset[idx]` for ordered subsets, so reading past the end of the target yields `undefined` and `expect([1]).to.include.ordered.members([1, undefined])`, `expect([]).to.include.ordered.members([undefined])` and `assert.includeOrderedMembers([1], [1, undefined])` all pass. On this branch they fail with the expected message, while `[1, 2, 3]` including `[1, 2]`, `[1, undefined]` including `[1, undefined]`, and `[]` including `[]` still pass. The non-`include` path already checks lengths, so the new guard only affects `contains`. After `npm run build`, the node suite is 511 passing. LGTM.

